PR #34, opened 2023-02-02
Configurable generator prefixes/suffixes

### LINKED ISSUE #21: Make generated prefix/suffix configurable

mhuster23:
closes #21
Maybe additional discussion about the default values is required.


languitar (commented):
Is there some chance to add tests to show that  this works as expected?

slisson:
The prefixes should be configurable in the gradle plugin.
https://github.com/modelix/modelix.core/blob/main/metamodel-gradle/src/main/kotlin/org/modelix/metamodel/gradle/MetaModelGradleSettings.kt

slisson:
You could use create the NameConfig object already in MetaModelGradleSettings instead of copying all the values. I would also prefer a bit more structure when using it in gradle. For example like this:

```
metamodel {
    names {
        nodeWrapperInterface.prefix "N_"
        nodeWrapperImpl.suffix = "Impl"
    }
}
```

The default values should be current ones to avoid breaking changes.

slisson (approved):

PR #35, opened 2023-02-04
Test project for the metamodel-gradle plugin

slisson:
This new test project generates typed model API Kotlin classes for all the languages that are part of MPS. It is executed as part of the github actions workflow. So far the generator was only tested indirectly in the projects using it.

I fixed the handling of duplicate concept members (same name) in a concept hierarchy.
In MPS there are the following concepts:

```
ClassConcept extends Classifier implements IBLDeprecatable

Classifier {
    isDeprecated: boolean
}

IBLDeprecatable {
    isDeprecated: boolean
}

```

This was generated into

```
class ClassConcept : Classifier, IBLDeprecatable {
    override var isDeprecated: Boolean
}

class Classifier {
    var isDeprecated: Boolean
}

class IBLDeprecatable {
    var isDeprecated: Boolean
}
```

which is incorrect. These are two different properties that are also handled are two different properties in MPS. There is override happening in MPS.

Now, the concept name is added as a suffix to resolve this ambiguity.

```
class ClassConcept : Classifier, IBLDeprecatable {

}

class Classifier {
    var isDeprecated_Classifier: Boolean
}

class IBLDeprecatable {
    var isDeprecated_IBLDeprecatable: Boolean
}
```


slisson:
Possible next steps
- Also generate and compile TypeScript code as part of the test project.
- Write some unit tests that use the generated API. The plugin already supports exporting models from MPS. Those could be loaded as test data.

PR #42, opened 2023-02-09
Implement automated releases

languitar:


commitlint.config.js
```diff
@@ -0,0 +1,21 @@
+module.exports = {
+  extends: ["@commitlint/config-conventional"],
+  rules: {
+    "scope-enum": [
+      2,
+      "always",
+      [
+        "kotlin-js-store",
```
languitar, line 8: This one is not needed

languitar: removed

languitar:
I've fixed the last bits and pieces. I've also validated that the commit scope can be left empty. So this would be ready for review/merge from my side now.

nkoester (dismissed):
LGTM

nkoester (approved):
LGTM

PR #59, opened 2023-03-21
MODELIX-179: Browser-based explorer for model server content

mhuster23:
Accessible via the route `/content/`.


model-server/src/main/kotlin/org/modelix/model/server/handlers/ContentExplorer.kt
```diff
@@ -0,0 +1,87 @@
...
+    fun init(application: Application) {
+        application.routing {
+            get("/content/") {
+                call.respond(ThymeleafContent("content", mapOf("rootNodes" to rootNodes)))
+            }
+            get("/content/{nodeId}/") {
+                val id = call.parameters["nodeId"]!!.toLong()
+                var found: PNodeAdapter? = null
+                for (node in rootNodes) {
+                    val candidate = node.findChildRec(id)
```
slisson, line 60: `PNodeAdapter(id, rootNode.branch).takeIf { it.isValid }`

model-server/src/main/kotlin/org/modelix/model/server/handlers/ContentExplorer.kt
```diff
@@ -0,0 +1,87 @@
...
+import org.modelix.model.ModelFacade
+import org.modelix.model.api.*
+import org.modelix.model.client.IModelClient
+import org.modelix.model.lazy.CLVersion
+import org.modelix.model.lazy.RepositoryId
+import org.modelix.model.metameta.MetaModelBranch
+
+class ContentExplorer(private val client: IModelClient) {
+
+    private val knownRepositoryIds: Set<HistoryHandler.RepositoryAndBranch>
```
slisson, line 17: There is now a `RepositoriesManager` that provides this information, but I just realised that it isn't yet backward compatible to this legacy way of storing repository meta information (aka "info" repository).

model-server/src/main/kotlin/org/modelix/model/server/handlers/ContentExplorer.kt
```diff
@@ -0,0 +1,87 @@
...
+            }
+            return nodeList
+        }
+
+    fun init(application: Application) {
+        application.routing {
+            get("/content/") {
+                call.respond(ThymeleafContent("content", mapOf("rootNodes" to rootNodes)))
+            }
+            get("/content/{nodeId}/") {
```
slisson, line 56: The path should contain the versionHash and then load the correct version. This allows to browse the content of a specific branch or even a single commit.

slisson (changes requested):

model-server/src/main/kotlin/org/modelix/model/server/handlers/ContentExplorer.kt
slisson: The version hash is unique across repositories and since all repositories are stored in the same database, the repository ID is not required to load a version.

slisson:
There are still a few things I would change and I already tried to do them myself, but after playing around with Thymeleaf I have to say this is not an improvement over the Kotlin HTML DSL.

There is this overhead of preparing the data for the templates and then the expressions are just strings without code completion or any static checks by the compiler. Extension methods or other Kotlin features don't work. You have to run the template to find out that there is an error in the expressions. This makes this approach unstable and hard to maintain.

At least for the model server use case, where we just want to provide some basic functionality, but not build a complex application, I would say the Thymeleaf approach has more disadvantages than advantages.

mhuster23:
Yeah, I see your point. 

The lack of support for extension functions is something that I did not realize beforehand. It bothered me while implementing this because we make heavy use of them. You can still call the extension functions statically in templates but that makes the template less readable and defeats the purpose of having extension functions in the first place.

I considered the lack of static checks acceptable for this use case. But combined with the other downsides I have to agree, that the cons outweigh the pros of using Thymeleaf in this case.

For now, I will reimplement this with the Kotlin HTML DSL.

slisson:
Choosing a version should look more like in the history view at `/history/` (table with repository and branch name). It probably makes sense to have a single "Repositories" page that lists the repositories and branches with two links for the history and the content next to each of them. "History" and "Content" can then be removed from the navigation bar.

slisson (approved):

PR #61, opened 2023-03-21
fix: support .filterLoaded() on a child list of typed nodes

slisson:
You can now write for example node.content.filterLoaded()


languitar:
Isn't this a new feature? So more like a `feat` commit?

slisson:
I thought it's more like a missing part of a previously added feature.

languitar:
> I thought it's more like a missing part of a previously added feature.

But there wasn't anything really broken in the current state that needs patching. The agile world would probably say just a missing feature.

languitar (approved):

PR #63, opened 2023-03-22
MODELIX-280 ci: only build on semantic version tags

nkoester:


slisson:
The documentation doesn't mention that regular expressions are supported. It only talks about "glob patterns".

nkoester:
When tagging our upcoming releases with the release tag (e.g. `modelix-23.1`) these would then also be build and published. However, we do not need additional artifacts as we plan to provide component collections as BOM or gradle plugin. The tag is then only required for logical organization and documentation generation. Thus only building `x.y.z*` would be sufficient.

languitar:
> But I also want to publish a branch without making it an official release. See https://semver.org/#spec-item-9

If publishing pre-releases is required regularly, that should probably become part of the semantic-release configuration: https://semantic-release.gitbook.io/semantic-release/recipes/release-workflow/pre-releases

slisson (approved):

PR #66, opened 2023-03-23
MODELIX-54 Higher level API for the advanced-model-client

slisson:
There is a new REST API and a re-implementation of the "advanced" model client. The old implementation still exists. This change is more like a preview of this new API. It Still needs some more extensive testing and performance optimization.


slisson:
The most relevant change (the REST API) is in [ModelReplicationServer.kt](https://github.com/modelix/modelix.core/pull/66/files#diff-730e8af4d3eaec9de7a7e47e8d26ec520f8a4f0c1013c283222876b7aecf9e61).

PR #67, opened 2023-03-23
support for "references" and "referencesAndChildren" in ModelQL

slisson:


languitar:
This probably also needs a release of the sever part?

languitar (approved):

slisson:
Yes, the query is executed on the server side and requires a new version of the MPS plugin.

PR #70, opened 2023-03-31
MODELIX-402: Publish pipeline in modelix.core doesn't run on releases

mhuster23:
For authorization the checkout action uses the GITHUB_TOKEN by default.
This prevents the publish pipeline to be triggered by the tag created by semantic release.
Now the personal access token is used instead.


slisson (approved):

languitar:
👍

PR #72, opened 2023-04-03
fix: module import for vite

mve-itemis:
the provided file does not exist and consequently results in vite throwing an error


languitar:
The commit message must be reworded to match the [commit convention](https://github.com/modelix/modelix.core#commit-convention). That's the reason for one of the pipeline steps failing.

PR #85, opened 2023-04-27
build: use version catalog instead of properties

slisson:
To support dependabot.

### DIFF
--- a/model-client/build.gradle.kts
+++ b/model-client/build.gradle.kts
@@ -49,14 +49,14 @@ kotlin {
                 api(project(":model-api"))
                 api(project(":model-server-api"))
                 kotlin("stdlib-common")
-                implementation("org.jetbrains.kotlinx:kotlinx-collections-immutable:0.3.4")
-                implementation("org.jetbrains.kotlinx:kotlinx-coroutines-core:$kotlinCoroutinesVersion")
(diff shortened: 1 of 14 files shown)


model-client/build.gradle.kts
```diff
@@ -69,27 +69,22 @@ kotlin {
             dependencies {
                 kotlin("stdlib-jdk8")
 
-                implementation("io.vavr:vavr:0.10.3")
-                implementation("org.apache.commons:commons-lang3:3.11")
-                implementation("com.google.guava:guava:29.0-jre")
-                implementation("org.glassfish.jersey.core:jersey-client:2.31")
-                implementation("org.glassfish.jersey.inject:jersey-hk2:2.31")
-                implementation("org.glassfish.jersey.media:jersey-media-sse:2.31")
-                implementation("javax.xml.bind:jaxb-api:2.3.1")
```
mhuster23, line 78:
```
implementation("org.glassfish.jersey.core:jersey-client:2.31")
implementation("org.glassfish.jersey.inject:jersey-hk2:2.31")
implementation("org.glassfish.jersey.media:jersey-media-sse:2.31")
implementation("javax.xml.bind:jaxb-api:2.3.1")
```
Is it intentional to remove these entirely?

slisson: These are old unused deps. The green build confirms that.

mhuster23 (approved):

PR #115, opened 2023-05-23
MODELIX-435 new tryReference(...) in ModelQL

slisson:
Doesn't throw an exception if the reference doesn't exist on the concept of the current node.

A cleaner solution would be to have some concept switch in ModelQL that applies the subquery only to instances of a specific concept, but the ModelQL will be replaced soon. This is just a quick fix for an urgent issue.

https://issues.modelix.org/issue/MODELIX-435/Using-sometimes-non-existing-references-in-ModelQL-results-in-a-ServerSideException


languitar:
This means we would have to adapt our queries to get rid of the exception that is raised right now?

languitar (approved):

PR #130, opened 2023-06-07
MODELIX-411 /health endpoint

slisson:
https://issues.modelix.org/issue/MODELIX-411

### DIFF
--- a/model-server-lib/src/main/kotlin/org/modelix/model/server/light/LightModelServer.kt
+++ b/model-server-lib/src/main/kotlin/org/modelix/model/server/light/LightModelServer.kt
@@ -56,8 +56,7 @@ import java.time.Duration
 import java.util.*
 import kotlin.time.Duration.Companion.seconds
 
-class LightModelServer(val port: Int, val rootNode: INode, val ignoredRoles: Set<IRole> = emptySet(), additionalHealthChecks: List<IHealthCheck>) {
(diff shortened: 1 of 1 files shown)


model-server-lib/src/main/kotlin/org/modelix/model/server/light/LightModelServer.kt
```diff
@@ -54,14 +56,26 @@ import java.time.Duration
 import java.util.*
 import kotlin.time.Duration.Companion.seconds
 
-class LightModelServer(val port: Int, val rootNode: INode, val ignoredRoles: Set<IRole> = emptySet()) {
+class LightModelServer(val port: Int, val rootNode: INode, val ignoredRoles: Set<IRole> = emptySet(), additionalHealthChecks: List<IHealthCheck>) {
+    constructor(port: Int, rootNode: INode, ignoredRoles: Set<IRole> = emptySet()) : this(port, rootNode, ignoredRoles, emptyList())
```
mhuster23, line 60:
```suggestion
class LightModelServer(
    val port: Int,
    val rootNode: INode,
    val ignoredRoles: Set<IRole> = emptySet(),
    additionalHealthChecks: List<IHealthCheck> = emptyList()
) {
```
A secondary constructor seems quite verbose to me in this case.

slisson: The reason was to avoid a breaking change in MPS, but `@JvmOverloads` is probably a better alternative.

mhuster23 (approved):

PR #131, opened 2023-06-08
MODELIX-443 ModelQL v2

slisson:


mhuster23 (commented):
I only noticed one minor thing: quite a few files have a doubled license header (e.g. `AllChildrenTraversalStep.kt`. Is there sth wrong with the config?

light-model-client/src/commonMain/kotlin/org/modelix/client/light/LightModelClient.kt
```diff
@@ -40,6 +80,13 @@ class LightModelClient internal constructor(
     private var exceptions: MutableList<ExceptionData> = ArrayList()
     private var currentModelQuery: ModelQuery? = null
     private var unappliedQuery: ModelQuery? = null
+    private val modelQLClient: ModelQLClient? = when (connection) {
```
odzhychko, line 83: Suggestion: Replace pattern matching with polymorphism.

`IConnection` could declare and `WebsocketConnection` could implement a method like `getModelQLClient`.

light-model-client/src/commonMain/kotlin/org/modelix/client/light/LightModelClient.kt
```diff
@@ -40,6 +80,13 @@ class LightModelClient internal constructor(
...
+        is WebsocketConnection -> ModelQLClient(
+            connection.url.replace(Regex("ws://(.*)/ws"), "http://$1/query"),
```
odzhychko, line 85: Suggestion: Do not use a regex. Add fields `host` and `port` in `WebsocketConnection` and use the to create the URL for the `ModelQLClient`.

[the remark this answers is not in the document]
odzhychko: Ah, the websocket connection is also unencrypted (`ws://` and not `wss://`).
... so the regex would fail for e.g. `wss://example.com/ws`.

mhuster23 (approved):

PR #138, opened 2023-06-13
MODELIX-458 Using generated enums leads to IllegalArgumentException

mhuster23:
https://issues.modelix.org/issue/MODELIX-458

### DIFF
--- a/model-api-gen-runtime/src/commonMain/kotlin/org/modelix/metamodel/IPropertyValueSerializer.kt
+++ b/model-api-gen-runtime/src/commonMain/kotlin/org/modelix/metamodel/IPropertyValueSerializer.kt
@@ -80,7 +80,7 @@ object OptionalIntPropertySerializer : IPropertyValueSerializer<Int?> {
 
 class MandatoryEnumSerializer<E : Enum<*>>(
     private val memberIdOf: (E) -> String,
-    private val fromString: (String?) -> E
+    private val fromMemberId: (String?) -> E
 ) : IPropertyValueSerializer<E> {
 
     override fun serialize(value: E): String {
@@ -89,13 +89,13 @@ class MandatoryEnumSerializer<E : Enum<*>>(
 
     override fun deserialize(serialized: String?): E {
         val id = serialized?.substringBefore('/')
-        return fromString(id)
+        return fromMemberId(id)
     }
 }
 
 class OptionalEnumSerializer<E : Enum<*>>(
     private val memberIdOf: (E) -> String,
-    private val fromString: (String) -> E
+    private val fromMemberId: (String) -> E
 ) : IPropertyValueSerializer<E?> {
 
     override fun serialize(value: E?): String? {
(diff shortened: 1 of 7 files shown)


model-api-gen/src/main/kotlin/org/modelix/metamodel/generator/MetaModelGenerator.kt
```diff
@@ -317,7 +317,7 @@ class MetaModelGenerator(val outputDir: Path, val nameConfig: NameConfig = NameC
                         if (feature.type is EnumPropertyType) {
                             if (serializer == MandatoryEnumSerializer::class.asTypeName()) {
                                 propBuilder.initializer(
-                                    """newProperty(%S, %S, %T { if (it != null) %T.valueOf(it) else %T.defaultValue() }, ${feature.optional})""",
+                                    """newProperty(%S, %S, %T { if (it != null) %T.valueOf(it.substringAfterLast('/')) else %T.defaultValue() }, ${feature.optional})""",
```
slisson, line 320: This should go into MandatoryEnumSerializer/OptionalEnumSerializer. Then don't use the name, but export the ID of the enum member from MPS and use the part before the slash for deserialization.

The implementation of enum serialization/deserialization in MPS can be found here: https://github.com/JetBrains/MPS/blob/346e4440f6b05dabdc34f1be67849bdc942761b4/core/kernel/source/jetbrains/mps/smodel/adapter/structure/types/SEnumerationAdapter.java#L113-L181

slisson (changes requested):

model-api/src/commonMain/kotlin/org/modelix/model/data/MetaModelData.kt
```diff
@@ -51,8 +50,10 @@ data class EnumData(
 
 @Serializable
 data class EnumMemberData(
-    val uid: String?,
-    val name: String
+    val uid: String? = null,
+    val memberId: String,
```
slisson, line 54: We don't need two IDs. You can just write the MPS ID into `uid`.

model-api-gen-runtime/src/commonMain/kotlin/org/modelix/metamodel/IPropertyValueSerializer.kt
```diff
@@ -78,23 +78,32 @@ object OptionalIntPropertySerializer : IPropertyValueSerializer<Int?> {
...
+) : IPropertyValueSerializer<E?> {
+
+    override fun serialize(value: E?): String? {
+        return value?.let { memberIdOf(it) + "/" + it.name}
     }
 
-    override fun deserialize(serialized: String?): Enum<*>? {
-        return serialized?.let { valueOf(serialized) }
+    override fun deserialize(serialized: String?): E? {
+        return serialized?.let { fromString(it) }
```
slisson, line 106: The .substringBefore part is missing here. Make OptionalEnumSerializer a subclass of MandatoryEnumSerializer to remove the code duplication. Also rename `fromString` to make clear that it expects the memberId.

slisson (changes requested):

PR #139, opened 2023-06-14
MODELIX-456 exportMetaModelFromMps caches too eagerly

mhuster23:
https://issues.modelix.org/issue/MODELIX-456

### DIFF
--- a/model-api-gen-gradle/src/main/kotlin/org/modelix/metamodel/gradle/MetaModelGradlePlugin.kt
+++ b/model-api-gen-gradle/src/main/kotlin/org/modelix/metamodel/gradle/MetaModelGradlePlugin.kt
@@ -38,6 +38,7 @@ class MetaModelGradlePlugin: Plugin<Project> {
             task.antScriptFile.set(getAntScriptFile())
             task.exporterDir.set(exporterDir.absolutePath)
             task.moduleFolders.addAll(settings.moduleFolders.map { it.absolutePath })
+            task.inputs.property("coreVersion", modelixCoreVersion)
         }
 
         val antDependencies = project.configurations.create("metamodel-ant-dependencies")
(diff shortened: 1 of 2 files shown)


model-api-gen-gradle/src/main/kotlin/org/modelix/metamodel/gradle/MetaModelGradlePlugin.kt
```diff
@@ -79,6 +80,7 @@ class MetaModelGradlePlugin: Plugin<Project> {
         }
         val generateMetaModelSources = project.tasks.register("generateMetaModelSources", GenerateMetaModelSources::class.java) {task ->
             task.dependsOn(exportMetaModelFromMps)
+            task.inputs.property("jsonFormatVersion", LanguageData.formatVersion) //necessary for cache invalidation (see https://issues.modelix.org/issue/MODELIX-456)
```
slisson, line 83: There is already a variable `modelixCoreVersion` that you can use instead. See https://github.com/modelix/modelix.core/blob/9ad7a778a4b45ebda67e8259b8ba50df8eae400d/model-api-gen-gradle/src/main/kotlin/org/modelix/metamodel/gradle/MetaModelGradlePlugin.kt#L23

slisson (changes requested):

model-api-gen-gradle/src/main/kotlin/org/modelix/metamodel/gradle/MetaModelGradlePlugin.kt
mhuster23: Yes, but that would invalidate the cache for every new modelix core version, even if the format (or model-api-gen) did not change at all. But using the core version would be better for maintainability.

slisson: I just realized that this is also the wrong task. `exportMetaModelFromMps` writes the JSON files.

slisson: All tasks have to rerun if the version changes.

slisson (approved):

PR #140, opened 2023-06-15
MODELIX-447 model-sync-lib: New component for exporting/importing from model-api to JSON

mhuster23:


model-sync-lib/src/main/kotlin/org/modelix/model/sync/ModelImporter.kt
```diff
@@ -0,0 +1,99 @@
...
+
+    private fun syncChildOrder(node: INode, nodeData: NodeData) {
+        val existingChildren = node.allChildren.toList()
+        val specifiedChildren = nodeData.children
+        require(existingChildren.size == specifiedChildren.size)
+
+        for ((actualNode, specifiedNode) in existingChildren zip specifiedChildren) {
+            val actualId = actualNode.originalId()
+            if (actualId != specifiedNode.originalId() ) {
+                val targetIndex = specifiedChildren.indexOfFirst { actualId == it.originalId() }
```
slisson, line 85: That's the wrong index. moveChild expects the index inside the role, but this is the index across all roles.

model-sync-lib/src/main/kotlin/org/modelix/model/sync/ModelImporter.kt
```diff
@@ -0,0 +1,99 @@
...
+
+        val data = ModelData.fromJson(jsonFile.readText())
+        import(data)
+    }
+    
+    fun import(data: ModelData) {
+        originalIdToRef = mutableMapOf()
+        branch.runWrite {
+            val rootNode = branch.getRootNode()
+            if (rootNode.allChildren.none()) {
```
slisson, line 24: Remove this case. The importer should run its own implementation. `ModelData.load` should probably be removed later and replaced by this importer.

model-sync-lib/src/test/kotlin/org/modelix/model/sync/ModelImporterTest.kt
```diff
@@ -0,0 +1,96 @@
...
+            val node1 = branch.getRootNode().allChildren.toList()[1]
+
+            assertEquals(node1, node0.getReferenceTarget("sibling"))
+            assertEquals(node0, node1.getReferenceTarget("sibling"))
+            assertEquals(branch.getRootNode(), node1.getReferenceTarget("root"))
+        }
+    }
+
+    @Test
+    fun `can sync children`() {
```
slisson, line 71: These tests (or additional ones) should also check if the import is actually incremental which means that an existing model is updated with a minimal set of operations. The importer could just delete and recreate everything and the test would pass, but that's not what we want.

model-sync-lib/src/test/kotlin/org/modelix/model/sync/ModelImporterTest.kt
```diff
@@ -0,0 +1,96 @@
...
+    companion object {
+        private lateinit var model: ModelData
+        private lateinit var newModel: ModelData
+        private lateinit var branch: IBranch
+
+        @JvmStatic
+        @BeforeAll
+        fun `load and import model`() {
+            model = ModelData.fromJson(File("src/test/resources/model.json").readText())
+            val newModelFile = File("src/test/resources/newmodel.json")
```
slisson, line 30: In addition to this hand written test data I would also use same randomly generated input to cover more corner cases. See `RandomTreeChangeGenerator` and its usages for an example. It's good at finding all the cases that you would never think about.

slisson (changes requested):

model-sync-lib/src/main/kotlin/org/modelix/model/sync/ModelImporter.kt
```diff
@@ -0,0 +1,332 @@
...
+
+        val data = ModelData.fromJson(jsonFile.readText())
+        import(data)
+    }
+    
+    fun import(data: ModelData) {
+        stats?.reset()
+
+        syncProperties(root, data.root) // root original id is required for following operations
+        val allExistingNodes = root.getDescendants(true).toList()
```
slisson, line 22: The algorithm creates too many collections containing all nodes. For large models one collection is already bad for the scalability, but here we have five collections (`allExistingNodes`, `originalIdToSpec`, `originalIdToExisting`, `allNodes` and `originalIdToRef`) which all are about as big as the model.

I think you only need the `originalIdToExistingMap`, which gets updated while new nodes are added.

slisson (changes requested):

slisson (approved):

PR #142, opened 2023-06-19
model-api-gen: added proposal for references in ts gen (untested)

slisson:


slisson:
I changed the commit message and created the new PR #144

PR #147, opened 2023-06-21
MODELIX-465 Ordering of versions in the generated API Reference Landing Page is broken

mhuster23:
After merging this and a new deployment triggered by the next minor release, the ordering will be fixed automatically.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -7,8 +7,8 @@ import org.semver.Version
 
 buildscript {
     dependencies {
-        classpath("org.jetbrains.dokka:versioning-plugin:1.8.10")
-        classpath("org.semver:api:0.9.33")
+        classpath(libs.dokka.versioning)
+        classpath(libs.semver)
     }
 }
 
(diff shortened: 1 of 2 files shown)


build.gradle.kts
```diff
@@ -202,9 +203,22 @@ fun createDocsIndexPage(): String {
                         div("cover") {
                             h2 { +"Available versions:" }
                             div("table") {
+                                val regex = """(\d*).(\d*).(\d*)""".toRegex()
+                                val comparator = Comparator { a: File, b: File ->
```
slisson, line 207: Use `org.semver.Version.parse()` instead of implementing your own comparator. (is part of `org.semver:api:0.9.33`).

Maybe it's also better to move this HTML generating code into a separate file in the `buildSrc` folder.

slisson (changes requested):

build.gradle.kts
mhuster23: I agree, it would be nicer to extract the helper methods in a separate file. But I ran into some issues with gradle when trying to do that.

slisson (approved):

PR #148, opened 2023-06-21
(model-api-gen): fix primitive property and reference link code generator bugs

benedekh:
4cbe09e fixes follow-up bugs of #144 (cc @abstraktor):
- the argument of `LanguageRegistry.INSTANCE.wrapNode` cannot be `undefined`.
- `LanguageRegistry.INSTANCE.wrapNode` returns `ITypedNode`, but we have to cast it to the return type of the method.

683e95e fixes a bug that the `raw_*` fields are not generated for primitive type fields. Instead of storing the raw string values in the `raw_*` fields, I would propose to store them as property value. This way, we do not have to declare and define the `raw_*` fields on the owner classes and interfaces, which makes the code generator simpler.


slisson:
replaced by #149

PR #163, opened 2023-07-13
MODELIX-466 Add an interface to all generated enums enforcing the presentation field

mhuster23:


languitar (approved):

model-api/src/commonMain/kotlin/org/modelix/model/api/IEnum.kt
```diff
@@ -0,0 +1,18 @@
+package org.modelix.model.api
+
+/**
+ * Superinterface for enums generated by model-api-gen
+ */
+interface IEnum {
```
slisson, line 6: The name `IEnum` is a bit too generic. Maybe `IPropertyValueEnum`?

mhuster23: Sure, I also moved the interface from model-api to model-api-gen-runtime, because i think it makes more sense there.

slisson (approved):

PR #164, opened 2023-07-17
docs(model-api): explain local testing

languitar:
Adds a how-to entry for local testing with a local branch.

### DIFF
--- a/docs/global/modules/core/pages/howto/testing-against-model-api.adoc
+++ b/docs/global/modules/core/pages/howto/testing-against-model-api.adoc
@@ -1,0 +1,19 @@
+= How-To test `model-api` client code without a model-server
+:navtitle: Test `model-api` client code without a model-server
+
+If you want to test code that processes `model-api` instances, you can do this in-process without a running (light-) model-server.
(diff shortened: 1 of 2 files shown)


docs/global/modules/core/pages/howto/testing-against-model-api.adoc
```diff
@@ -0,0 +1,17 @@
+= How-To test `model-api` client code without a model-server
+:navtitle: Test `model-api` client code without a model-server
+
+If you want to test code that processes `model-api` instance, you can do this in-processes without a running (light-) model-server.
```
mhuster23, line 4:
```suggestion
If you want to test code that processes `model-api` instances, you can do this in-processes without a running (light-) model-server.
```

mhuster23 (commented):
I think it would be beneficial to make it more explicit, that `C_SomeConcept`, `N_SomeConcept` and `C_OtherConcept` are all classes generated by `model-api-gen`.

mhuster23 (approved):

mhuster23 (approved):

PR #168, opened 2023-07-20
MODELIX-448 bulk-model-sync-gradle: Gradle plugin for syncing models between model server and MPS

mhuster23:
This is an initial version of the bulk-model-sync-gradle plugin.
It does not fully synchronize repositories yet (only models) and it might not scale for larger models.


model-server/src/main/kotlin/org/modelix/model/server/CmdLineArgs.kt
```diff
@@ -40,6 +40,9 @@ internal class CmdLineArgs {
     )
     var schemaInit = false
 
+    @Parameter(names = ["-useroleids"], description = "Use ids for roles instead of simpleNames", converter = BooleanConverter::class)
```
languitar, line 43:
```suggestion
    @Parameter(names = ["-useroleids"], description = "Use IDs for roles instead of simpleNames", converter = BooleanConverter::class)
```

docs/global/modules/core/pages/reference/component-model-sync-gradle.adoc
languitar, line 1: Does using the plugin for syncing mandate to start the model-server with the new `-useroleids` command line switch? If so, that should be documented.

model-server/src/main/kotlin/org/modelix/model/server/CmdLineArgs.kt
languitar, line 43: It would be good to give a short hint when one wants to use this option.

model-sync-gradle-test/build.gradle.kts
```diff
@@ -0,0 +1,119 @@
...
+val modelixCoreVersion: String = file("../version.txt").readText()
+
+version = modelixCoreVersion
+
+repositories {
+    mavenLocal()
+    maven { url = uri("https://repo.maven.apache.org/maven2") }
+    maven { url = uri("https://plugins.gradle.org/m2/") }
+    mavenCentral()
+    maven { url = uri("https://artifacts.itemis.cloud/repository/maven-mps/") }
```
languitar, line 43: To speed up dependency resolution and to avoid leaking names of dependencies to public repositories, it would be a good idea to set up [repository content filtering](https://docs.gradle.org/current/userguide/declaring_repositories.html#sec:repository-content-filtering). Right now, Gradle would try to resolve all dependencies from our maven-mps repo first in all repos mentioned before. That takes a lot of time and also means that mavenCentral will know about all dependencies we want to request from other repos.

model-sync-gradle-test/build.gradle.kts
```diff
@@ -0,0 +1,119 @@
...
+tasks.test {
+    useJUnitPlatform()
+}
+
+kotlin {
+    jvmToolchain(11)
+}
+
+tasks.register("runModelServer", JavaExec::class) {
+    group = "modelix"
```
languitar, line 71:
```suggestion
    group = "modelix"
    description = "Launches a model-server instance to be used as a target for the test. This task will block and is intended to be run in a separate process, apart from the actual test execution".
```

model-sync-gradle-test/ci.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/sh
+
+set -e
+(
```
languitar, line 4: Is there a specific reason for this subshell being started here? Everything is included, so couldn't that also just be the outer script shell instance?

model-sync-gradle-test/ci.sh
```diff
@@ -0,0 +1,37 @@
...
+  fi
+
+  ./gradlew runModelServer --console=plain &
+  MODEL_SERVER_PID=$!
+  sleep 5
+
+  #CI needs more time
+  if [ "${CI}" = "true" ]; then
+    sleep 10
+  fi
```
languitar, line 29: You can make this more reliable by replacing it with [wait-for-it](https://github.com/vishnubob/wait-for-it). That waits for the target port to become available.

model-sync-gradle-test/gradle/wrapper/gradle-wrapper.jar
languitar, line 1: Is it necessary to duplicate this binary file? It could at least be symlinked from the top-level project. Most other subprojects actually don't include a complete gradle wrapper setup at all.

model-sync-gradle-test/graph-lang-api/build.gradle.kts
```diff
@@ -0,0 +1,68 @@
...
+dependencies {
+    mps("com.jetbrains:mps:2021.1.4")
+    api("org.modelix:model-api-gen-runtime:$modelixCoreVersion")
+}
+
+val copyGeneratedApiToSrc by tasks.registering(Sync::class) {
+    dependsOn(tasks.named("generateMetaModelSources"))
+    from(kotlinGenDir)
+    into("src/main/kotlin")
+}
```
languitar, line 30: Why not modifying the sourceSet instead of copying files? I would avoid putting generated files into the actual source tree. That makes it easy to accidentally commit them.

model-sync-gradle-test/graph-lang-api/gradle/wrapper/gradle-wrapper.jar
languitar, line 1: Same as above.

model-sync-gradle-test/src/test/kotlin/org/modelix/model/sync/gradle/test/PullTest.kt
```diff
@@ -0,0 +1,76 @@
...
+import org.junit.jupiter.api.Test
+import java.io.File
+import kotlin.test.assertEquals
+
+class PullTest {
+
+    @Test
+    fun `nodes were synced to local`() {
+        val localModelFile = File("build/test-repo/solutions/GraphSolution/models/GraphSolution.example.mps")
+        val expected = """
```
languitar, line 12: Maybe it would be better to put this into a file that's being read from resources. That way, one would actually have editor support for reading the long-ish XML file.

model-sync-gradle-test/src/test/kotlin/org/modelix/model/sync/gradle/test/PullTest.kt
```diff
@@ -0,0 +1,76 @@
...
+                  <property role="TrG5h" value="D" />
+                </node>
+                <node concept="1DmExO" id="pSCM1J8Fg0" role="1DmyQT">
+                  <property role="TrG5h" value="E" />
+                </node>
+              </node>
+            </model>
+        """.trimIndent()
+
+        assertEquals(expected, localModelFile.readText().trim())
```
languitar, line 74: Is it valid to assume that the exact XML format is always created, without spacing changes etc.? Otherwise I'd propose to use something like [xmlunit](https://www.xmlunit.org/) to be safe from such issues.

model-sync-gradle-test/src/test/kotlin/org/modelix/model/sync/gradle/test/PushTest.kt
```diff
@@ -0,0 +1,76 @@
...
+                .getDescendants(false)
+                .filter { it.getConceptReference() == ConceptReference(_C_UntypedImpl_Node.getUID()) }
+                .map { it.typed<N_Node>() }
+                .toList()
+
+            graphNodes[0].name = "X"
+            graphNodes[1].name = "Y"
+            graphNodes[2].name = "Z"
+        }
+        sleep(5000) // wait for changes to be sent to server
```
languitar, line 74: @slisson told me that things should be propagated by the end of the write transaction. Why is this potentially flaky sleep needed then?

model-sync-gradle-test/src/test/kotlin/org/modelix/model/sync/gradle/test/PushTest.kt
```diff
@@ -0,0 +1,76 @@
...
+        val client = ModelClientV2PlatformSpecificBuilder().url(url).build()
+
+        val branch = runBlocking {
+            client.init()
+            client.getReplicatedModel(branchRef).start()
+        }
+        branch.runRead {
+            branch.getRootNode().allChildren.forEachIndexed { index, child ->
+                assertEquals(inputModel.root.children[index], child.asExported())
+            }
```
languitar, line 56: There's an [assertIterableEquals](https://junit.org/junit5/docs/5.0.1/api/org/junit/jupiter/api/Assertions.html#assertIterableEquals-java.lang.Iterable-java.lang.Iterable-) in JUnit 5. Using that method probably gives better error messages when assertions fail.

model-sync-gradle/src/main/kotlin/org/modelix/model/sync/gradle/ModelSyncGradlePlugin.kt
```diff
@@ -0,0 +1,220 @@
...
+        syncDirection: SyncDirection,
+        project: Project,
+        previousTask: TaskProvider<*>,
+    ) {
+        val baseDir = project.buildDir.resolve("model-sync").apply { mkdirs() }
+        val jsonDir = baseDir.resolve(syncDirection.name).apply { mkdir() }
+        val sourceTask = when (syncDirection.source) {
+            is LocalSource -> registerTasksForLocalSource(syncDirection, project, previousTask, jsonDir)
+            is ServerSource -> registerTasksForServerSource(syncDirection, project, previousTask, jsonDir)
+            else -> previousTask
```
languitar, line 70: This condition would benefit from some comment that explains why this is valid.

model-sync-gradle/src/main/kotlin/org/modelix/model/sync/gradle/ModelSyncGradlePlugin.kt
```diff
@@ -0,0 +1,220 @@
...
+            it.antScriptFile.set(antScript)
+            it.mpsDependenciesPath.set(getDependenciesDir(project).absolutePath)
+            it.jsonDirPath.set(jsonDir.absolutePath)
+            it.exportFlag.set(true)
+            it.includedModules.set(syncDirection.includedModules)
+        }
+
+        val exportFromMps = project.tasks.register("${syncDirection.name}ExportFromMps", ExportFromMps::class.java) {
+            it.dependsOn(generateAntScript)
+            it.outputs.cacheIf { false }
```
languitar, line 124: Can this be optimized later on? In theory we could hash the full source MPS project and use that as a cache key?

model-sync-gradle/src/main/kotlin/org/modelix/model/sync/gradle/ModelSyncGradlePlugin.kt
```diff
@@ -0,0 +1,220 @@
...
+        return project.buildDir.resolve("model-sync")
+    }
+
+    private fun getDependenciesDir(project: Project): File {
+        return getBaseDir(project).resolve("dependencies")
+    }
+
+    private fun readModelixCoreVersion(): String? {
+        val resources: Enumeration<URL>? = javaClass.classLoader.getResources("modelix.core.version.properties")
+        while (resources != null && resources.hasMoreElements()) {
```
languitar, line 214:
```suggestion
        val resources: Enumeration<URL>? = javaClass.classLoader.getResources("modelix.core.version.properties") ?: return null
        if (resources.hasMoreElements()) {
```

There's no need for the loop. The later code only takes the first element if it exists.

model-sync-gradle/src/main/kotlin/org/modelix/model/sync/gradle/config/ModelSyncGradleSettings.kt
```diff
@@ -0,0 +1,162 @@
...
+        (includedModules as MutableSet).add(module)
+    }
+
+    fun registerLanguage(language: ILanguage) {
+        (registeredLanguages as MutableSet).add(language)
+    }
+}
+
+interface SyncEndPoint {
+    fun getValidationErrors(): String
```
languitar, line 64: To avoid the ugly string concatenation logic in the implementations below, it would probably be better to return a list of strings and let the caller decide how to deal with multiple error conditions. A list / iterable would also be easier to understand for a method that has a plural name.

model-sync-gradle/src/main/kotlin/org/modelix/model/sync/gradle/config/ModelSyncGradleSettings.kt
```diff
@@ -0,0 +1,162 @@
...
+            append(super.getValidationErrors())
+            if (revision == null) {
+                if (repositoryId == null && branchName == null) {
+                    appendLine("Invalid server source. Please either specify a revision or repositoryId and branchName.")
+                } else if (repositoryId == null) {
+                    appendUndefinedServerFieldError("repositoryId")
+                } else if (branchName == null) {
+                    appendUndefinedServerFieldError("branchName")
+                }
+            }
```
languitar, line 127:
```suggestion
            if (revision != null) {
                // We do not need anything else if a revision is specified
                return
            }
            
            // Give users a proper hint in case the configuration is completely off
            if (revision == null && repositoryId == null && branchName == null) {
                    appendLine("Invalid server source. Please either specify a revision or repositoryId and branchName.")
                    return
            }
            
            // There are minor configuration issues.
            if (repositoryId == null) {
                    appendUndefinedServerFieldError("repositoryId")
            } else if (branchName == null) {
                    appendUndefinedServerFieldError("branchName")
            }
```

I had a hard time understanding the nested conditions. Maybe this version would be easier to understand?

model-sync-gradle/src/main/kotlin/org/modelix/model/sync/gradle/ModelSyncGradlePlugin.kt
```diff
@@ -0,0 +1,220 @@
...
+        settings = project.extensions.create("modelSync", ModelSyncGradleSettings::class.java)
+        getBaseDir(project).mkdirs()
+
+        project.afterEvaluate {
+            val validateSyncSettings = project.tasks.register("validateSyncSettings", ValidateSyncSettings::class.java) {
+                settings.taskDependencies.forEach { dependency ->
+                    it.dependsOn(dependency)
+                }
+                it.settings.set(settings)
+            }
```
languitar, line 38: Why is this validation done that late? Wouldn't it be possible to immediately fail the build in case things are configured improperly?

model-sync-gradle/src/main/kotlin/org/modelix/model/sync/gradle/tasks/ExportFromModelServer.kt
```diff
@@ -0,0 +1,80 @@
...
+        val client = ModelClientV2PlatformSpecificBuilder().url(url.get()).build().apply { runBlocking { init() } }
+        val branch = if (revision.isPresent) {
+            getBranchByRevision(client)
+        } else {
+            getBranchByRepoIdAndBranch(client)
+        }
+
+        branch.runRead {
+            val root = branch.getRootNode()
+            println("Got root node: $root")
```
languitar, line 57: Logger instead?

model-sync-gradle/src/main/kotlin/org/modelix/model/sync/gradle/tasks/ExportFromModelServer.kt
```diff
@@ -0,0 +1,80 @@
...
+            val outputDir = outputDir.get().asFile
+            root.allChildren.forEach {
+                val outputFile = outputDir.resolve("${it.getPropertyValue(IProperty.fromName(RepositoryLanguage.NamePropertyUID))}.json")
```
languitar, line 60: That's a lot of code inline in the string expansion. Some named variable would help understanding what's done here.

languitar, line 60: I suspect the same JSON file naming convention also resides somewhere in the MPS export implementation? If so, it would probably be good to use a shared implementation that derives the file location. Otherwise, if this ever changes, one of the locations is probably forgotten in the first change set.

model-sync-gradle/src/main/kotlin/org/modelix/model/sync/gradle/tasks/GenerateAntScriptForMps.kt
```diff
@@ -0,0 +1,125 @@
...
+        buildPropertiesFile.inputStream().use { buildProperties.load(it) }
+
+        return listOfNotNull(
+            buildProperties["mpsBootstrapCore.version.major"],
+            buildProperties["mpsBootstrapCore.version.minor"],
+            buildProperties["mpsBootstrapCore.version.eap"],
+        )
+            .map { it.toString().trim('.') }
+            .filter { it.isNotEmpty() }
+            .joinToString(".")
```
languitar, line 123: if only minor would be `null`, this would result in a strange version being returned. Maybe this should be made safer via some assertions?

model-sync-gradle/src/main/kotlin/org/modelix/model/sync/gradle/tasks/ImportIntoModelServer.kt
```diff
@@ -0,0 +1,69 @@
...
+        val repoId = RepositoryId(repositoryId.get())
+
+        val branchRef = ModelFacade.createBranchReference(repoId, branchName.get())
+        val client = ModelClientV2PlatformSpecificBuilder().url(url.get()).build()
+        val branch = runBlocking {
+            client.init()
+            client.getReplicatedModel(branchRef).start()
+        }
+
+        val files = inputDir.listFiles()?.filter { it.extension == "json" } ?: error("no json files found")
```
languitar, line 59: Is that error condition correctly generated? The error would trigger only if `listFiles` returns `null` but not if after filtering, the list is empty. Is that desired?

model-sync-gradle/src/main/kotlin/org/modelix/model/sync/gradle/tasks/ImportIntoModelServer.kt
```diff
@@ -0,0 +1,69 @@
...
+
+        branch.runWrite {
+            val rootNode = branch.getRootNode()
+            println("Got root node: $rootNode")
+            println("Importing...")
+            ModelImporter(branch.getRootNode()).importFilesAsRootChildren(*files.toTypedArray())
+            println("Import finished")
```
languitar, line 66: Gradle logger instead?

model-sync-gradle/src/main/kotlin/org/modelix/model/sync/gradle/tasks/ValidateSyncSettings.kt
```diff
@@ -0,0 +1,92 @@
...
+            if (direction.source == null) {
+                appendLine("Undefined source.")
+            }
+            if (direction.target == null) {
+                appendLine("Undefined target.")
+            }
+
+            if (direction.source == null || direction.target == null) {
+                return@buildString
+            }
```
languitar, line 63: Maybe this can be reduced to this to avoid the duplicate boolean conditions:

```suggestion
            if (direction.source == null || direction.target == null) {
                appendLine("Both source and target have to be defined.")
                return@buildString
            }
```

mps-model-adapters/src/main/kotlin/org/modelix/model/mpsadapters/MPSArea.kt
```diff
@@ -49,23 +51,29 @@ data class MPSArea(val repository: SRepository) : IArea {
...
+        return takeIf { ref == it }
     }
 
     override fun <T> executeRead(f: () -> T): T {
         var result: T? = null
         repository.modelAccess.runReadAction {
             result = f()
         }
-        return result as T
+        return result!!
```
languitar, line 66: Can't result be immediately returned when calling `f()`? That would avoid the bangs.

mps-model-adapters/src/main/kotlin/org/modelix/model/mpsadapters/MPSLanguageRepository.kt
```diff
@@ -0,0 +1,50 @@
...
+import org.modelix.model.api.IConcept
+import org.modelix.model.api.ILanguageRepository
+
+class MPSLanguageRepository(private val repository: SRepository) : ILanguageRepository {
+    override fun resolveConcept(uid: String): IConcept? {
+        if (uid.startsWith("mps:")) return null
+
+        val conceptId = try {
+            SConceptId.deserialize(uid.substring(4))
+        } catch (e: Exception) { return null } ?: return null
```
languitar, line 34: Maybe at least some logging in case of such a generic catch and also a comment why that's needed?

mps-model-adapters/src/main/kotlin/org/modelix/model/mpsadapters/MPSModuleAsNode.kt
```diff
@@ -116,7 +117,7 @@ data class MPSModuleAsNode(val module: SModule) : IDeprecatedNodeDefaults {
     }
 
     override fun setPropertyValue(property: IProperty, value: String?) {
-        TODO("Not yet implemented")
+        // TODO("Not yet implemented")
```
languitar, line 120: Why this?

languitar (commented):
Some ideas during a first look

mps-model-adapters/src/main/kotlin/org/modelix/model/mpsadapters/MPSArea.kt
mhuster23: return is not allowed inside of runReadAction according to the kotlin compiler. Only `return@runReadAction`, but this would not help in this case.

model-sync-gradle-test/gradle/wrapper/gradle-wrapper.jar
languitar: Ok, then let's avoid adding another binary blob by using a symlink.

model-sync-gradle-test/graph-lang-api/build.gradle.kts
mhuster23: The generated api is the only source code for that project a separate source set is not necessary. Accidentally committing files should be quite hard, as this src dir is part of the .gitignore file.

model-sync-gradle-test/src/test/kotlin/org/modelix/model/sync/gradle/test/PushTest.kt
mhuster23: Changes are pushed asynchronously to the server. If we don't wait, the process stops before the changes are propagated. I will clarify the comment.

languitar: I'd say that's an unfortunate implementation. At least an optional explicit waiting would be needed here to make such a test (real application) reliable.

model-sync-gradle-test/src/test/kotlin/org/modelix/model/sync/gradle/test/PullTest.kt
languitar: And we are sure that the formatting doesn't change on the next MPS bump?

mps-model-adapters/build.gradle.kts
```diff
@@ -3,14 +3,15 @@ plugins {
...
-val mpsVersion = project.findProperty("mps.version")?.toString().takeIf { !it.isNullOrBlank() } ?: "2020.3.6"
+val mpsVersion = project.findProperty("mps.version")?.toString().takeIf { !it.isNullOrBlank() } ?: "2021.1.4"
 
 dependencies {
     api(project(":model-api"))
 
-    compileOnly("com.jetbrains:mps-openapi:$mpsVersion")
-    compileOnly("com.jetbrains:mps-core:$mpsVersion")
-    compileOnly("com.jetbrains:mps-environment:$mpsVersion")
+    implementation("com.jetbrains:mps-openapi:$mpsVersion")
```
mhuster23, line 11: I'm still debating if changing this to implementation was the right decision.
An alternative would be to keep this declared as compileOnly but then the user would have to add `mps-openapi` and `mps-core` buildscript and implementation dependencies in their mps version manually as it's needed for `RepositoryLanguage`, which would be unintuitive.

.github/workflows/build.yaml
```diff
@@ -28,12 +28,7 @@ jobs:
           arguments: |
             --build-cache
             build
-            publishToMavenLocal
```
languitar, line 31: What I though about, when proposing to put things into separate jobs that run in parallel, was to still do the publishing here and then put the `~.m2` folder into a job artifact that's used by the downstream jobs. The changes right now result in building larger parts of the core three times due to the inefficient build cache.

bulk-model-sync-gradle-test/.gitignore
```diff
@@ -0,0 +1,49 @@
...
+
+### Mac OS ###
+.DS_Store
+/test-repo/languages/GraphLang/classes_gen/
+/test-repo/languages/GraphLang/generator/
+/test-repo/languages/GraphLang/source_gen/
+/test-repo/languages/GraphLang/source_gen.caches/
+/test-repo/solutions/GraphSolution/source_gen/
+/test-repo/solutions/GraphSolution/classes_gen/
+/test-repo/solutions/GraphSolution/source_gen.caches/
```
languitar, line 49: This could be reduced with some wildcards. Moreover, that would then also be resilient against language renamings.

bulk-model-sync-gradle-test/build.gradle.kts
```diff
@@ -0,0 +1,123 @@
...
+    from(projectDir.resolve("test-repo"))
+    into(repoDir)
+}
+
+modelSync {
+    dependsOn(resolveMps)
+    dependsOn(copyTestRepo)
+    direction("testPush") {
+        registerLanguage(L_GraphLang)
+        registerLanguage(L_jetbrains_mps_lang_core)
```
languitar, line 99: How would this work when many languages are used? Can the registration helper being generated by the generator plugin be used here?

bulk-model-sync-gradle-test/ci.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/sh
+
+set -e
+
+cd "$(dirname "$0")"
+
+(
+  cd graph-lang-api
+  ./gradlew publishToMavenLocal --console=plain
```
languitar, line 9: Here's some more potential for speed up in CI. If the build was performed by the GitHub Action with enable build cache, then recompiling / regenerating on every CI execution could be avoided.

bulk-model-sync-gradle-test/gradlew
languitar, line 1: Could also be symlinked from the main project.

bulk-model-sync-gradle-test/gradlew.bat
languitar, line 1: Symlink?

bulk-model-sync-gradle-test/graph-lang-api/gradlew
languitar, line 1: Symlink?

bulk-model-sync-gradle-test/graph-lang-api/gradlew.bat
languitar, line 1: Symlink?

bulk-model-sync-gradle-test/src/test/kotlin/org/modelix/model/sync/gradle/test/PullTest.kt
```diff
@@ -0,0 +1,32 @@
...
+            |\s*</node>
+            |\s*<node concept="1DmExO" id="pSCM1J8FfZ" role="1DmyQT">
+            |\s*<property role="TrG5h" value="D" />
+            |\s*</node>
+            |\s*<node concept="1DmExO" id="pSCM1J8Fg0" role="1DmyQT">
+            |\s*<property role="TrG5h" value="E" />
+            |\s*</node>
+        """.trimMargin().toRegex()
+
+        assert(expectedNodesRegex.containsMatchIn(localModel))
```
languitar, line 30: I'd be more in favor of an XPath solution that's resilient against slight formatting changes.

bulk-model-sync-gradle-test/src/test/kotlin/org/modelix/model/sync/gradle/test/PushTest.kt
```diff
@@ -0,0 +1,74 @@
...
+
+        val branch = runBlocking {
+            client.init()
+            client.getReplicatedModel(branchRef).start()
+        }
+        branch.runRead {
+            assertContentEquals(inputModel.root.children, branch.getRootNode().allChildren.map { it.asExported() })
+        }
+
+        applyChangesForPullTest(branch)
```
languitar, line 57: Does this mean the pull test depends on the previous execution of the push test? Can't we break that up. Or if that's really necessary, let's at least express this clearly in the tests by using [test method ordering](https://www.baeldung.com/junit-5-test-order).

bulk-model-sync-gradle-test/src/test/kotlin/org/modelix/model/sync/gradle/test/PushTest.kt
```diff
@@ -0,0 +1,74 @@
...
+                .getDescendants(false)
+                .filter { it.getConceptReference() == ConceptReference(_C_UntypedImpl_Node.getUID()) }
+                .map { it.typed<N_Node>() }
+                .toList()
+
+            graphNodes[0].name = "X"
+            graphNodes[1].name = "Y"
+            graphNodes[2].name = "Z"
+        }
+        sleep(5000) // changes are pushed asynchronously to the server. wait for the propagation
```
languitar, line 72: Can successful propagation be tested by querying the server? If so, the sleep could be replaced with a polling loop

bulk-model-sync-gradle/src/main/kotlin/org/modelix/model/sync/bulk/gradle/ModelSyncGradlePlugin.kt
```diff
@@ -0,0 +1,234 @@
...
+        getBaseDir(project).mkdirs()
+
+        project.afterEvaluate {
+            val validateSyncSettings = project.tasks.register("validateSyncSettings", ValidateSyncSettings::class.java) {
+                settings.taskDependencies.forEach { dependency ->
+                    it.dependsOn(dependency)
+                }
+                it.settings.set(settings)
+            }
+            val modelixCoreVersion = readModelixCoreVersion() ?: throw RuntimeException("modelix.core version not found")
```
languitar, line 53: We could probably give the user of the plugin a better hint if there's anything he can do to avoid this exception of if this should have never happened.

bulk-model-sync-gradle/src/main/kotlin/org/modelix/model/sync/bulk/gradle/ModelSyncGradlePlugin.kt
```diff
@@ -0,0 +1,234 @@
...
+            val antDependencies = project.configurations.create("model-sync-ant-dependencies")
```
languitar, line 54: The name of the configuration is duplicated in this file. You can put in in a constant.

bulk-model-sync-gradle/src/main/kotlin/org/modelix/model/sync/bulk/gradle/ModelSyncGradlePlugin.kt
```diff
@@ -0,0 +1,234 @@
...
+        val jsonDir = baseDir.resolve(syncDirection.name).apply { mkdir() }
+        val sourceTask = when (syncDirection.source) {
+            is LocalSource -> registerTasksForLocalSource(syncDirection, project, previousTask, jsonDir)
+            is ServerSource -> registerTasksForServerSource(syncDirection, project, previousTask, jsonDir)
+            else -> error("Unknown sync direction source")
+        }
+
+        when (syncDirection.target) {
+            is LocalTarget -> registerTasksForLocalTarget(syncDirection, project, sourceTask, jsonDir)
+            is ServerTarget -> registerTasksForServerTarget(syncDirection, project, sourceTask, jsonDir)
```
languitar, line 89: If the targets were provided to the `register*` method as arguments, the need for casting inside those methods could be eliminated.

bulk-model-sync-gradle/src/main/kotlin/org/modelix/model/sync/bulk/gradle/config/ModelSyncGradleSettings.kt
```diff
@@ -0,0 +1,187 @@
...
+    fun includeModule(module: String) {
+        (includedModules as MutableSet).add(module)
+    }
+
+    fun registerLanguage(language: ILanguage) {
+        (registeredLanguages as MutableSet).add(language)
+    }
+}
+
+interface SyncEndPoint {
```
languitar, line 79:
```suggestion
interface SyncEndpoint {
```

Endpoint seems to be spelled as a single word.

bulk-model-sync-gradle/src/main/kotlin/org/modelix/model/sync/bulk/gradle/tasks/ExportFromModelServer.kt
```diff
@@ -0,0 +1,101 @@
...
+
+        branch.runRead {
+            val root = branch.getRootNode()
+            logger.info("Got root node: {}", root)
+            val outputDir = outputDir.get().asFile
+            root.allChildren.forEach {
+                val nameRole = IProperty.fromName(RepositoryLanguage.NamePropertyUID)
+                val fileName = it.getPropertyValue(nameRole) ?: it.getPropertyValue(IProperty.fromName(NodeData.idPropertyKey))
+                    ?.substringBeforeLast(")", missingDelimiterValue = "null")
+                    ?.substringAfterLast("(", missingDelimiterValue = "null") // fallback if module name is not in properties
```
languitar, line 80: That's pretty hard to read. Maybe a method or some intermediate variables explaining the pipeline steps could help understanding what's done here to the node to derive the filename

bulk-model-sync-gradle/src/main/kotlin/org/modelix/model/sync/bulk/gradle/tasks/ExportFromModelServer.kt
```diff
@@ -0,0 +1,101 @@
...
+
+    @Input
+    @Optional
+    val revision: Property<String> = of.property(String::class.java)
+
+    @OutputDirectory
+    val outputDir: DirectoryProperty = of.directoryProperty()
+
+    @TaskAction
+    fun export() {
```
languitar, line 64: Why isn't the whole method being executed in a big `runBlocking` statement? Right now, these ugly things are somewhere in every detail method.

bulk-model-sync-gradle/src/main/kotlin/org/modelix/model/sync/bulk/gradle/tasks/GenerateAntScriptForMps.kt
```diff
@@ -0,0 +1,141 @@
...
+
+    private fun getAntLibs(mpsHome: File): List<String> {
+        val mpsVersion = getMpsVersion(mpsHome)
+        return if (mpsVersion < "2021.2") {
+            listOf("lib/ant/lib/ant-mps.jar", "lib/log4j.jar", "lib/jdom.jar")
+        } else if (mpsVersion < "2021.3") {
+            listOf("lib/ant/lib/ant-mps.jar", "lib/util.jar")
+        } else {
+            listOf("lib/ant/lib/ant-mps.jar", "lib/util.jar", "lib/3rd-party-rt.jar")
+        }
```
languitar, line 123: Can you add some explanation for each of the conditions on why the exact combination was chosen. If there was ever a bug affecting this code, people would easily be lost on why the current solution was used.

bulk-model-sync-lib/src/jvmMain/kotlin/org/modelix/model/sync/bulk/PlatformSpecific.kt
```diff
@@ -0,0 +1,42 @@
...
+ * based on the [ModelData] specification contained in the given file.
+ *
+ * @param jsonFile json file containing the model specification
+ *
+ * @throws IllegalArgumentException if the file is not a json file or the file does not exist.
+ */
+@JvmName("importFile")
+fun ModelImporter.import(jsonFile: File) {
+    require(jsonFile.exists())
+    require(jsonFile.extension == "json")
```
languitar, line 33: Why is that condition required? Would work with any file extension, right?

bulk-model-sync-solution/build.gradle.kts
```diff
@@ -0,0 +1,42 @@
...
+    alias(libs.plugins.modelix.mps.buildtools)
+}
+
+group = "org.modelix.mps"
+
+val generatorLibs: Configuration by configurations.creating
+
+dependencies {
+    "generatorLibs"(project(":bulk-model-sync-lib"))
+    generatorLibs(project(":mps-model-adapters"))
```
languitar, line 14: Why are two different styles used here?

docs/global/modules/core/pages/reference/component-bulk-model-sync-gradle.adoc
```diff
@@ -0,0 +1,145 @@
...
+--
+https://github.com/modelix/modelix.core[Repository^] | https://github.com/modelix/modelix.core/blob/main/model-sync-gradle/build.gradle.kts[buildfile^]
+--
+
+== Overview
+
+The `bulk-model-sync` Gradle plugin synchronizes models between `model-api` endpoints.
+A common example would be synchronizing a local MPS project to a `model-server`.
+
+IMPORTANT: For correct synchronization the used model-server is required to use IDs for roles. This can be achieved by starting the model-server with the `-useroleids` flag.
```
languitar, line 15:
```suggestion
IMPORTANT: For correct synchronization the used model-server is required to use IDs for roles.
This can be achieved by starting the model-server with the `-useroleids` flag.
```

That way the sentence per line style continues.

docs/global/modules/core/pages/reference/component-bulk-model-sync-gradle.adoc
```diff
@@ -0,0 +1,145 @@
...
+
+.Bulk Synchronization between MPS and model-server
+image::bulk-model-sync-gradle.overview.png[Bulk Synchronization between MPS and model-server]
+
+It allows the definition of sync directions by specifying source and target endpoint.
```
languitar, line 20: What is "It"?

docs/global/modules/core/pages/reference/component-bulk-model-sync-gradle.adoc
```diff
@@ -0,0 +1,145 @@
...
+        toModelServer {
+            url = "http://0.0.0.0:28101/v2"
+            repositoryId = "my-repo"
+            branchName = "dev"
+        }
+    }
+}
+--
+
+Generated gradle task to perform synchronization: `runSyncPushToMyServer`.
```
languitar, line 145:
```suggestion
Generated Gradle task to perform synchronization: `runSyncPushToMyServer`.
```

mps-model-adapters/src/main/kotlin/org/modelix/model/mpsadapters/MPSLanguageRepository.kt
languitar: How was this resolved?

mps-model-adapters/src/main/kotlin/org/modelix/model/mpsadapters/MPSRepositoryAsNode.kt
```diff
@@ -78,34 +81,34 @@ data class MPSRepositoryAsNode(val repository: SRepository) : IDeprecatedNodeDef
...
     }
 
     override fun getPropertyLinks(): List<IProperty> {
-        TODO("Not yet implemented")
+        return emptyList()
     }
 
     override fun getReferenceLinks(): List<IReferenceLink> {
-        TODO("Not yet implemented")
+        return emptyList()
```
languitar, line 112: These changes should probably somehow preserve the information that the methods are actually not implemented using comments?

languitar (commented):
I can't really review all aspects of this large PR, especially the MPS side. But here's what I've still found for the general usage and kotlin code.

mps-model-adapters/src/main/kotlin/org/modelix/model/mpsadapters/MPSLanguageRepository.kt
```diff
@@ -0,0 +1,50 @@
...
+import jetbrains.mps.smodel.language.ConceptRegistry
+import jetbrains.mps.smodel.language.LanguageRegistry
+import jetbrains.mps.smodel.runtime.illegal.IllegalConceptDescriptor
+import org.jetbrains.mps.openapi.module.SRepository
+import org.modelix.model.api.IConcept
+import org.modelix.model.api.ILanguageRepository
+
+class MPSLanguageRepository(private val repository: SRepository) : ILanguageRepository {
+    override fun resolveConcept(uid: String): IConcept? {
+        if (uid.startsWith("mps:")) return null
```
slisson, line 30:
```suggestion
        if (!uid.startsWith("mps:")) return null
```

mps-model-adapters/src/main/kotlin/org/modelix/model/mpsadapters/MPSModelAsNode.kt
```diff
@@ -99,13 +103,15 @@ data class MPSModelAsNode(val model: SModel) : IDeprecatedNodeDefaults {
...
+        } else if (property.getSimpleName() == NodeData.idPropertyKey) {
+            model.modelId.toString()
         } else {
             null
         }
     }
 
     override fun setPropertyValue(property: IProperty, value: String?) {
-        TODO("Not yet implemented")
+        // TODO("Not yet implemented")
```
slisson, line 114:
```suggestion
        if (getPropertyValue(property) != value) {
            throw UnsupportedOperationException("Property $property of $concept is read-only")
        }
```

mps-model-adapters/src/main/kotlin/org/modelix/model/mpsadapters/MPSNodeReference.kt
```diff
@@ -37,3 +45,16 @@ fun INodeReference.toMPSNodeReference(): MPSNodeReference {
...
+object MPSNodeReferenceSerializer : INodeReferenceSerializerEx {
+    override val prefix = "mps"
+    override val supportedReferenceClasses: Set<KClass<out INodeReference>> = setOf(MPSNodeReference::class)
+
+    override fun serialize(ref: INodeReference): String {
+        return (ref as MPSNodeReference).ref.nodeId.toString()
+    }
+
+    override fun deserialize(serialized: String): INodeReference {
+        TODO("Not yet implemented")
```
slisson, line 58:
```suggestion
        return MPSNodeReference(SNodePointer.deserialize(serialized))
```

mps-model-adapters/src/main/kotlin/org/modelix/model/mpsadapters/MPSRepositoryAsNode.kt
```diff
@@ -78,34 +81,34 @@ data class MPSRepositoryAsNode(val repository: SRepository) : IDeprecatedNodeDef
...
     }
 
     override fun setReferenceTarget(link: IReferenceLink, target: INode?) {
-        TODO("Not yet implemented")
+        return
     }
 
     override fun setReferenceTarget(role: IReferenceLink, target: INodeReference?) {
-        TODO("Not yet implemented")
+        return
```
slisson, line 92:
```suggestion
        if (target != null) {
            throw IllegalArgumentException("$concept doesn't contain a reference link $link")
        }
```

mps-model-adapters/src/main/kotlin/org/modelix/model/mpsadapters/MPSRepositoryAsNode.kt
```diff
@@ -78,34 +81,34 @@ data class MPSRepositoryAsNode(val repository: SRepository) : IDeprecatedNodeDef
     }
 
     override fun getReferenceTarget(link: IReferenceLink): INode? {
-        TODO("Not yet implemented")
+        return null
     }
 
     override fun setReferenceTarget(link: IReferenceLink, target: INode?) {
-        TODO("Not yet implemented")
+        return
```
slisson, line 88:
```suggestion
        if (target != null) {
            throw IllegalArgumentException("$concept doesn't contain a reference link $role")
        }
```

mps-model-adapters/src/main/kotlin/org/modelix/model/mpsadapters/MPSRepositoryAsNode.kt
```diff
@@ -78,34 +81,34 @@ data class MPSRepositoryAsNode(val repository: SRepository) : IDeprecatedNodeDef
...
+        return null
     }
 
     override fun getPropertyValue(property: IProperty): String? {
-        TODO("Not yet implemented")
+        return null
     }
 
     override fun setPropertyValue(property: IProperty, value: String?) {
         TODO("Not yet implemented")
```
slisson, line 104:
```suggestion
        if (value != null) {
            throw IllegalArgumentException("$concept doesn't contain a property $property")
        }
```

mps-model-adapters/build.gradle.kts
mhuster23: The only exception is `RepositoryLanguage`. That's an object one wants to use outside of MPS. For example to register it. But it also has dependencies to these jars. `SConceptAdapterByID` and `SInterfaceConceptAdapterByID` are required to initialize `BaseConcept` and `INamedConcept`.
Could we somehow get rid of those dependencies?

bulk-model-sync-gradle-test/src/test/kotlin/org/modelix/model/sync/gradle/test/PushTest.kt
mhuster23: These tests are not meant to be run together, they perform checks during different stages of the integration test.
They are only run as part of the `ci.sh` test script, which orchestrates the integration test.

bulk-model-sync-solution/build.gradle.kts
mhuster23: This must have been caused by IntelliJ while renaming the component.

bulk-model-sync-gradle-test/src/test/kotlin/org/modelix/model/sync/gradle/test/PushTest.kt
languitar: So is there a dependency between the tests or not? ;-)

mps-model-adapters/src/main/kotlin/org/modelix/model/mpsadapters/MPSRepositoryAsNode.kt
languitar: Oh right, then maybe a test case verifying these conditions or some comment would help to avoid this confusion the next time. The implementation locks very much like a stub.

.github/workflows/build.yaml
mhuster23: I'll keep it as it is for now, since it's still faster than the current solution.
Building 3 times shouldn't be that expensive once we get the cache working in a followup ticket.

languitar: The only thing is that this script executed this way might not pick up the cache.

bulk-model-sync-gradle/src/main/kotlin/org/modelix/model/sync/bulk/gradle/tasks/ExportFromModelServer.kt
slisson: It's an ugly hack that makes assumptions about the format of the ID. The export should not be MPS specific.

bulk-model-sync-gradle/src/main/kotlin/org/modelix/model/sync/bulk/gradle/tasks/GenerateAntScriptForMps.kt
slisson: It's just copy pasted from the api-gen gradle plugin. https://issues.modelix.org/issue/MODELIX-542/Deduplicate-GenerateAntScriptForMps

docs/global/modules/core/images/bulk-model-sync-gradle.overview.png
slisson, line 1: When I ever have to update this diagram, where do I find the source code?

slisson:
I'm done with the review and fixed most of what I found myself. I created [MODELIX-542](https://issues.modelix.org/issue/MODELIX-542/Deduplicate-GenerateAntScriptForMps) to not let this PR grow to much.

bulk-model-sync-gradle/src/main/kotlin/org/modelix/model/sync/bulk/gradle/tasks/GenerateAntScriptForMps.kt
languitar: That still makes it hard to understand why different libraries are used. ;-)

docs/global/modules/core/images/bulk-model-sync-gradle.overview.png
mhuster23: it's in the modelix drive

bulk-model-sync-gradle/src/main/kotlin/org/modelix/model/sync/bulk/gradle/ModelSyncGradlePlugin.kt
mhuster23: This would just move the cast to the invocation.
Smart cast is unavailable because the field is mutable.

bulk-model-sync-gradle-test/ci.sh
mhuster23: Seems like caching would still work like this, as the build action is used in a prior action, see [here](https://blog.gradle.org/gh-actions)

bulk-model-sync-gradle/src/main/kotlin/org/modelix/model/sync/bulk/gradle/ModelSyncGradlePlugin.kt
languitar: You can assign it to a `val` in the function to avoid that problem.

bulk-model-sync-lib/src/jvmMain/kotlin/org/modelix/model/sync/bulk/PlatformSpecific.kt
mhuster23: Yes, as long as the format is json. I put it in as a safeguard, but we can give the user more freedom here.

bulk-model-sync-gradle-test/src/test/kotlin/org/modelix/model/sync/gradle/test/PushTest.kt
mhuster23: I'm not sure. We should have a method that waits until the changes are propagated, but this is out of scope for this PR

bulk-model-sync-gradle/src/main/kotlin/org/modelix/model/sync/bulk/gradle/tasks/ExportFromModelServer.kt
mhuster23: Raising the indent level for everything just because of one coroutine call sounds pretty ugly to me. But I will improve the readability for this case in another way.

bulk-model-sync-gradle-test/src/test/kotlin/org/modelix/model/sync/gradle/test/PushTest.kt
languitar: For that we should have a follow up ticket and reference it here. More reliable waiting would still be an option to perform here.

bulk-model-sync-gradle/src/main/kotlin/org/modelix/model/sync/bulk/gradle/tasks/ExportFromModelServer.kt
languitar: Seems to be tha multiple coroutines are called? Otherwise there wouldn't be the need for multiple runBlockings in called methods?

mhuster23:
I integrated all the suggestions. Please have a final look.

bulk-model-sync-gradle-test/build.gradle.kts
```diff
@@ -0,0 +1,120 @@
...
+    maven { url = uri("https://artifacts.itemis.cloud/repository/maven-mps/") }
+}
+
+val mps by configurations.creating
+val mpsDir = buildDir.resolve("mps").apply { mkdirs() }
+val kotlinGenDir = buildDir.resolve("metamodel/kotlin").apply { mkdirs() }
+
+dependencies {
+    mps("com.jetbrains:mps:2021.2.5")
+    implementation("org.jetbrains.kotlinx:kotlinx-coroutines-core:1.7.2")
```
languitar, line 50: Woudn't it make sense to put this into the version catalog, too?

mps-model-adapters/src/main/kotlin/org/modelix/model/mpsadapters/MPSRepositoryAsNode.kt
```diff
@@ -78,34 +82,40 @@ data class MPSRepositoryAsNode(val repository: SRepository) : IDeprecatedNodeDef
...
     override fun getPropertyValue(property: IProperty): String? {
-        TODO("Not yet implemented")
+        return null
     }
 
     override fun setPropertyValue(property: IProperty, value: String?) {
-        TODO("Not yet implemented")
+        if (value != null) {
+            throw IllegalArgumentException("$concept doesn't contain a property $property")
+        }
```
languitar, line 111: Why is setting a `null` value to any non-existing property valid? I would have expected the same exception or an comment explaining the strange case.

languitar (approved):
I've found a few bits and pieces but nothing serious.

mps-model-adapters/src/main/kotlin/org/modelix/model/mpsadapters/MPSRepositoryAsNode.kt
slisson: Because only then its symmetric to `getPropertyValue`. It would be unexpected for setPropertyValue to throw an exception when writing `setPropertyValue(property, getPropertyValue(property))`.

languitar: That means there's no chance to disambiguate between an unknown property and a known one without a value from the reaction of `getPropertyValue`? Is that a good thing?

slisson: INode is not responsible for enforcing the conformance to the meta model. Usually, implementations of INode are generic and store property values in a simple map. This MPSRepositoryAsNode is just a special case that cannot store any properties. Maybe the error message is a bit misleading. It would be legal to set a property that is not defined in the concept, but the implementation is just not capable of storing it.

bulk-model-sync-gradle-test/build.gradle.kts
mhuster23: Will do that in the next PR.

PR #175, opened 2023-07-27
MODELIX-475: Let dependabot add reviewers

nkoester:


languitar (commented):
Why do we do this only for dependabot. CODEOWNERS would also assign us automatically for PRs created by everyone else.

.github/CODEOWNERS
languitar, line 1: This CODEOWNERS file contains errors

    Invalid owner on line 1:

    *       @modelix/all/developers

languitar (changes requested):

languitar (approved):

PR #189, opened 2023-08-08
mps-model-adapters: fixed incompatibility with MPS 2021.2

slisson:
It's actually just an issue of the Kotlin compiler and the bytecode would have been compatible.
It's still good to have the workflow that detects possible incompatibilities.

### DIFF
--- a/.github/workflows/mps-model-adapters.yaml
+++ b/.github/workflows/mps-model-adapters.yaml
@@ -1,0 +1,34 @@
+name: MPS compatibility
+
+on:
+  push:
+    branches:
+      - '**'
+  pull_request: {}
+  # allow manual execution just in case
+  workflow_dispatch:
+
+jobs:
+  build-mps-model-adapters:
+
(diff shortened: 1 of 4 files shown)


.github/workflows/mps-model-adapters.yaml
```diff
@@ -0,0 +1,34 @@
+name: MPS compatibility
+
+on:
+  push:
+    branches:
+      - '**'
```
mhuster23, line 6:
```suggestion
      - 'main'
```
`**` will cause the workflow to trigger twice on pushes in PRs

mhuster23 (changes requested):

.github/workflows/mps-model-adapters.yaml
slisson: There is an action `skip-duplicate-actions` that avoids that and still allows to build branches without having to create a PR.

PR #191, opened 2023-08-08
MODELIX-461: enable ktlint in modelix core

languitar:
This PR enables ktlint and a few other linters across the whole code base and all subprojects. Instead of the previous state with selective application for a set of subprojects, now all projects are linted and the code style is enforced.

The primary means of enforcing the linters is through pre-commit. That way, we have a unified mechanism that can be applied locally and in CI and that already prevents invalid code from leaking into commits.

As there was a partial Gradle setup for ktlint, this setup has been retained and extended to all subprojects with the exception of a few JS/TS files where I wasn't sure if adapting the code style (and namings) can be done without side effects.

To facilitate the review of this PR, all configuration changes are in individual commits. Only the last commit in this PR actually adapts the code style of the whole project to make it compliant.


slisson (approved):

slisson:
I get this output when I run `pre-commit run`:

                   
```
trim trailing whitespace.............................(no files to check)Skipped
fix end of files.....................................(no files to check)Skipped
check toml...........................................(no files to check)Skipped
check yaml...........................................(no files to check)Skipped
check for added large files..........................(no files to check)Skipped
check for merge conflicts............................(no files to check)Skipped
check for broken symlinks............................(no files to check)Skipped
fix utf-8 byte order marker..........................(no files to check)Skipped
mixed line ending....................................(no files to check)Skipped
trim trailing whitespace.............................(no files to check)Skipped
KTLint...............................................(no files to check)Skipped
```

Is there something missing in the configuration?

PR #192, opened 2023-08-09
fix(model-api-gen): disable ModelQL generation by default

slisson:
The generated code requires a dependency on the ModelQL runtime which breaks existing projects.
You now have to specify a 'modelqlKotlinDir' to enable the generator which is also useful if you have to split the generated API classes into a separate subproject because of Kotlin compiler memory usage.


languitar (changes requested):
The fix looks good. But the new option needs documentation in https://github.com/modelix/modelix.core/blob/main/docs/global/modules/core/pages/reference/component-model-api-gen-gradle.adoc

mhuster23 (approved):

PR #197, opened 2023-08-09
fix: MODELIX-481 rename docker image modelix/modelix-model to modelix/model-server

slisson:
Also updated the description at
https://hub.docker.com/repository/docker/modelix/modelix-model/general

The old tag is still published to avoid breaking existing projects, but new projects should use the new name.
At least our helm chart should be updated soon: https://github.com/modelix/modelix/blob/46bf6ccb902050ae7035ef07b81bb4180213a66c/helm/modelix/templates/common/model-deployment.yaml#L31

### DIFF
--- a/docker-build-model.sh
+++ b/docker-build-model.sh
@@ -8,9 +8,12 @@ TAG=$( ./modelix-version.sh )
   cd model-server
   if [ "${CI}" = "true" ]; then
     docker buildx build --platform linux/amd64,linux/arm64 --push \
+    -t modelix/model-server:latest -t "modelix/model-server:${TAG}"
     -t modelix/modelix-model:latest -t "modelix/modelix-model:${TAG}" .
   else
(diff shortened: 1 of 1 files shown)


languitar:
> https://hub.docker.com/repository/docker/modelix/modelix-model/general

https://hub.docker.com/repository/docker/modelix/modelix-model/general I can't open that URL without a login. Is that expected?

languitar (commented):
Where is the push actually happening?

slisson:
> https://hub.docker.com/repository/docker/modelix/modelix-model/general I can't open that URL without a login. Is that expected?

I guess that's a requirement of docker hub, but it's publicly visible to any user.

> Where is the push actually happening?

There is `--push` argument.

languitar:
https://hub.docker.com/r/modelix/modelix-model also shows no README that would redirect me towards the new image.

languitar:
> There is a `--push` argument.

oops, didn't see that one.

languitar (approved):

PR #200, opened 2023-08-10
MODELIX-467 Caching in model-api-gen does not add new concepts on rebuild

odzhychko:


model-api-gen-gradle/src/main/kotlin/org/modelix/metamodel/gradle/GenerateAntScriptForMpsMetaModelExport.kt
```diff
@@ -101,8 +98,8 @@ abstract class GenerateAntScriptForMpsMetaModelExport @Inject constructor(of: Ob
         """.trimIndent())
     }
 
-    private fun getMpsBuildPropertiesFile() = mpsHome.get().asFile.resolve("build.properties")
-    private fun getMpsLanguagesDir() = mpsHome.get().asFile.resolve("languages")
+    private fun getMpsBuildPropertiesFile() = File(mpsHome.get()).resolve("build.properties")
+    private fun getMpsLanguagesDir() = File(mpsHome.get()).resolve("languages")
```
languitar, line 102: Why are these changes necessary?

model-api-gen-gradle/src/main/kotlin/org/modelix/metamodel/gradle/GenerateAntScriptForMpsMetaModelExport.kt
```diff
@@ -1,31 +1,28 @@
...
     val antScriptFile: RegularFileProperty = of.fileProperty()
 
-    @InputDirectory
-    @PathSensitive(PathSensitivity.RELATIVE)
+    @Input
     val exporterDir: Property<String> = of.property(String::class.java)
 
-    @InputFiles
-    @PathSensitive(PathSensitivity.RELATIVE)
+    @Input
```
languitar, line 25: What's the reason behind these changes?

odzhychko: The task `GenerateAntScriptForMpsMetaModelExport` had directories and the containing files as inputs.

But generating the Ant script only needed the directory names as inputs. Therefore, it was executed more often the necessary.

model-api-gen-gradle/src/main/kotlin/org/modelix/metamodel/gradle/GenerateAntScriptForMpsMetaModelExport.kt
odzhychko: That was necessary, because `mpsHome` became as `Property<String>` instead of a `DirectoryProperty`,
`DirectoryProperty` can not be used with  the `InputDirectory` annotation.

languitar: Right, overlooked that.

model-api-gen-gradle/src/main/kotlin/org/modelix/metamodel/gradle/GenerateAntScriptForMpsMetaModelExport.kt
odzhychko: `@InputFiles`/`@ InputDirectory` marks the file/directory name and the content of the file/directory as input . 
But the task does not use the content of the directories and files.

> Note: To make the task dependent on the directory's location but not its contents, expose the path of the directory as an [Input](https://docs.gradle.org/current/javadoc/org/gradle/api/tasks/Input.html) property instead.

see. https://docs.gradle.org/current/javadoc/org/gradle/api/tasks/InputDirectory.html

> This will cause the task to be considered out-of-date when the property has changed. This annotation cannot be used on a File object. If you want to refer to the file path, independently of its contents, r**eturn a String instead which returns the absolute path**. If, instead, you want to refer to the contents and path of a file or a directory, use [InputFile](https://docs.gradle.org/current/javadoc/org/gradle/api/tasks/InputFile.html) or [InputDirectory](https://docs.gradle.org/current/javadoc/org/gradle/api/tasks/InputDirectory.html) respectively.

see also. https://docs.gradle.org/current/javadoc/org/gradle/api/tasks/Input.html

languitar (approved):

PR #201, opened 2023-08-10
MODELIX-501 Replace ts-model-api

slisson:


slisson:
Replaced by https://github.com/modelix/modelix.core/pull/319

PR #208, opened 2023-08-17
fix(model-server-lib): increase websocket timeouts

languitar:
The existing timeouts sometimes caused disconnects when requesting larger amounts of data despite everything being functional.

### DIFF
--- a/model-server-lib/src/main/kotlin/org/modelix/model/server/light/LightModelServer.kt
+++ b/model-server-lib/src/main/kotlin/org/modelix/model/server/light/LightModelServer.kt
@@ -182,8 +182,8 @@ class LightModelServer @JvmOverloads constructor(val port: Int, val rootNodeProv
 
     fun Application.installHandlers() {
         install(WebSockets) {
-            pingPeriod = Duration.ofSeconds(15)
-            timeout = Duration.ofSeconds(15)
+            pingPeriod = Duration.ofSeconds(30)
+            timeout = Duration.ofSeconds(30)
             maxFrameSize = Long.MAX_VALUE
             masking = false
         }


odzhychko:
Should we make the timeout configurable?
Ktor provides mechanisms for it https://ktor.io/docs/configuration-file.html
I used them for example here: workspace-manager/src/main/resources/application.conf

languitar:
Configuration would be great. However, whether the configuration file of ktor is read or not depends on whether `engineMain` is used or not. Moreover, we would then also need a mechanism to configure a custom application configuration in the light-model-server instance running inside MPS.

From what I have seen, most libraries seem to default to 20 seconds for pings, but they usually assume simple tasks where communicating parties are not frequently blocked by cpu-intensive long-running tasks. Therefore, maybe this is still also a sensible default.

odzhychko:
I see, we use `embeddedServer`. So the configuration is done in code.

We could add the `timeout` as field to `LightModelServer` and `LightModelServerBuilder` (like `port`).
The field `LightModelServerBuilder` could have a default value.
This would make `LightModelServer` configurable, if it is used outside our MPS plugin.

Of course in the MPS Plugin, where `LightModelServer` is build. we would still hard code the timeout or decide to use the default.

languitar:
Inside MPS, this would either requiring to go through the MPS config mechanism or to just read some environment variable.

In any case, maybe giving a bit more headroom here by default won't hurt?

odzhychko:
I understood, that a configuration in the MPS plugin would be much more to implement and currently not needed.
I was just suggesting this https://github.com/modelix/modelix.core/compare/main...feature/lmx-timeouts-suggestion to pull out constants of `LightModelServer`.

languitar:
> I understood, that a configuration in the MPS plugin would be much more to implement and currently not needed. I was just suggesting this [main...feature/lmx-timeouts-suggestion](https://github.com/modelix/modelix.core/compare/main...feature/lmx-timeouts-suggestion) to pull out constants of `LightModelServer`.

Looks fine. Do you want to open a PR with that change as a replacement for my PR here?

odzhychko:
Sure. Closing in favor of #211

PR #211, opened 2023-08-21
feat(model-server-lib): make timeouts for websockets configurable

odzhychko:


mps-model-server-plugin/src/main/kotlin/org/modelix/model/server/mps/MPSModelServer.kt
```diff
@@ -47,6 +48,7 @@ class MPSModelServer : Disposable {
             val rootNodeProvider: () -> INode? = { MPSModuleRepository.getInstance()?.let { MPSRepositoryAsNode(it) } }
             server = LightModelServer.builder()
                 .port(48305)
+                .websocketTimeout(Duration.ofSeconds(30))
```
languitar, line 51: Is there any specific reason to only increase this in the MPS case?

odzhychko: I though, it was the goal of your PR #208 to make the current Plugin `mps-model-server-plugin` handle larger models.

odzhychko: It would indeed be strange, if it works with MPS but not the model server.
I currently don't know, and would have to take a more in-depth look into, where and how `model-server` and `model-server-lib` are used. Their WebSocket settings are currently inconsistent anyway because the `model-server` does not send pings.

Could you please explain to me, why you increased the timeout in the first place? Where trying to avoid disconnects, communicating with the server running in MPS or with a running `model-server`?

The changes made in #208 only increased the timeouts for the server running in MPS, so I assumed you wanted only to change it there.

languitar: We are currently running against light model server inside MPS. There, we see constant disconnects due to not responding to pings within timeout when a huge bulk of data is initially loaded by the client.

The thing I am wondering about in the PR is why we not simply update the default to 30 seconds in addition to the ability to configure this timeout. My assumption was that the model-server also takes these values here. But if that's not the case, then I am also fine with the current approach.

languitar: Ok, after both PRs things are at least in a similar state now.

languitar (approved):

odzhychko:
Closed in favor of #212

odzhychko:
Closed in favor of #212

PR #216, opened 2023-08-23
MODELIX-502 Extract model data structure into a separate library

slisson:
Requires #213 to be merged first.


slisson:
Is now part of #213

PR #226, opened 2023-09-04
fix(model-api): node resolution inside coroutines

slisson:
The ContextValue class was used to specify how to resolve cross model node references for the current thread. The new ModelQL API is based on coroutines and since they are executed on a thread pool, there is the CoroutineContext as an alternative for ThreadLocal.

The new implementation of ContextValue connects both worlds by keeping the values in ThreadLocal and CoroutineContext in sync. In ModelQLServer.handleCall a coroutine is started where the ContextValue is initially defined in the CoroutineContext and then synchronized with the ThreadLocal for all the non-coroutines code.

While for the JVM target Kotlin already provides the function ThreadLocal.asContextElement, for the JS target a custom solution was necessary (there are no ThreadLocals in JS).

### DIFF
--- a/kotlin-utils/src/commonMain/kotlin/org/modelix/kotlin/utils/ContextValue.kt
+++ b/kotlin-utils/src/commonMain/kotlin/org/modelix/kotlin/utils/ContextValue.kt
@@ -11,14 +11,25 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.modelix.model.api
+package org.modelix.kotlin.utils
 
-@Deprecated("use org.modelix.kotlin.utils.ContextValue from org.modelix:kotlin-utils")
 expect class ContextValue<E> {
 
     constructor()
     constructor(defaultValue: E)
 
-    fun getValue(): E?
-    fun <T> computeWith(newValue: E, r: () -> T): T
+    fun getValue(): E
+    fun getValueOrNull(): E?
+    fun getAllValues(): List<E>
+    fun <T> computeWith(newValue: E, body: () -> T): T
+
+    suspend fun <T> runInCoroutine(newValue: E, body: suspend () -> T): T
+}
+
+fun <E, T> ContextValue<E>.offer(value: E, body: () -> T): T {
+    return if (getAllValues().isEmpty()) {
+        computeWith(value, body)
+    } else {
+        body()
+    }
 }
--- a/kotlin-utils/src/commonMain/kotlin/org/modelix/kotlin/utils/ContextValue.kt
+++ b/kotlin-utils/src/commonMain/kotlin/org/modelix/kotlin/utils/ContextValue.kt
@@ -13,11 +13,20 @@
  */
 package org.modelix.kotlin.utils
 
+/**
+ * A common abstraction over ThreadLocal and CoroutineContext that integrates both worlds.
+ * Allows to set a value that can be read from everywhere on the current thread or coroutine.
+ * A suspendable function can call non suspendable functions and the value is synchronized between the CoroutineContext
+ * and the internal ThreadLocal.
+ */
 expect class ContextValue<E> {
 
     constructor()
     constructor(defaultValue: E)
 
+    /**
+     * @throws NoSuchElementException if no value is set.
+     */
     fun getValue(): E
     fun getValueOrNull(): E?
     fun getAllValues(): List<E>
--- a/kotlin-utils/src/commonTest/kotlin/org/modelix/kotlin/utils/ContextValueTests.kt
+++ b/kotlin-utils/src/commonTest/kotlin/org/modelix/kotlin/utils/ContextValueTests.kt
@@ -27,7 +27,10 @@ import kotlin.time.Duration.Companion.milliseconds
 class ContextValueTests {
 
     @Test
-    fun multipleCoroutines() = runTest {
+    fun testIsolation() = runTest {
+        // run multiple suspendable and non-suspendable functions in parallel to ensure they always read their own value
+        // and not a value from a different coroutine/thread.
+
         val contextValue = ContextValue<String>("a")
         assertEquals("a", contextValue.getValueOrNull())
         coroutineScope {
--- a/kotlin-utils/src/commonTest/kotlin/org/modelix/kotlin/utils/ContextValueTests.kt
+++ b/kotlin-utils/src/commonTest/kotlin/org/modelix/kotlin/utils/ContextValueTests.kt
@@ -26,11 +26,16 @@ import kotlin.time.Duration.Companion.milliseconds
 
 class ContextValueTests {
 
+    /*
+     * Runs multiple suspendable and non-suspendable functions in parallel to ensure they always read their own value
+     * and not a value from a different coroutine/thread.
+
+     * This test starts two coroutines and tries to run into a race-condition. A successful test doesn't proof the
+     * correctness, but a failing test proofs its incorrectness. If it ever becomes unstable, meaning it first fails
+     * and then succeeds after a second run, this shouldn't be ignored.
+     */
     @Test
     fun testIsolation() = runTest {
-        // run multiple suspendable and non-suspendable functions in parallel to ensure they always read their own value
-        // and not a value from a different coroutine/thread.
-
         val contextValue = ContextValue<String>("a")
         assertEquals("a", contextValue.getValueOrNull())
         coroutineScope {
--- a/kotlin-utils/src/commonTest/kotlin/org/modelix/kotlin/utils/ContextValueTests.kt
+++ b/kotlin-utils/src/commonTest/kotlin/org/modelix/kotlin/utils/ContextValueTests.kt
@@ -1,0 +1,68 @@
+/*
+ * Copyright (c) 2023.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.modelix.kotlin.utils
+
+import kotlinx.coroutines.coroutineScope
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.test.runTest
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.time.Duration.Companion.milliseconds
+
+class ContextValueTests {
+
+    @Test
+    fun multipleCoroutines() = runTest {
+        val contextValue = ContextValue<String>("a")
+        assertEquals("a", contextValue.getValueOrNull())
+        coroutineScope {
+            launch {
+                for (i in 1..10) {
+                    contextValue.runInCoroutine("b1") {
+                        contextValue.computeWith("b11") {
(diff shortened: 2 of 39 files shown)


slisson:
I updated the description.

kotlin-utils/src/commonMain/kotlin/org/modelix/kotlin/utils/ContextValue.kt
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.modelix.kotlin.utils
+
+expect class ContextValue<E> {
```
languitar, line 16: Parts of the PR description could be a good fit for a kdoc comment here, explaining the purpose of this class.

kotlin-utils/src/commonMain/kotlin/org/modelix/kotlin/utils/ContextValue.kt
```diff
@@ -0,0 +1,35 @@
...
+
+    constructor()
+    constructor(defaultValue: E)
+
+    fun getValue(): E
```
languitar, line 21: Would that throw something in case no value is present? Some documentation on the Exception to expect would be good.

kotlin-utils/src/commonTest/kotlin/org/modelix/kotlin/utils/ContextValueTests.kt
```diff
@@ -0,0 +1,68 @@
...
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.test.runTest
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.time.Duration.Companion.milliseconds
+
+class ContextValueTests {
+
+    @Test
+    fun multipleCoroutines() = runTest {
```
languitar, line 30: This test is pretty complicated and hard to understand. Maybe the test method name can be rephrased to some human-readable statement on which condition is being tested here?

kotlin-utils/src/commonTest/kotlin/org/modelix/kotlin/utils/ContextValueTests.kt
```diff
@@ -0,0 +1,68 @@
...
+        val contextValue = ContextValue<String>("a")
+        assertEquals("a", contextValue.getValueOrNull())
+        coroutineScope {
+            launch {
+                for (i in 1..10) {
```
languitar, line 35: Is it really necessary to launch that many instances for verifying the desired behavior?

kotlin-utils/src/jsMain/kotlin/org/modelix/kotlin/utils/ContextValue.js.kt
```diff
@@ -0,0 +1,100 @@
...
+import kotlin.coroutines.Continuation
+import kotlin.coroutines.ContinuationInterceptor
+import kotlin.coroutines.CoroutineContext
+
+actual class ContextValue<E> {
+    private val initialStack: List<E>
+    private var synchronousValueStack: List<E>? = null
+    private var stackFromCoroutine: List<E>? = null
+    private val contextElementKey = object : CoroutineContext.Key<ContextValueElement<E>> {}
+    private var isInSynchronousBlock = false
```
languitar, line 28: I am not sure about the JS backend and the expected access pattern here, but are you sure this doesn't need a `@Volatile` annotation?

kotlin-utils/src/jsMain/kotlin/org/modelix/kotlin/utils/ContextValue.js.kt
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    private suspend fun getAllValuesFromCoroutine(): List<E>? {
+        return currentCoroutineContext()[contextElementKey]?.stack
+    }
+
+    actual suspend fun <T> runInCoroutine(newValue: E, body: suspend () -> T): T {
+        return withContext(ContextValueElement((getAllValuesFromCoroutine() ?: initialStack) + newValue)) {
+            val parentInterceptor = checkNotNull(currentCoroutineContext()[ContinuationInterceptor]) {
+                "No ContinuationInterceptor found in the context"
```
languitar, line 71: When would this happen? Maybe the exception could give the caller some more hints on what he or she has done wrong.

kotlin-utils/src/jsMain/kotlin/org/modelix/kotlin/utils/RunSynchronized.js.kt
```diff
@@ -0,0 +1,21 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.modelix.kotlin.utils
+
+actual inline fun <R> runSynchronized(lock: Any, block: () -> R): R {
+    return block()
```
languitar, line 20: Some hint on why the lock is not needed here would be good for understanding.

model-api/src/commonMain/kotlin/org/modelix/model/api/ContextValue.kt
```diff
@@ -13,6 +13,7 @@
  */
 package org.modelix.model.api
 
+@Deprecated("use org.modelix.kotlin.utils.ContextValue from org.modelix:kotlin-utils")
 expect class ContextValue<E> {
```
languitar, line 17: Looks like this class is a subset of the class from the new utils package. Could this deprecation be made easier for clients if the new implementation was type-aliased here? That would also avoid having two implementations in parallel?

Or can we treat the whole context value as being something internal and just avoid the deprecation cycle? I can't remember accessing that from client code.

model-api/src/commonMain/kotlin/org/modelix/model/api/INodeResolutionScope.kt
```diff
@@ -1,41 +1,113 @@
...
+     */
+    @Deprecated("Don't call this method directly. Use INodeReference.resolveIn instead.")
     fun resolveNode(ref: INodeReference): INode?
 
-    companion object Key : CoroutineContext.Key<INodeResolutionScope>
+    /**
+     * All node references inside the body are resolved against this scope. Compared to runWithAlso, the existing scopes
+     * in the current context are not used, meaning they are replaced.
+     */
+    fun <T> runWithOnly(body: () -> T): T = contextScope.computeWith(this, body)
```
languitar, line 23: The method name sound as if some word was forgotten. Only "what"?

model-api/src/commonMain/kotlin/org/modelix/model/api/INodeResolutionScope.kt
```diff
@@ -1,41 +1,113 @@
...
-    override fun <R> fold(initial: R, operation: (R, CoroutineContext.Element) -> R): R {
-        // coroutines are not compiled with -Xjvm-default=all-compatibility
-        // to not break existing IAreas implemented in Java, this override is necessary
-        return super.fold(initial, operation)
-    }
+    /**
+     * All node references inside the body are resolved against this scope and if that fails against any other scope in
+     * the current context.
+     */
+    fun <T> runWithAlso(body: () -> T): T = runWithAlso(this, body)
```
languitar, line 34: Same as above. Adding one more word what is "also" included here would make this more readable.

model-datastructure/src/commonMain/kotlin/org/modelix/model/lazy/PrefetchCache.kt
```diff
@@ -59,12 +58,12 @@ class PrefetchCache(private val store: IDeserializingKeyValueStore) : IDeseriali
     }
 
     companion object {
-        private val contextValue: ContextValue<PrefetchCache> = ContextValue()
+        private val contextValue: org.modelix.kotlin.utils.ContextValue<PrefetchCache> = org.modelix.kotlin.utils.ContextValue()
```
languitar, line 61: Why not imported?

modelql-core/src/commonMain/kotlin/org/modelix/modelql/core/ContextValue.kt
```diff
@@ -13,6 +13,7 @@
  */
 package org.modelix.modelql.core
 
+@Deprecated("use org.modelix.kotlin.utils.ContextValue from org.modelix:kotlin-utils")
```
languitar, line 16: All new deprecations should probably use the [ReplaceWith](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-deprecated/-init-.html) argument.

languitar (commented):
What I am missing here is a test that demonstrates what was actually fixed. There's a test for the utility implementation behind the problem but I can't see which test shows that the overall feature works now.

kotlin-utils/src/jsMain/kotlin/org/modelix/kotlin/utils/ContextValue.js.kt
slisson: There is no `@Volatile` in JS.

model-api/src/commonMain/kotlin/org/modelix/model/api/ContextValue.kt
slisson: The class is used by Java code written in MPS. Making it a type-alias would be a breaking change.

modelql-core/src/commonMain/kotlin/org/modelix/modelql/core/ContextValue.kt
slisson: The API is different and an automated replacement not possible.

kotlin-utils/src/commonTest/kotlin/org/modelix/kotlin/utils/ContextValueTests.kt
slisson: The loop is inside the launch. It doesn't launch new coroutines.

languitar: But why is it required to loop 10 times? Is there a chance to simplify the test here and to limit it to the bare minimum required to verify the condition being tested here?

kotlin-utils/src/jsMain/kotlin/org/modelix/kotlin/utils/ContextValue.js.kt
slisson: It's not expected to ever happen.

kotlin-utils/src/commonTest/kotlin/org/modelix/kotlin/utils/ContextValueTests.kt
slisson: It tries to find a race-condition between the coroutines. Since a successful test doesn't proof the correctness there is no minimum value. The higher the value the more likely it is that it detects such a race-condition.

languitar: Too bad [thread-weaver](https://github.com/google/thread-weaver) isn't a thing anymore...

Can you add a comment explaining exactly this? That would make the code a lot more understandable.

kotlin-utils/src/jsMain/kotlin/org/modelix/kotlin/utils/ContextValue.js.kt
languitar:
```suggestion
                "No ContinuationInterceptor found in the context. This must not happen."
```

languitar (approved):

PR #231, opened 2023-09-04
fix(modelql): CCE from SimpleStepOutput to MultiplexedOutput

slisson:
There was a mismatch between the elements returned by `ZipElementAccessStep.createFlow` and what the serializer of `ZipElementAccessStep.getOutputSerializer` could handle. `ZipElementAccessStep` doesn't implement it's own serializer, but forwards the serializer and the elements of the step that produced the zip input.

The bug was that `ZipElementAccessStep` didn't forward the `IStepOutput`, but unwrapped it and rewrapped into a new `SimpleStepOutput`, because that's the default behavior of the super class `MonoTransformingStep`.

### DIFF
--- a/modelql-core/src/commonMain/kotlin/org/modelix/modelql/core/IStep.kt
+++ b/modelql-core/src/commonMain/kotlin/org/modelix/modelql/core/IStep.kt
@@ -183,7 +183,7 @@ abstract class MonoTransformingStep<In, Out> : TransformingStep<In, Out>(), IMon
     fun connectAndDowncast(producer: IFluxStep<In>): IFluxStep<Out> = also { producer.connect(it) }
 
     override fun createFlow(input: StepFlow<In>, context: IFlowInstantiationContext): StepFlow<Out> {
-        return input.map { transform(context.evaluationContext, it.value).asStepOutput(this) }
+        return input.map { transform(context.evaluationContext, it) }
     }
 
     override fun createSequence(evaluationContext: QueryEvaluationContext, queryInput: Sequence<Any?>): Sequence<Out> {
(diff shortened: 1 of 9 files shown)


slisson:
Still checking for similar cases and ways to prevent this in the future.

slisson:
I updated the description of the PR explaining the bug.

modelql-core/src/commonMain/kotlin/org/modelix/modelql/core/IStep.kt
```diff
@@ -200,6 +200,12 @@ abstract class MonoTransformingStep<In, Out> : TransformingStep<In, Out>(), IMon
     override fun evaluate(evaluationContext: QueryEvaluationContext, queryInput: Any?): Optional<Out> {
         return getProducer().evaluate(evaluationContext, queryInput).map { transform(evaluationContext, it) }
     }
+
+    protected open fun transform(evaluationContext: QueryEvaluationContext, input: IStepOutput<In>): IStepOutput<Out> {
+        return transform(evaluationContext, input.value).asStepOutput(this)
+    }
+
+    @Deprecated("use the function that preserves the IStepOutput")
     abstract fun transform(evaluationContext: QueryEvaluationContext, input: In): Out
```
languitar, line 209: Is there a need for this deprecation cycle? If that method is something that's used only internally, it could probably be dropped directly?

languitar (approved):
Thanks for the explanations. The  code is still a bit hard to read due to the various `transform` methods lacking documentation but I think I could roughly follow the ideas now.

modelql-core/src/commonMain/kotlin/org/modelix/modelql/core/IStep.kt
slisson: I removed the annotation. Since most steps produce a new value and don't just forward the input, using that overload is still valid and reduces duplication.

PR #235, opened 2023-09-06
Feature/make modelclientv2 closable

odzhychko:
This useful, when the application lives longer than the model client. The HTTP client of the model client can then be disposed of.


odzhychko:
> That's a good addition. Can you add a test for this?
Sure 😄

PR #236, opened 2023-09-06
fix(modelql): access to references outside of queries

languitar:
Fixes: MODELIX-530

TODO

* [x] add a test. @slisson is there an easy way to test this?


slisson:
I added a test and a proper fix.

slisson (dismissed):

PR #237, opened 2023-09-06
MODELIX-480: Migrate sync relevant features to kotlin lib

benedekh:
# Related YouTrack epics and issues

- https://issues.modelix.org/issue/MODELIX-658
- https://issues.modelix.org/issue/MODELIX-480
   - https://issues.modelix.org/issue/MODELIX-637
   - https://issues.modelix.org/issue/MODELIX-628
   - https://issues.modelix.org/issue/MODELIX-635
   - https://issues.modelix.org/issue/MODELIX-634
   - https://issues.modelix.org/issue/MODELIX-629


benedekh:
Hi @modelix/developers, 

Could you please do a code review on the [MPSLanguageRepository.kt](https://github.com/modelix/modelix.core/pull/237/files#diff-02f7cfb500daed71771a1d0b45ba14a61be6b519be34ed52623269092f750c57) class in this pull request? 

I made the following changes to the class:
1. I've extended the class with a new function, `getConcept(uid:String)`, that I need in the [SNodeFactory](https://github.com/modelix/modelix.core/pull/237/files#diff-c915e3478be6c462fc657000d04039addfab4eb29afee4fbe80e40da3b719b9dR52) class when creating new SNodes in MPS. To implement this class, I cache all SConcepts with their UIDs in the `init` block of the class. This helps the lookup time, but has some memory implications.
2. I've implemented its `hashCode` and `equals` methods so that when we store the ILanguageRepositories in a `Set` then the structural equality will be checked even if we create more than one MPSLanguageRegistries on the run. An example for such `Set` is in `ILanguageRepository.Companion.repositories`. An example for such registration is in [SyncServiceImpl.registerLanguages(project: MPSProject)](https://github.com/modelix/modelix.core/pull/237/files#diff-730a0aca4f1127b59ccbb0180c5da5322b2d2e9390732d92484a80c5b4d0cc52R186).

The rest of this pull request does not impact how the existing code base works, so they should be safe to merge. Nevertheless, feel free to take a look at the `mps-sync-plugin` and `mps-sync-plugin-lib` projects.

I look forward to your feedback.

mps-model-adapters/src/main/kotlin/org/modelix/model/mpsadapters/MPSLanguageRepository.kt
```diff
@@ -21,11 +21,25 @@ import jetbrains.mps.smodel.adapter.structure.MetaAdapterFactory
...
 import jetbrains.mps.smodel.runtime.illegal.IllegalConceptDescriptor
+import org.jetbrains.mps.openapi.language.SAbstractConcept
 import org.jetbrains.mps.openapi.module.SRepository
 import org.modelix.model.api.IConcept
 import org.modelix.model.api.ILanguageRepository
 
-class MPSLanguageRepository(private val repository: SRepository) : ILanguageRepository {
+class MPSLanguageRepository(val repository: SRepository) : ILanguageRepository {
+
+    private val concepts = mutableMapOf<String, MPSConcept>()
```
mhuster23, line 31: I'm not sure, if we can make this memory trade-off.
Since we have projects with a lot of concepts, it may not be feasible to cache them all.
@languitar what do you think?

mps-model-adapters/src/main/kotlin/org/modelix/model/mpsadapters/MPSLanguageRepository.kt
```diff
@@ -40,13 +54,20 @@ class MPSLanguageRepository(private val repository: SRepository) : ILanguageRepo
...
+    override fun equals(other: Any?) =
+        if (this === other) {
+            true
+        } else if (javaClass != other?.javaClass) {
+            false
+        } else {
+            repository == (other as MPSLanguageRepository).repository
+        }
+
+    override fun hashCode() = repository.hashCode()
```
mhuster23, line 72: In case the concept cache is dropped, we could just make this a `data class` to let kotlin handle these methods.

commitlint.config.js
```diff
@@ -23,6 +23,10 @@ module.exports = {
         "mps-model-adapters",
         "mps-model-server",
         "mps-model-server-plugin",
+        "mps-sync-plugin",
+        "mps-sync-plugin-lib",
+        "sync-plugin-lib",
+        "mps-sync-lib",
```
mhuster23, line 29: These are probably in here because of the linting of the oldest commits. We should remove them before merging. To do that we can override the linter check for this PR.

mhuster23 (commented):
I had a quick look and added some thoughts.

Thinking about the changelog, it might be worth to reword or squash a few commits, since it's generated out of the commit messages.

mps-model-adapters/src/main/kotlin/org/modelix/model/mpsadapters/MPSLanguageRepository.kt
languitar: I have no clue how much memory all concepts would take up in our use case. We'd have to test that. But they are probably loaded somewhere in any case?

commitlint.config.js
benedekh: > To do that we can override the linter check for this PR.

Can you help me with that, please?

mps-model-adapters/src/main/kotlin/org/modelix/model/mpsadapters/MPSLanguageRepository.kt
benedekh: @mhuster23 Can you give me such example projects? I could do some measurements to see if caching is a real issue in that case.

languitar: We probably can't give you access to the project. I just counted and found 6500 concepts

slisson: I don't understand why this cache was introduced. It's not an expensive computation and the only usage I can find already does some caching.

mps-model-adapters/src/main/kotlin/org/modelix/model/mpsadapters/MPSLanguageRepository.kt
```diff
@@ -40,13 +54,20 @@ class MPSLanguageRepository(private val repository: SRepository) : ILanguageRepo
...
     }
 
-    override fun getAllConcepts(): List<IConcept> {
-        val result = mutableListOf<IConcept>()
-        LanguageRegistry.getInstance(repository).withAvailableLanguages { language ->
-            result.addAll(language.identity.concepts.map { MPSConcept(it) }.toList())
-        }
-        return result
-    }
+    fun getConcept(uid: String): SAbstractConcept? = concepts[uid]?.concept
```
slisson, line 57: Why not `resolveConcept(uid)?.concept`?

benedekh: Good point! I've missed that method for some reasons. I applied your suggestion and now we do not need the cache anymore. Thank you for the suggestion again.

mps-model-adapters/src/main/kotlin/org/modelix/model/mpsadapters/MPSLanguageRepository.kt
benedekh: In [5c08a24](https://github.com/modelix/modelix.core/pull/237/commits/5c08a244b6c5bacb28a89d0a24716589a4d8caf2), I removed the cache so this conversation became outdated.

mps-model-adapters/src/main/kotlin/org/modelix/model/mpsadapters/MPSLanguageRepository.kt
benedekh: fixed in[976691a](https://github.com/modelix/modelix.core/pull/237/commits/976691a5cbd5356adf6fe455c15ca77b4103575f)

commitlint.config.js
mhuster23: The alternative would be to reword all commits that use the old commit scopes and then force push, but that's quite a few.

benedekh:
Deprecating in favor of https://github.com/modelix/modelix.core/pull/482.

#482 has the commit history cleaned up so commitlint will not complain and the old commits will not appear in the changelog. Let's continue the review discussions there.

PR #238, opened 2023-09-07
Draft: feat(model-api-gen-gradle): add custom task for usage of generated languages

odzhychko:


model-api-gen-gradle-test/build.gradle.kts
```diff
@@ -108,6 +110,19 @@ metamodel {
...
 }
 
+@OptIn(UnstableGeneratorFeature::class)
+tasks.register<UseExportedLanguagesTask>("generateUsedLanguagesSummary") {
+    val outputFile = buildDir.resolve("languagesSummary.txt")
+    outputs.file(outputFile)
+    dependsOn("exportMetaModelFromMps")
+    includedLanguages.set(listOf("jetbrains.mps.lang.core"))
+    doWithLanguages { languages ->
+        // An actual user case would extract the custom generator into its own class.
```
odzhychko, line 120: The main use case of this feature is to use the generated languages for custom generators.

Currently, prototyping generators needs directly modifying the plugin.
And after prototyping the generator, it must become part of the offical plugin or a fork has to be maintained.
See. https://github.com/abstraktor/modelix.core/tree/feature/VuejsMMGenerator-proxy

I think this feature is a good escape hatch. Because I'm not certain about the API, and it exposes internal structures almost directly, I marked it as `UnstableGeneratorFeature`.

For example, I will use this escape hatch to develop (at least the first iteration of) samples with Vue and React.

I'm not certain such an escape hatch is maybe feature creep. Perhaps users should rather the generation capacities of MPS. I am curious about your opinion.

model-api-gen-gradle-test/src/test/kotlin/GeneratedApiTest.kt
```diff
@@ -43,6 +43,13 @@ class GeneratedApiTest {
         }
     }
 
+    @Test
+    fun `execute custom UseExportedLanguagesTask`() {
```
languitar, line 47: That's not consequently applied in this project here so far, however, I am generally in favor of [expressing readable requirements with test case names](https://www.semipol.de/posts/2021/08/on-writing-useful-unit-tests/#_use_test_case_names_to_express_readable_requirements). Right now, I have not much of a clue what is actually being tested here and in case of a failure I also wouldn't exactly know which requirement is violated from the test name.

```suggestion
    fun `UseExportedLanguagesTask has access to all used MPS langauges`() {
```

That would give a lot more hints.

model-api-gen-gradle/src/main/kotlin/org/modelix/metamodel/gradle/UseExportedLanguagesTask.kt
```diff
@@ -0,0 +1,80 @@
...
+/**
+ * This task allows to execute custom build steps which work with the generated and processed languages
+ * from the "exportMetaModelFromMps" task.
+ *
+ * The languages can be accessed through the [ReadonlyProcessedLanguageSet] structure.
+ * Custom user actions have to be registered as [ActionWithLanguages]] in [doWithLanguages]
+ */
+@CacheableTask
+@UnstableGeneratorFeature
+abstract class UseExportedLanguagesTask @Inject constructor(of: ObjectFactory) : DefaultTask() {
```
languitar, line 45: I am not 100 % happy with the name of the task. "Use" somehow doesn't communicate well the intention of this task. But the ideas I can come up with are also not too good:

* WithExportedLanguagesTask
* PerformWithExportedLanguagesTask

model-api-gen-gradle/src/main/kotlin/org/modelix/metamodel/gradle/LanguageDataReading.kt
```diff
@@ -0,0 +1,30 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.modelix.metamodel.gradle
+
+import org.modelix.model.data.LanguageData
+import java.io.File
+
+object LanguageDataReading {
```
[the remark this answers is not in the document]
odzhychko: I renamed the function to something more appropriate.
I liked using `ClassName.staticMethod` in Java to see that the method comes from a different file.
But I'm not particular about it. I change it.

model-api-gen-gradle-test/src/test/kotlin/GeneratedApiTest.kt
odzhychko: Ok, that was a good suggestion.

model-api-gen-gradle/src/main/kotlin/org/modelix/metamodel/gradle/UseExportedLanguagesTask.kt
odzhychko: I liked `PerformWithExportedLanguagesTask`.

It's an @UnstableGeneratorFeature, so we can always rename it 😆

languitar: Still, if we can avoid renaming things, people are happier usually ;-)

model-api-gen/src/main/kotlin/org/modelix/metamodel/generator/ReadonlyGeneratorInput.kt
```diff
@@ -0,0 +1,61 @@
...
+    fun fqName(): String
+
+    fun getDirectSuperConcepts(): Sequence<ReadonlyProcessedConcept>
+    fun getAllSuperConceptsAndSelf(): Sequence<ReadonlyProcessedConcept>
+    fun getDuplicateSuperConcepts(): List<ReadonlyProcessedConcept>
+
+    fun getOwnRoles(): List<ReadonlyProcessedRole>
+}
+
+sealed interface ReadonlyProcessedRole : ReadonlyProcessedDeprecatable {
```
odzhychko, line 32: I need to take a look at theclass  hierarchy of `ReadonlyProcessedRole`. 
Also, I should look in general over the other hierarchies.

model-api-gen/src/main/kotlin/org/modelix/metamodel/generator/MetaModelGenerator.kt
```diff
@@ -949,7 +945,7 @@ class MetaModelGenerator(val outputDir: Path, val nameConfig: NameConfig = NameC
     private fun ProcessedConcept.conceptWrapperInterfaceClass() =
         ClassName(language.name, nameConfig.typedConcept(name))
 
-    private fun ProcessedLanguage.generatedClassName() = ClassName(name, nameConfig.languageClass(name))
+    private fun ReadonlyProcessedLanguage.generatedClassName() = ClassName(name, nameConfig.languageClass(name))
     private fun ProcessedConcept.nodeWrapperInterfaceName() = nameConfig.typedNode(name)
```
odzhychko, line 949: Here the read-only Intefarces should be used.

model-api-gen/src/main/kotlin/org/modelix/metamodel/generator/ReadonlyGeneratorInput.kt
```diff
@@ -0,0 +1,62 @@
+package org.modelix.metamodel.generator
+
+import org.modelix.model.data.PropertyType
+
+@RequiresOptIn(message = "This API is experimental. It may be changed in the future without notice.")
+@Retention(AnnotationRetention.BINARY)
+@Target(AnnotationTarget.CLASS, AnnotationTarget.FUNCTION)
+annotation class UnstableGeneratorFeature
+interface ReadonlyProcessedLanguageSet {
```
odzhychko, line 9: Think about naming again.

PR #239, opened 2023-09-07
fix(model-client): change wrong dependency to Closable

odzhychko:
Do not depend on Ktor utils.
Just provide a close method.
Closable is only a concept on the JVM.

### DIFF
--- a/model-client/src/commonMain/kotlin/org/modelix/model/client2/IModelClientV2.kt
+++ b/model-client/src/commonMain/kotlin/org/modelix/model/client2/IModelClientV2.kt
@@ -13,14 +13,26 @@
  */
 package org.modelix.model.client2
 
-import io.ktor.utils.io.core.Closeable
 import org.modelix.model.IVersion
 import org.modelix.model.api.IIdGenerator
 import org.modelix.model.lazy.BranchReference
 import org.modelix.model.lazy.RepositoryId
 import org.modelix.model.server.api.ModelQuery
 
-interface IModelClientV2 : Closeable {
+/**
+ * This interface is meant exclusively for model client usage.
+ *
+ * It is designed to ensure decoupling between model client usage operations and other aspects,
+ * such as lifecycle management.
+ * Users of this interface cannot incidentally depend on non-usage functionality.
+ * See also: [Interface segregation principle](https://en.wikipedia.org/wiki/Interface_segregation_principle)
+ *
+ * Specifically, this interface should not be used for managing the client's lifecycle,
+ * as the lifecycle management may vary depending on the specific implementation.
+ * If you need to manage the client's lifecycle, use the methods in the class interface of the concrete implementations,
+ * such as [ModelClientV2].
+ */
+interface IModelClientV2 {
     fun getClientId(): Int
     fun getIdGenerator(): IIdGenerator
     fun getUserId(): String?
@@ -45,6 +57,4 @@ interface IModelClientV2 : Closeable {
 
     suspend fun poll(branch: BranchReference, lastKnownVersion: IVersion?): IVersion
     suspend fun poll(branch: BranchReference, lastKnownVersion: IVersion?, filter: ModelQuery): IVersion
-
-    override fun close()
 }
--- a/model-client/src/commonMain/kotlin/org/modelix/model/client2/ModelClientV2.kt
+++ b/model-client/src/commonMain/kotlin/org/modelix/model/client2/ModelClientV2.kt
@@ -50,7 +50,7 @@ import kotlin.time.Duration.Companion.seconds
 class ModelClientV2(
     private val httpClient: HttpClient,
     val baseUrl: String,
-) : IModelClientV2 {
+) : IModelClientV2, Closable {
     private var clientId: Int = 0
     private var idGenerator: IIdGenerator = IdGeneratorDummy()
     private var userId: String? = null
--- a/model-client/src/commonMain/kotlin/org/modelix/model/client2/Closable.kt
+++ b/model-client/src/commonMain/kotlin/org/modelix/model/client2/Closable.kt
@@ -1,0 +1,21 @@
+/*
+ * Copyright (c) 2023.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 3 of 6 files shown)


model-client/src/commonMain/kotlin/org/modelix/model/client2/IModelClientV2.kt
```diff
@@ -13,14 +13,13 @@
...
 package org.modelix.model.client2
 
-import io.ktor.utils.io.core.Closeable
 import org.modelix.model.IVersion
 import org.modelix.model.api.IIdGenerator
 import org.modelix.model.lazy.BranchReference
 import org.modelix.model.lazy.RepositoryId
 import org.modelix.model.server.api.ModelQuery
 
-interface IModelClientV2 : Closeable {
```
languitar, line 23: Shouldn't this be implemented instead: https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-auto-closeable/

languitar: in the JVM implementation then

odzhychko: https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-auto-closeable/ is `@ExperimentalStdlibApi`, though I avoided it.

Also, deriving a JVM specific implementation to have it implement https://docs.oracle.com/javase/8/docs/api/java/io/Closeable.html made not too much sense.

languitar: As far as I can tell, this change is now breaking for JVM customers. The ktor [Closeable](https://api.ktor.io/ktor-io/io.ktor.utils.io.core/-closeable/index.html) as an implementation of JVM closeable. Thus, in the JVM world, IModelClientV2 could be used with `use`. Without providing some means of implementing Closeable in JVM again, using the client inside `use` or try-with-resource in Java is not possible anymore now. Therefore, this would need to be marked as a breaking change.

odzhychko: Ok, I see. I will add something, that does not break the JVM implementation.

odzhychko: You can have a look at it again.

languitar: Wouldn't it be less code to simple do the same trick that ktor does with their Closeable: declare that as a platform specific interface and implement it differently per platform. Then, you wouldn't have to rebuild the whole IModelClientV2 interface in each platform.

odzhychko: But then we would have something like `org.modelix.core.Closable` which would become part of our API and if we were to remove or move it, it would be a breaking change. For now, I would like to avoid a modelix-specific `Closable`-interface.

slisson:
The `close` method doesn't belong into the `IModelClientV2` interface. Whoever instantiates ModelClientV2 is responsible for cleaning it up. The interface is intended for just using the client, not for managing it's lifecycle. Please remove Closable from IModelClientV2 and add it to the ModelClientV2 class only.

model-client/src/commonMain/kotlin/org/modelix/model/client2/IModelClientV2.kt
odzhychko: This seems to do the trick.

```
package org.modelix.model.client2

internal expect interface Closable {
    fun close()
}
```

Because the interface is internal, it does not get exposed. 
But still Javas `java.io.Closeable`, `java.io.Closeable.close` and the `use`-extension functions work.

odzhychko:
> The `close` method doesn't belong into the `IModelClientV2` interface. Whoever instantiates ModelClientV2 is responsible for cleaning it up. The interface is intended for just using the client, not for managing it's lifecycle. Please remove Closable from IModelClientV2 and add it to the ModelClientV2 class only.

I change it.

model-client/src/commonMain/kotlin/org/modelix/model/client2/ModelClientV2.kt
```diff
@@ -227,6 +231,7 @@ abstract class ModelClientV2Builder {
         return ModelClientV2(
             httpClient?.config { configureHttpClient(this) } ?: createHttpClient(),
```
odzhychko, line 232: `httpClient?.config` copies the client.
See. https://api.ktor.io/ktor-client/ktor-client-core/io.ktor.client/-http-client/config.html

So even when a custom `httpClient` is set, we should probably close its copy.

slisson (approved):

languitar:
> The `close` method doesn't belong into the `IModelClientV2` interface. Whoever instantiates ModelClientV2 is responsible for cleaning it up. The interface is intended for just using the client, not for managing it's lifecycle. Please remove Closable from IModelClientV2 and add it to the ModelClientV2 class only.

Can we add such important design decisions to the kdoc comments of the interface? Having such documentation place would have avoided all the discussion and PR loops here.

model-client/src/commonMain/kotlin/org/modelix/model/client2/IModelClientV2.kt
languitar: The current solution looks fine. A lot less code to maintain now.

model-client/src/commonMain/kotlin/org/modelix/model/client2/IModelClientV2.kt
```diff
@@ -13,14 +13,18 @@
...
+/**
+ * This interface is meant exclusively for client usage.
+ *
+ * It is explicitly not intended for managing the lifecycle of the client.
```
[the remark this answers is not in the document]
odzhychko: Ok, you can have a look again ^^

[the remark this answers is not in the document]
languitar: :+1:

languitar (approved):

PR #242, opened 2023-09-12
MODELIX-520 Make concept alias available in generated source code

mhuster23:


slisson:
I would like to have a more type safe way of accessing the "concept properties" by generating a separate field for each property instead of this generic metaInfo field.

The generated code should then look like this:
```kotlin
interface C_BaseConcept {
    val alias: String? get() = null
    companion object : C_BaseConcept {
        val alias: String? = null
    }
}
interface C_MyConcept {
    val alias: String? = "myConcept"
    companion object : C_MyConcept {
        val alias: String? = null
    }
}
```

This requires to first analyze all concepts and collect the defined concept properties (they can then be stored in `ProcessedLanguageSet`) to then generate them into the whole concept hierarchy.

mhuster23:
If we want to do it like that, we will probably need to generate interfaces for each metaInfo attribute e.g. 
```kotlin
interface IHasAlias {
   val alias: String?
}
```  
So that we can access this meta info generically across concepts.

model-api-gen/src/main/kotlin/org/modelix/metamodel/generator/GeneratorInput.kt
```diff
@@ -220,6 +227,7 @@ internal class ProcessedConcept(
     var abstract: Boolean,
     val extends: MutableList<ProcessedConceptReference>,
     override var deprecationMessage: String?,
+    val metaInfo: Map<String, String>?,
```
languitar, line 230: Is it necessary to make this nullable? Can't that always be an empty map if nothing provided? That would simplify usage for clients.

model-api-gen-runtime/src/commonMain/kotlin/org/modelix/metamodel/IMetaInfoProvider.kt
```diff
@@ -0,0 +1,27 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.modelix.metamodel
+
+/**
+ * Provides a map for storing arbitrary information
```
languitar, line 20: If that is only a map, why is a separate interface needed?

model-api-gen/src/main/kotlin/org/modelix/metamodel/generator/GeneratorInput.kt
```diff
@@ -62,6 +63,22 @@ internal class ProcessedLanguageSet(dataList: List<LanguageData>) : IProcessedLa
...
+        val concepts = languages.flatMap { it.getConcepts() }
+        concepts.forEach { concept ->
+            val keys = concept.metaProperties.keys
+            // TODO the computation of rootConcept will visit some concepts multiple times across loop iterations; could potentially be optimized
+            val rootConcept = concept.getNonInterfaceRootConcept()
+            if (conceptFqName2RootMetaProperties.containsKey(rootConcept.fqName())) {
+                conceptFqName2RootMetaProperties[rootConcept.fqName()]?.addAll(keys)
+            } else {
+                conceptFqName2RootMetaProperties[rootConcept.fqName()] = mutableSetOf<String>().apply { addAll(keys) }
+            }
```
languitar, line 80:
```suggestion
            conceptFqName2RootMetaProperties[rootConcept.fqName()].addAll(keys)
```

model-api-gen/src/main/kotlin/org/modelix/metamodel/generator/GeneratorInput.kt
```diff
@@ -62,6 +63,22 @@ internal class ProcessedLanguageSet(dataList: List<LanguageData>) : IProcessedLa
         initIndexes()
         resolveConceptReferences()
         fixRoleConflicts()
+        propagateMetaProperties()
+    }
+
+    private fun propagateMetaProperties() {
+        conceptFqName2RootMetaProperties = mutableMapOf()
```
languitar, line 70:
```suggestion
        conceptFqName2RootMetaProperties = mutableMapOf().withDefault {
            mutableSetOf<String>()
        }
```

languitar (dismissed):
Some minor ideas for code improvements.

model-api-gen/src/main/kotlin/org/modelix/metamodel/generator/GeneratorInput.kt
```diff
@@ -253,6 +278,17 @@ internal class ProcessedConcept(
     fun getAllSuperConcepts(): Sequence<ProcessedConcept> = getAllSuperConcepts_().distinct()
     fun getAllSuperConceptsAndSelf(): Sequence<ProcessedConcept> = getAllSuperConceptsAndSelf_().distinct()
     fun getDuplicateSuperConcepts() = getAllSuperConcepts_().groupBy { it }.filter { it.value.size > 1 }.map { it.key }
+
+    fun getNonInterfaceSuperConcept() = getDirectSuperConcepts().firstOrNull()
```
slisson, line 282: There are no "interface concepts" in Modelix. This is a term from MPS, but the generator shouldn't make any MPS specific assumptions. But after replacing `conceptFqName2RootMetaProperties` with a single Set, these method shouldn't be required anymore.

model-api-gen/src/main/kotlin/org/modelix/metamodel/generator/MetaModelGenerator.kt
```diff
@@ -647,6 +647,15 @@ class MetaModelGenerator(val outputDir: Path, val nameConfig: NameConfig = NameC
             for (extended in concept.getDirectSuperConcepts()) {
                 addSuperinterface(extended.conceptWrapperInterfaceClass().parameterizedBy(nodeT))
             }
+
+            for (metaPropertyName in concept.language.languageSet.getMetaPropertyRoots(concept.fqName())) {
```
slisson, line 651: I guess this has to be replaced with `if (concept.getDirectSuperConcepts().isEmpty())` after changing `conceptFqName2RootMetaProperties`.

model-api/src/commonMain/kotlin/org/modelix/model/data/MetaModelData.kt
```diff
@@ -43,7 +43,12 @@ data class ConceptData(
     val references: List<ReferenceLinkData> = emptyList(),
     val extends: List<String> = emptyList(),
     override val deprecationMessage: String? = null,
-) : IDeprecatable
+    val metaProperties: MutableMap<String, String> = mutableMapOf(),
+) : IDeprecatable {
+    companion object {
+        const val ALIAS_KEY = "alias"
```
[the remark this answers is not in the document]
mhuster23: It's used in the metamodel-export MPS solution.
Should I keep it like this or just hardcode the string into the solution, since it's not used anywhere else right now?

model-api-gen/src/main/kotlin/org/modelix/metamodel/generator/MetaModelGenerator.kt
mhuster23: I had a similar approach earlier but in the case of MPS languages this can lead to problems, since interface concepts like `ScopeProvider` don't have any super concepts either. So the `val alias: String get() = null` would be generated into `C_BaseConcept` and `C_ScopeProvider` (among other interfaces). If a concept then both extends these interfaces the Kotlin Compiler is not able to disambiguate these functions even though their implementation is the same.

That's why I went for the approach with the map.

slisson: Here are some solutions.

### A. Repeat the default implementation in the interface where the conflict happens.

```kotlin
interface ConceptA {
    val alias: String? get() = null
}
interface ConceptB {
    val alias: String? get() = null
}
interface ConceptC : ConceptA, ConceptB {
    override val alias: String? get() = null
}
class ConceptCImpl : ConceptC {
}
```

### B. Don't generate a default implementation and let all concept implement it.

```kotlin
interface ConceptA {
    val alias: String?
}
interface ConceptB {
    val alias: String?
}
interface ConceptC : ConceptA, ConceptB {
}
class ConceptCImpl : ConceptC {
    override val alias: String? get() = null
}
```
### C. Generate a common interface for all concept

```kotlin
interface IConceptProperties {
    val alias: String? get() = null
}
interface ConceptA : IConceptProperties {
}
interface ConceptB : IConceptProperties {
}
interface ConceptC : ConceptA, ConceptB {
}
class ConceptCImpl : ConceptC {
    override val alias: String get() = "abc"
}
```

### D. Generate our own BaseConcept

Similar to option C we could generate a whole new concept that is the super concept of all other concept.
This guarantees that there is a single root in the concept hierarchy and might be useful in the future for solving similar problems.

When choosing this option the concept can just be added during the preprocessing of the languages and doesn't require any changes to the generator itself.

Both option C and D require to generate a unique name for the new interface/concept to avoid duplicate classes/interfaces on the classpath.

mhuster23: I went with suggestion C. In order to have a unique name for the generated interface, I let the user configure the name of the interface similar to `registrationHelperName`. This way name clashes can be avoided, even if generated APIs from multiple language sets are used.
If no name is specified, no concept meta-properties will be generated.

docs/global/modules/core/pages/reference/component-model-api-gen-gradle.adoc
```diff
@@ -100,7 +100,12 @@ Inside of the `metamodel` block the following settings can be configured.
 
 |`registrationHelperName`
 |String
-|Name of the registration helper
+|Name of the generated language registration helper
+
+|`conceptPropertiesInterfaceName`
+|String
+|Name of the generated interface, that contains the concept meta-properties of this language set.
+If null (default), neither the concept meta-properties nor the corresponding interface will be generated.
```
languitar, line 108:
```suggestion
If `null` (default), neither the concept meta-properties nor the corresponding interface will be generated.
```

languitar, line 108: Is there any overhead in generating one more interface? Or why do we want to include more conditional logic in the whole generation process? Isn't it more stuff to maintain if we make this conditional?

model-api-gen-gradle-test/src/test/kotlin/org/modelix/metamodel/generator/GeneratedApiTest.kt
```diff
@@ -55,15 +60,28 @@ class GeneratedApiTest {
...
+        assertTrue(foundDeprecatedNodeProperty)
+        assertTrue(foundDeprecatedNodeChildLink)
+        assertTrue(foundDeprecatedNodeReference)
+    }
+
+    @Test
+    fun `metaProperty alias is generated`() {
+        val hasAlias = IMetaConceptProperties::class.members.any { it.name == ConceptData.ALIAS_KEY }
+        assertTrue(hasAlias)
+        assertNull(C_BaseConcept.alias)
```
languitar, line 78: Isn't the last line already sufficient for this test? Without the interface being generated, this tests wouldn't compile.

model-api-gen/src/main/kotlin/org/modelix/metamodel/generator/MetaModelGenerator.kt
```diff
@@ -117,11 +122,41 @@ class MetaModelGenerator(val outputDir: Path, val nameConfig: NameConfig = NameC
             .write()
     }
 
+    private fun generateConceptMetaPropertiesInterface(languages: IProcessedLanguageSet) {
+        val fqName = checkNotNull(conceptPropertiesInterfaceName)
+        val interfaceName = ClassName(fqName.substringBeforeLast("."), fqName.substringAfterLast("."))
```
languitar, line 127: Might need some error handling with good messages to the (Gradle) user. Right now, this probably fails in case the name does not contain a period?

docs/global/modules/core/pages/reference/component-model-api-gen-gradle.adoc
mhuster23: This is mainly for backwards compatibility, so that generation doesn't break for existing users if no name is specified.

languitar: There could simply be a reasonable default?

model-api-gen-gradle-test/src/test/kotlin/org/modelix/metamodel/generator/GeneratedApiTest.kt
mhuster23: It's true, it wouldn't compile in that case. But this also ensures that the alias property exists in the interface and isn't just directly defined in the concepts.

docs/global/modules/core/pages/reference/component-model-api-gen-gradle.adoc
languitar: Ah, what about putting that interface into the model-api library itself instead of generating it?

languitar: Similar to that enum base class/interface ;-)

mhuster23: Because the meta properties are not known in advance. From MPS we only export the alias by default. But the user can add arbitrary meta properties in the ConceptData json, which will be generated into the interface.

languitar: I don't quite understand the reasoning. For enums we generically know that there's an "presentation", but for concepts we cannot assume that there's an optional "alias". Both things sound pretty similar to me.

mhuster23: Enums (including their presentation) were added to the json schema. Since the concept alias has no semantic meaning and is too MPS specific, we did not want to add it to the schema directly but still support it. That's why this PR adds the extremely generic metaInfo map for meta properties to concepts. Our export from MPS adds the alias meta property, but the generator does not know about this.
We do not know, which properties the user puts in there. That's why the interface must be generated.

slisson: https://issues.modelix.org/issue/MODELIX-555/Support-meta-properties-on-all-meta-model-elements

PR #246, opened 2023-09-18
Global license header configuration

slisson:


languitar:
The spotless Plugin is actually a source of potential configuration issues when setting up the project freshly in IntelliJ and it also contributes to the overall build time inside Gradle. I was wondering if it wouldn't be better to drop this aspect from the longish Gradle build and to instead use only pre-commit for this purpose? github.com/Lucas-C/pre-commit-hooks could be an option.

PR #247, opened 2023-09-18
build: suppress warnings from unused variables for source sets

languitar:
Source set access in multi-platform projects is expected to happen by delegation. Yet, the created vals are usually not accessed afterwards and result in compilation warnings when Gradle compiles the build file.

This commit adds suppressions for these warnings. As IntelliJ understands these warnings as unneeded, despite having an effect on command-line compilation, another suppression for the unneeded suppressions is needed.

Cf.: https://youtrack.jetbrains.com/issue/KT-38871

### DIFF
--- a/bulk-model-sync-lib/build.gradle.kts
+++ b/bulk-model-sync-lib/build.gradle.kts
@@ -16,6 +16,7 @@ kotlin {
             useJUnitPlatform()
         }
     }
+    @Suppress("UNUSED_VARIABLE", "KotlinRedundantDiagnosticSuppress")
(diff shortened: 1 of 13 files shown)


odzhychko:
LGTM

I created a ticket to remove the annotation after the Kotlin compiler with the appropriate fix is published.
https://issues.modelix.org/issue/MODELIX-544/Do-not-suppress-warnings-from-unused-variables-for-source-sets

odzhychko (approved):

languitar:
What is actually wrong here with the build?

PR #252, opened 2023-09-19
MODELIX-549 incorrect serialization of MPS node references

slisson:
They contained only the node ID instead of the whole reference:
  `mps:741118295176715139`
instead of
  `mps:r:54d7f6a3-9a78-48d5-97cc-b47cbb1f005c(my.model)/741118295176715139`

### DIFF
--- a/mps-model-adapters/src/main/kotlin/org/modelix/model/mpsadapters/MPSNodeReference.kt
+++ b/mps-model-adapters/src/main/kotlin/org/modelix/model/mpsadapters/MPSNodeReference.kt
@@ -51,7 +51,7 @@ object MPSNodeReferenceSerializer : INodeReferenceSerializerEx {
     override val supportedReferenceClasses: Set<KClass<out INodeReference>> = setOf(MPSNodeReference::class)
 
     override fun serialize(ref: INodeReference): String {
-        return (ref as MPSNodeReference).ref.nodeId.toString()
+        return SNodePointer.serialize((ref as MPSNodeReference).ref)
     }
 
     override fun deserialize(serialized: String): INodeReference {


model-api/src/commonMain/kotlin/org/modelix/model/api/INodeReferenceSerializer.kt
```diff
@@ -84,6 +84,7 @@ interface INodeReferenceSerializer {
             }
         }
 
+        @Deprecated("use INodeReference.serialize()", ReplaceWith("ref.serialize()"))
```
languitar, line 87: Any chance to lift out adding these deprecations and maybe the related refactorings into a separate commit with a separate rationale? That makes it easier to understand why they were added later on.

mps-model-adapters/src/main/kotlin/org/modelix/model/mpsadapters/MPSLanguageRepository.kt
```diff
@@ -27,17 +27,20 @@ import org.modelix.model.api.ILanguageRepository
 
 class MPSLanguageRepository(private val repository: SRepository) : ILanguageRepository {
     override fun resolveConcept(uid: String): IConcept? {
-        if (!uid.startsWith("mps:")) return null
+        for (prefix in listOf("mps:", "mps-concept:")) {
```
languitar, line 30: These prefixes seem to be repeated throughout the whole code base. Can we introduce constants for them?

languitar, line 30: Why is mps-concept added here in this PR? Is that a separate bug?

languitar (changes requested):
Thanks for the quick fix. Some thing inline. Apart from that, can we add a test that prevents this from happening again in the future?

languitar (approved):

PR #254, opened 2023-09-22
MODELIX-514 Make changes from the MPS model sync plugin visible to the ModelClientV2

slisson:
The old REST API now also uses the RepositoriesManager to write to branches in a safe way (in a transaction with conflict resolution) instead of just overwriting the existing value. A write to a key that matches the branch pattern is dispatched to the RepositoriesManager and not written directly to the key-value store anymore.

While trying to make the tests succeed I had to fix a few issues, especially a performance issue in LinearHistory that also was another reason for MODELIX-524.

There is a new test that writes concurrently using the old and new client. Some of the iterations always fail which is probably a bug in the old model-client. I disabled the test for now and will fix that in a separate PR ([MODELIX-554](https://issues.modelix.org/issue/MODELIX-554/Unstable-ModelClient-v1)).

### DIFF
--- a/model-client/src/commonMain/kotlin/org/modelix/model/client2/IModelClientV2.kt
+++ b/model-client/src/commonMain/kotlin/org/modelix/model/client2/IModelClientV2.kt
@@ -55,5 +55,11 @@ interface IModelClientV2 {
 
     suspend fun pullHash(branch: BranchReference): String
 
+    /**
+     * While `pull` returns immediately `poll` returns as soon as a new version, that is different from the given
+     * `lastKnownVersion`, is pushed to the server or after some timout specified by the server (usually ~30 seconds).
+     */
     suspend fun poll(branch: BranchReference, lastKnownVersion: IVersion?): IVersion
+
+    suspend fun pollHash(branch: BranchReference, lastKnownVersion: IVersion?): String
 }
--- a/model-client/src/commonMain/kotlin/org/modelix/model/client2/IModelClientV2.kt
+++ b/model-client/src/commonMain/kotlin/org/modelix/model/client2/IModelClientV2.kt
@@ -53,5 +53,7 @@ interface IModelClientV2 {
 
     suspend fun pull(branch: BranchReference, lastKnownVersion: IVersion?): IVersion
 
+    suspend fun pullHash(branch: BranchReference): String
+
     suspend fun poll(branch: BranchReference, lastKnownVersion: IVersion?): IVersion
 }
--- a/model-client/src/commonMain/kotlin/org/modelix/model/client2/IModelClientV2.kt
+++ b/model-client/src/commonMain/kotlin/org/modelix/model/client2/IModelClientV2.kt
@@ -57,7 +57,7 @@ interface IModelClientV2 {
 
     /**
      * While `pull` returns immediately `poll` returns as soon as a new version, that is different from the given
-     * `lastKnownVersion`, is pushed to the server or after some timout specified by the server (usually ~30 seconds).
+     * `lastKnownVersion`, is pushed to the server or after some timeout specified by the server (usually ~30 seconds).
      */
     suspend fun poll(branch: BranchReference, lastKnownVersion: IVersion?): IVersion
 
--- a/model-client/src/commonMain/kotlin/org/modelix/model/client2/ModelClientV2.kt
+++ b/model-client/src/commonMain/kotlin/org/modelix/model/client2/ModelClientV2.kt
@@ -173,7 +173,20 @@ class ModelClientV2(
             }
         }
         val receivedHash: String = response.body()
-//        LOG.debug { "${clientId.toString(16)}.pullHash($branch) -> $receivedHash" }
+        return receivedHash
+    }
+
+    override suspend fun pollHash(branch: BranchReference, lastKnownVersion: IVersion?): String {
+        val response = httpClient.get {
+            url {
+                takeFrom(baseUrl)
+                appendPathSegmentsEncodingSlash("repositories", branch.repositoryId.id, "branches", branch.branchName, "pollHash")
+                if (lastKnownVersion != null) {
+                    parameters["lastKnown"] = lastKnownVersion.getContentHash()
+                }
+            }
+        }
+        val receivedHash: String = response.body()
         return receivedHash
     }
 
--- a/model-client/src/commonMain/kotlin/org/modelix/model/client2/ModelClientV2.kt
+++ b/model-client/src/commonMain/kotlin/org/modelix/model/client2/ModelClientV2.kt
@@ -165,6 +165,18 @@ class ModelClientV2(
         return receivedVersion
     }
 
+    override suspend fun pullHash(branch: BranchReference): String {
+        val response = httpClient.get {
+            url {
+                takeFrom(baseUrl)
+                appendPathSegmentsEncodingSlash("repositories", branch.repositoryId.id, "branches", branch.branchName, "hash")
+            }
+        }
+        val receivedHash: String = response.body()
+//        LOG.debug { "${clientId.toString(16)}.pullHash($branch) -> $receivedHash" }
+        return receivedHash
+    }
+
     override suspend fun poll(branch: BranchReference, lastKnownVersion: IVersion?): IVersion {
         require(lastKnownVersion is CLVersion?)
         LOG.debug { "${clientId.toString(16)}.poll($branch, $lastKnownVersion)" }
--- a/model-datastructure/src/commonMain/kotlin/org/modelix/model/LinearHistory.kt
+++ b/model-datastructure/src/commonMain/kotlin/org/modelix/model/LinearHistory.kt
@@ -30,7 +30,7 @@ class LinearHistory(val baseVersionHash: String?) {
             collect(fromVersion)
         }
 
-        var result: List<Long> = ArrayList()
+        var result: List<Long> = emptyList()
 
         for (version in versions.values.filter { !it.isMerge() }.sortedBy { it.id }) {
             val descendantIds = collectAllDescendants(version.id).filter { !versions[it]!!.isMerge() }.sorted().toSet()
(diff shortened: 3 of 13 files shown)


model-datastructure/src/commonMain/kotlin/org/modelix/model/LinearHistory.kt
```diff
@@ -61,3 +68,73 @@ class LinearHistory(val baseVersionHash: String?) {
...
+
+class LinearHistory(val baseVersionHash: String?) {
+
+    val version2directDescendants: MutableMap<Long, Set<Long>> = HashMap()
+    val versions: MutableMap<Long, CLVersion> = HashMap()
+
+    /**
+     * Oldest version first
+     */
+    fun load(vararg fromVersions: CLVersion): List<CLVersion> {
```
odzhychko, line 80: suggestion: Could you add a comment, how versions are sorted/which properties the sorted result has?

One can understand the code of the sorting. It is a kind of topological sorting.
But still, it is hard to understand, which properties of the sorting are essential and guaranteed and which are only accidental.
One needs to follow and understand the existing usages, and even than that might not contain all the intended usages.

odzhychko (commented):
I commented on some initial observations. I am aware, that this still a draft PR.

model-client/src/commonMain/kotlin/org/modelix/model/client2/IModelClientV2.kt
```diff
@@ -53,5 +53,7 @@ interface IModelClientV2 {
 
     suspend fun pull(branch: BranchReference, lastKnownVersion: IVersion?): IVersion
 
+    suspend fun pullHash(branch: BranchReference): String
+
     suspend fun poll(branch: BranchReference, lastKnownVersion: IVersion?): IVersion
```
languitar, line 58: How do pull and poll differ here and why is there only a `pullHash` but not a `pollHash` function?

model-client/src/commonMain/kotlin/org/modelix/model/client2/ModelClientV2.kt
```diff
@@ -165,6 +165,18 @@ class ModelClientV2(
         return receivedVersion
     }
 
+    override suspend fun pullHash(branch: BranchReference): String {
+        val response = httpClient.get {
+            url {
+                takeFrom(baseUrl)
+                appendPathSegmentsEncodingSlash("repositories", branch.repositoryId.id, "branches", branch.branchName, "hash")
```
languitar, line 172: This repeats throughout the different methods implemented in this class (apart from the last argument). Refactoring the path creation for a repo and branch into a separate method would be good.

model-datastructure/src/commonMain/kotlin/org/modelix/model/LinearHistory.kt
```diff
@@ -5,7 +5,14 @@ import org.modelix.model.lazy.IDeserializingKeyValueStore
...
 
-class LinearHistory(val baseVersionHash: String?) {
+/**
+ * Was introduced in https://github.com/modelix/modelix/commit/19c74bed5921028af3ac3ee9d997fc1c4203ad44
+ * together with the UndoOp. The idea is that an undo should only revert changes if there is no other change that relies
+ * on it. In that case the undo should do nothing to not indirectly undo newer changes.
+ * For example, if you added a node and someone else started changing properties on the that node, your undo should not
+ * remove the node to not lose the property changes.
+ */
+class SlowLinearHistory(val baseVersionHash: String?) {
```
languitar: I don't understand the naming of the class given the comment. Is this a special implementation that support undo operations and is only accidentally slow due to the algorithm?

model-datastructure/src/commonTest/kotlin/LinearHistoryTest.kt
```diff
@@ -0,0 +1,209 @@
...
+import org.modelix.model.operations.IOperation
+import org.modelix.model.persistent.MapBaseStore
+import kotlin.test.Test
+import kotlin.test.assertEquals
+
+class LinearHistoryTest {
+    val initialTree = CLTree.builder(ObjectStoreCache(MapBaseStore())).repositoryId("LinearHistoryTest").build()
+
+    @Test
+    fun simpleTest1() {
```
languitar, line 31: Can this get a better name on which condition is being tested here? Something like "stacks changes without conflicts" or whatever is actually tested here. These test case names really help understanding the algorithm for new code readers.

model-datastructure/src/commonTest/kotlin/LinearHistoryTest.kt
```diff
@@ -0,0 +1,209 @@
...
+        val v20 = version(20, null)
+        val v21 = version(21, null)
+
+        assertHistory(v20, v21, listOf(v20, v21))
+    }
+
+    @Test
+    fun simpleTest2() {
```
languitar, line 39: As above.

model-datastructure/src/commonTest/kotlin/LinearHistoryTest.kt
```diff
@@ -0,0 +1,209 @@
...
+        val v10 = version(10, null)
+        val v20 = version(20, v10)
+        val v21 = version(21, v10)
+
+        assertHistory(v20, v21, listOf(v20, v21))
+    }
+
+    @Test
+    fun linearHistoryPerformance() {
```
languitar, line 48: If there was ever a bug, these random variable names and IDs would make debugging pretty annoying.

model-server/src/test/kotlin/org/modelix/model/server/ReplicatedRepositoryTest.kt
```diff
@@ -194,6 +197,136 @@ class ReplicatedRepositoryTest {
...
+        val serverVersion = clients[0].pull(branchId, null)
+        val childrenOnServer = getChildren(PBranch(serverVersion.getTree(), IdGeneratorDummy()))
+
+        assertEquals(createdNodes, childrenOnServer)
+
+        for (client in clients) {
+            assertEquals(createdNodes, getChildren(localVersions[client]!!))
+        }
+
+        println("all successful")
```
languitar, line 329: Can we drop this?

model-server/src/test/kotlin/org/modelix/model/server/ReplicatedRepositoryTest.kt
```diff
@@ -327,6 +329,101 @@ class ReplicatedRepositoryTest {
...
+            fun getChildren(model: ReplicatedModel): SortedSet<String> {
+                return getChildren(model.getBranch())
+            }
+            fun getChildren(model: ReplicatedRepository): SortedSet<String> {
+                return getChildren(model.branch)
+            }
+
+            assertEquals((clients.size + v1clients.size) * 10, createdNodes.size)
+
+            println("writing done. waiting for convergence.")
```
languitar, line 395: Can we drop this?

model-server/src/test/kotlin/org/modelix/model/server/ReplicatedRepositoryTest.kt
```diff
@@ -327,6 +329,101 @@ class ReplicatedRepositoryTest {
         println("all successful")
     }
 
+    @RepeatedTest(value = 10)
```
languitar, line 333: Why is it necessary to repeat this test? Overall, this test looks fairly complex. Is there no easier way to show that change propagate between different clients?

model-server/src/test/kotlin/org/modelix/model/server/ReplicatedRepositoryTest.kt
```diff
@@ -329,6 +329,7 @@ class ReplicatedRepositoryTest {
         println("all successful")
     }
 
+    @Ignore
```
languitar, line 332: Maybe a [`@Disabled`](https://junit.org/junit5/docs/current/api/org.junit.jupiter.api/org/junit/jupiter/api/Disabled.html) with a description for why this is actually disabled would be a better fit. If kotlin test shall remain the target, a comment would be good.

model-server/src/main/kotlin/org/modelix/model/server/handlers/KeyValueLikeModelServer.kt
```diff
@@ -298,7 +280,46 @@ class KeyValueLikeModelServer(val storeClient: IStoreClient) {
...
+        val branchChanges = LinkedHashMap<BranchReference, String>()
+        val userDefinedEntries = LinkedHashMap<String, String?>()
+
+        for ((key, value) in newEntries) {
+            when {
+                HashUtil.isSha256(key) -> {
+                    hashedObjects[key] = value ?: throw IllegalArgumentException("No value provided for $key")
+                }
+                BranchReference.tryParseBranch(key) != null -> {
+                    branchChanges[BranchReference.tryParseBranch(key)!!] = value ?: throw IllegalArgumentException("Deleting branch $key not allowed")
```
odzhychko, line 297: question: Could and should we delete the branch in v1 and propagate it to v2?

I have seen users of the v1 server deleting branches like this.

```java
final RestWebModelClient client = ....
client.getAsyncStore().put(repositoryId.getBranchReference("master").getKey(), null);
```

This new check would break it. Or am I missing something?

model-datastructure/src/commonTest/kotlin/LinearHistoryTest.kt
slisson: This test is just a dump from one of the other failing tests to make debugging easier.

model-client/src/commonMain/kotlin/org/modelix/model/client2/IModelClientV2.kt
```diff
@@ -55,5 +55,11 @@ interface IModelClientV2 {
 
     suspend fun pullHash(branch: BranchReference): String
 
+    /**
+     * While `pull` returns immediately `poll` returns as soon as a new version, that is different from the given
+     * `lastKnownVersion`, is pushed to the server or after some timout specified by the server (usually ~30 seconds).
```
languitar, line 60:
```suggestion
     * `lastKnownVersion`, is pushed to the server or after some timeout specified by the server (usually ~30 seconds).
```

slisson:
The reason why the [3.1.1](https://github.com/modelix/modelix.core/actions/runs/6313733024) release failed should also be fixed by this PR.

PR #256, opened 2023-09-25
MODELIX-552 (2) ClassCastException executing query

slisson:
The serializer for the input of a query was computed with the assumption that a query is only used once and only one possible input step exists. After introducing `QueryCallStep` this isn't true anymore and the correct serializer depends on the context.

Extended the parameter of IProducingStep.getOutputSerializer to also carry the correct serializer for the QueryInput step.


languitar (changes requested):
A test to reproduce the issue and to prevent it in the future would be good.

languitar (approved):

PR #258, opened 2023-09-26
ci: include detekt with GitHub code security analysis

languitar:
This PR includes [detekt](https://detekt.dev/docs/intro/) as a linter in the project. In contrast to ktlint, detect is more focused on coding problems and might provide some valuable hints. I have disable failing the build when detekt issues exist due to the current amount of violations that already exist in the current code base.

Detekt results are passed into the GitHub Code Scanning feature. Effectively this means that PRs will receive inline annotations for detected problems and we can use these as a hint when reviewing PRs.

### DIFF
--- a/.detekt.yml
+++ b/.detekt.yml
@@ -1,0 +1,12 @@
+style:
+  # handled by ktlint
+  MaxLineLength:
+    active: false
+  ModifierOrder:
+    active: false
+
+  # No need for this
+  ReturnCount:
+    active: false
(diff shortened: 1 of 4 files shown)


languitar:
There's still something wrong with the reported file paths. Trying to fix that.

languitar:
Now paths are correct

odzhychko (dismissed):

odzhychko (approved):

PR #259, opened 2023-09-27
MODELIX-551 Support pagination

slisson:
Added .drop and .take steps which can be combined to select a range of elements.


languitar (dismissed):
Thanks for the quick implementation.

Just to be sure: effectively, we do not load all data we drop due to the stream nature of the input? Or will MPS / the model-server load all data before applying the drop step?

languitar (approved):

slisson:
> Thanks for the quick implementation.
> 
> Just to be sure: effectively, we do not load all data we drop due to the stream nature of the input? Or will MPS / the model-server load all data before applying the drop step?

Yes, most of the excluded data shouldn't be loaded at all, which means the execution of the query should be noticeably faster.

PR #261, opened 2023-09-28
build(model-api-gen-gradle): introduce subprojects in the test project

odzhychko:
The test project is restructured into subprojects in anticipation of an additional test project for Vue.js code generation in the future.


model-api-gen-gradle-test/build.gradle.kts
```diff
@@ -1,130 +1,16 @@
...
-        languageClass.prefix = "L_"
-        languageClass.baseNameConversion = { it.replace(".", "_") }
-        typedNode.prefix = ""
-        typedNodeImpl.suffix = "Impl"
+subprojects {
+    plugins.withType<NodePlugin> {
+        project.extensions.configure<NodeExtension> {
+            version.set(libs.versions.node)
+            download.set(true)
+        }
```
mhuster23, line 14: Is it necessary to configure the plugin here?
Isn't it just used in `/typescript-generation`?

model-api-gen-gradle-test/typescript-generation/build.gradle.kts
```diff
@@ -0,0 +1,31 @@
...
+
+plugins {
+    base
+    alias(libs.plugins.node)
+    id("org.modelix.model-api-gen") apply false
+}
+
+node {
+    version.set("18.3.0")
+    npmVersion.set("8.11.0")
```
mhuster23, line 11: We should probably use versions from the catalogue.

languitar: This is not a Maven artifact or Gradle plugin. I don't think arbitrary properties can be defined in the catalog.

model-api-gen-gradle-test/build.gradle.kts
odzhychko: You are right. For now, it is only needed for `typescript-generation`.

But it will be needed for the coming `vue-generation`, so I already extracted it.

model-api-gen-gradle-test/typescript-generation/build.gradle.kts
odzhychko: Arbitrary versions can be defined in the catalog and be used. 
(I'm not sure, if it is intended, but it works.)

I actually did it in this PR in `model-api-gen-gradle-test/build.gradle.kts`.

```
subprojects {
    plugins.withType<NodePlugin> {
        project.extensions.configure<NodeExtension> {
            version.set(libs.versions.node)
            download.set(true)
        }
}
```

---

I removed the configuration here in `model-api-gen-gradle-test/typescript-generation/build.gradle.kts`, because the configuration in `model-api-gen-gradle-test/build.gradle.kts` is enough.

mhuster23 (approved):

PR #265, opened 2023-09-29
MODELIX-560 bulk-model-sync quality of life improvements

mhuster23:
This PR adresses 2 pain points:

- Including multiple modules required listing them individually. As a shortcut they can now be included by specifying a prefix.
- The sync was almost like a black box. This PR adds some progress indicators.

### DIFF
--- a/bulk-model-sync-lib/src/commonMain/kotlin/org/modelix/model/sync/bulk/ModelImporter.kt
+++ b/bulk-model-sync-lib/src/commonMain/kotlin/org/modelix/model/sync/bulk/ModelImporter.kt
@@ -16,6 +16,7 @@
 
 package org.modelix.model.sync.bulk
 
+import mu.KotlinLogging
 import org.modelix.model.api.ConceptReference
 import org.modelix.model.api.INode
 import org.modelix.model.api.INodeReference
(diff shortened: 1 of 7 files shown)


bulk-model-sync-lib/src/commonMain/kotlin/org/modelix/model/sync/bulk/ModelImporter.kt
```diff
@@ -16,6 +16,7 @@
 
 package org.modelix.model.sync.bulk
 
+import mu.KotlinLogging
```
languitar, line 19: Why was this logging library chosen?

docs/global/modules/core/pages/reference/component-bulk-model-sync-gradle.adoc
```diff
@@ -70,6 +70,10 @@ This means that only a minimal amount of write operations is used to update the
 |`includeModule`
 |String
 |Includes the module specified by the given fully qualified name in the synchronisation process.
+
+|`includeModulesByPrefix`
+|String
+|Includes all modules, whose fully qualified name starts with the given prefix, in the synchronisation process.
```
languitar, line 76: We should decide once for all modelix if we speak British or American English ;-)

languitar (approved):
Looks fine to me. Feel free to close the open threads, which are slight remarks.

As a small side note: this PR addresses two distinct concerns: prefix config and progress logging. I would have preferred two separate PRs for easier reviewing.

bulk-model-sync-lib/src/commonMain/kotlin/org/modelix/model/sync/bulk/ModelImporter.kt
mhuster23: It's just the one we are already using across the core.

PR #266, opened 2023-09-29
Feature/js model client

odzhychko:
The model client in Kotlin is multiplatform, but not usable from JavaScript because it
does not export the appropriate functionality for usage from JavaScript and needs to be used inside
coroutines, which cannot be directly created in JavaScript.

This PR is part of [MODELIX-500](https://issues.modelix.org/issue/MODELIX-500)
The API of the package is marked as experimental. I intend to finalize it when finishing the ticket [MODELIX-500](https://issues.modelix.org/issue/MODELIX-500).

### DIFF
--- a/model-client/build.gradle.kts
+++ b/model-client/build.gradle.kts
@@ -4,6 +4,7 @@ plugins {
     id("com.diffplug.spotless")
     `java-library`
     jacoco
+    alias(libs.plugins.npm.publish)
 }
 
 java {
@@ -34,6 +35,8 @@ kotlin {
                 }
             }
         }
+        binaries.library()
+        generateTypeScriptDefinitions()
         useCommonJs()
     }
     @Suppress("UNUSED_VARIABLE", "KotlinRedundantDiagnosticSuppress")
@@ -84,6 +87,7 @@ kotlin {
             }
         }
         val jsMain by getting {
+            languageSettings.optIn("kotlin.js.ExperimentalJsExport")
             dependencies {
                 implementation(kotlin("stdlib-js"))
                 implementation(npm("uuid", "^8.3.0"))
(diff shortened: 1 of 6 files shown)


js-model-client/build.gradle.kts
slisson, line 1: We don't need a separate subproject for a JS model client. Just merge it into the existing model-client.

Also, there is Gradle plugin for publishing Kotlin projects as npm modules that simplifies this Gradle script a lot.

See https://github.com/modelix/modelix.core/blob/55d0ffd00108a5a2208441e85c8914f10eced7d0/model-client/build.gradle.kts#L138-L154
and https://github.com/modelix/modelix.core/blob/55d0ffd00108a5a2208441e85c8914f10eced7d0/gradle/libs.versions.toml#L19

slisson (changes requested):

js-model-client/build.gradle.kts
odzhychko: Done.
The PR can be reviewed again.

model-client/build.gradle.kts
```diff
@@ -126,3 +130,81 @@ spotless {
         )
     }
 }
+
+val productionLibraryByKotlinOutputDirectory = layout.buildDirectory.dir("compileSync/js/main/productionLibrary/kotlin")
+val preparedProductionLibraryOutputDirectory = layout.buildDirectory.dir("npmPublication")
+
+val patchTypesScriptInProductionLibrary = tasks.register("patchTypesScriptInProductionLibrary") {
```
languitar, line 137: As far as I can tell, this custom tasks is not marked to be cacheable. Any chance to enable caching here to avoid further extending the build times?

model-client/src/jsTest/kotlin/org/modelix/model/client2/ClientJSTest.kt
```diff
@@ -0,0 +1,108 @@
...
+import GeneratedConcept
+import org.modelix.kotlin.utils.UnstableModelixFeature
+import kotlin.test.Test
+import kotlin.test.assertEquals
+
+@OptIn(UnstableModelixFeature::class)
+class ClientJSTest {
+
+    @Test
+    fun canAddChildrenWithUnregisterConcept() {
```
languitar, line 28:
```suggestion
    fun canAddChildrenWithUnregisteredConcept() {
```

Is that the intended meaning of the test case?

model-client/build.gradle.kts
odzhychko: Thanks for the catch.

I already added `inputs` and `outputs`, but did not enable caching.

model-client/src/jsTest/kotlin/org/modelix/model/client2/ClientJSTest.kt
odzhychko: Done.

And yes, that was the intended meaning.

model-client/build.gradle.kts
odzhychko: Done.

model-client/src/jsTest/kotlin/org/modelix/model/client2/ClientJSTest.kt
```diff
@@ -0,0 +1,108 @@
...
+                ]
+            }
+        }
+        """.trimIndent()
+        val rootNode = loadModelsFromJson(arrayOf(data), {})
+        val child0 = rootNode.getAllChildren()[0]
+        val child1 = rootNode.getAllChildren()[1]
+        child0.setReferenceTargetRef("aReference", child1.getReference())
+        assertEquals(child0.getReferenceTargetNode("aReference"), child1)
+    }
```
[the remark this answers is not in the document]
languitar: That would be multiple test cases ;-)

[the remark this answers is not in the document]
odzhychko: I do not like the consequences of it, but you are right...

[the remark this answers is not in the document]
languitar: Writing concise test cases is more like an art. And if done, right, these test cases then really help documenting and for debugging because they give detailed feedback on which requirements are fulfilled and which ones are broken after a change. https://www.telerik.com/blogs/why-isolating-your-unit-tests-matters

slisson:
We should avoid having too much JS specific code, but add more @JsExport annotations to the sources in commonMain. Also getting rid of ts-model-api and re-implement it in Kotlin is something I explored in https://github.com/modelix/modelix.core/pull/201 and seems to allow reusing more of the common code. I will continue working on that PR first and then we can update or close this PR.

slisson (approved):

PR #270, opened 2023-10-02
MODELIX-562 Improve model-server content explorer readability

mhuster23:


model-server/src/main/kotlin/org/modelix/model/server/handlers/ContentExplorer.kt
```diff
@@ -138,8 +136,12 @@ class ContentExplorer(private val client: IModelClient, private val repoManager:
             div("nameField") {
                 attributes["data-nodeid"] = node.nodeId.toString()
                 b {
+                    val namePropertyUID = BuiltinLanguages.jetbrains_mps_lang_core.INamedConcept.name.getUID()
+                    val namedConceptRole = node.getPropertyRoles().find { it.endsWith(namePropertyUID) }
```
languitar, line 140: Is there no more structured way to access this? Why is the `endsWith` required here?

model-server/src/main/kotlin/org/modelix/model/server/handlers/ContentExplorer.kt
```diff
@@ -138,8 +136,12 @@ class ContentExplorer(private val client: IModelClient, private val repoManager:
...
                     if (node.getPropertyRoles().contains("name")) {
                         +"${node.getPropertyValue("name")}"
+                    } else if (namedConceptRole != null) {
+                        +"${node.getPropertyValue(namedConceptRole)}"
```
languitar, line 144: Is this ordering sensible? If something is a named concept, shouldn't we take that value with preference?

model-server/src/main/kotlin/org/modelix/model/server/handlers/ContentExplorer.kt
```diff
@@ -202,12 +204,13 @@ class ContentExplorer(private val client: IModelClient, private val repoManager:
                         th { +"Value" }
                     }
                 }
-                for (referenceRole in node.getReferenceRoles()) {
-                    tr {
-                        td { +referenceRole }
-                        td {
-                            // TODO MODELIX-387
```
languitar, line 209: Is this ticket fixed after the PR? Not mentioned in the PR description.

mhuster23: This PR was only possible, because that issue was fixed beforehand. It's part of this PR's issue description.

model-server/src/main/kotlin/org/modelix/model/server/handlers/ContentExplorer.kt
mhuster23: On second thought, it's not required.

model-server/src/main/kotlin/org/modelix/model/server/handlers/ContentExplorer.kt
languitar: Maybe adding a comment on why ordering was chosen this way would have been good, too. The next one might be wondering again.

languitar (approved):

PR #273, opened 2023-10-04
Integration of (light-)model-client with modelix.incremental

slisson:
Dynamic dependency tracking is a technique often used by UI frameworks for updating the UI automatically. [modelix.incremental](https://github.com/modelix/modelix.incremental/) provides some generic infrastructure for that. With the changes in this PR the [modelix.text-editor](https://github.com/modelix/modelix.text-editor) can work on models managed by the (light-)model-client.

The vue.js integration can also be simplified by just registering a listener in [DependencyTracking](https://github.com/modelix/modelix.incremental/blob/master/src/commonMain/kotlin/org/modelix/incremental/DependencyTracking.kt) and forwarding the read/write events to vue.js.

### DIFF
--- a/light-model-client/src/commonMain/kotlin/org/modelix/client/light/LightModelClient.kt
+++ b/light-model-client/src/commonMain/kotlin/org/modelix/client/light/LightModelClient.kt
@@ -5,6 +5,9 @@ import io.ktor.client.engine.HttpClientEngine
 import io.ktor.client.engine.HttpClientEngineFactory
 import io.ktor.client.plugins.websocket.WebSockets
 import kotlinx.coroutines.delay
+import org.modelix.incremental.DependencyTracking
+import org.modelix.incremental.IStateVariableGroup
+import org.modelix.incremental.IStateVariableReference
 import org.modelix.model.api.ConceptReference
 import org.modelix.model.api.IBranch
 import org.modelix.model.api.IConcept
@@ -60,9 +63,9 @@ class LightModelClient internal constructor(
     val autoFilterNonLoadedNodes: Boolean,
     val debugName: String = "",
     val modelQLClient: ModelQLClient? = null,
-) {
+) : IStateVariableGroup {
 
-    private val nodes: MutableMap<NodeId, NodeData> = HashMap()
+    private val nodes = NodesMap<NodeData>(this)
     private val area = Area()
     private var areaListeners: Set<IAreaListener> = emptySet()
     private var repositoryId: String? = null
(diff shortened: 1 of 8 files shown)


model-client/src/commonMain/kotlin/org/modelix/model/IncrementalBranch.kt
```diff
@@ -0,0 +1,488 @@
...
+
+        override fun addNewChild(
+            parentId: Long,
+            role: String?,
+            index: Int,
+            childId: Long,
+            concept: IConceptReference?,
+        ) {
+            transaction.addNewChild(parentId, role, index, childId, concept)
+            modified(ChildrenDependency(this@IncrementalBranch, parentId, role))
```
odzhychko, line 293: Do we not need a call to `modified(UnclassifiedNodeDependency(this@IncrementalBranch, nodeId))`, becauase of `.containsNode`?

model-api-gen-runtime/src/commonMain/kotlin/org/modelix/metamodel/ITypedNode.kt
```diff
@@ -15,7 +15,7 @@ interface ITypedNode {
 fun ITypedNode.untyped() = unwrap()
 fun ITypedNode.untypedConcept() = _concept.untyped()
 fun ITypedNode.typedConcept() = _concept
-fun ITypedNode.getPropertyValue(property: IProperty): String? = unwrap().getPropertyValue(property.name)
```
odzhychko, line 18: How does calling `getPropertyValue(property: IProperty)` instead of `getPropertyValue(role: String)` fixes the underlying bug?

Wouldn't the following example not still be broken?

```
val typedNode: ITypedNode = ...
typedNode.unwrap().getPropertyValue("someRole")
```

light-model-client/src/commonMain/kotlin/org/modelix/client/light/LightModelClient.kt
```diff
@@ -901,3 +907,66 @@ fun NodeData.asUpdateData(): NodeUpdateData {
...
+            DependencyTracking.modified(NodeDataDependency(client, key))
+        }
+    }
+
+    fun containsKey(key: NodeId): Boolean {
+        DependencyTracking.accessed(NodeDataDependency(client, key))
+        return map.containsKey(key)
+    }
+
+    fun isNotEmpty(): Boolean {
```
odzhychko, line 958: `isNotEmpty` is called in `hasTemporaryIds`. 
Why do we not need to track access and modifications for it? / How is it used? / Might it be used in an incremental function?

```kotlin
    fun hasTemporaryIds(): Boolean = requiresRead {
        temporaryNodeAdapters.isNotEmpty() || nodesReferencingTemporaryIds.isNotEmpty()
    }
```

model-client/src/commonMain/kotlin/org/modelix/model/IncrementalBranch.kt
slisson: Yes, that was indeed incorrect.

model-api-gen-runtime/src/commonMain/kotlin/org/modelix/metamodel/ITypedNode.kt
slisson: `INode.getPropertyValue(String)` is deprecated and will not get fixed.

light-model-client/src/commonMain/kotlin/org/modelix/client/light/LightModelClient.kt
slisson: Correct, but `hasTemporaryIds` isn't used anywhere. Just removed the dead code.

PR #279, opened 2023-10-09
MODELIX-561 Import into model-server is unsuccessful

mhuster23:
This PR fixes the issue, that imported models don't show up in the model-server.
Instead of using ReplicatedModel, which is intended for real-time collaboration (and doesn't seem to be able to handle the huge amount of changes that happen during the bulk sync), a new `runWrite` function is used to push the changes to the server.
By calling this method inside of `runBlocking` it is ensured, that the changes have been successfully pushed to the model-server.

### DIFF
--- a/bulk-model-sync-gradle-test/src/test/kotlin/org/modelix/model/sync/bulk/gradle/test/PushTest.kt
+++ b/bulk-model-sync-gradle-test/src/test/kotlin/org/modelix/model/sync/bulk/gradle/test/PushTest.kt
@@ -10,10 +10,10 @@ import org.modelix.metamodel.TypedLanguagesRegistry
 import org.modelix.metamodel.typed
 import org.modelix.model.ModelFacade
 import org.modelix.model.api.ConceptReference
-import org.modelix.model.api.IBranch
 import org.modelix.model.api.getDescendants
 import org.modelix.model.api.getRootNode
 import org.modelix.model.client2.ModelClientV2PlatformSpecificBuilder
+import org.modelix.model.client2.ReplicatedModel
 import org.modelix.model.client2.getReplicatedModel
 import org.modelix.model.data.ModelData
 import org.modelix.model.data.NodeData
@@ -40,21 +40,21 @@ class PushTest {
         val repoId = RepositoryId("ci-test")
         val branchName = "master"
         val url = "http://0.0.0.0:${Main.DEFAULT_PORT}/v2"
-        val branchRef = ModelFacade.createBranchReference(repoId, branchName)
-        val client = ModelClientV2PlatformSpecificBuilder().url(url).build()
 
-        val branch = runBlocking {
-            client.init()
-            client.getReplicatedModel(branchRef).start()
-        }
+        val branchRef = ModelFacade.createBranchReference(repoId, branchName)
+        val client = ModelClientV2PlatformSpecificBuilder().url(url).build().apply { runBlocking { init() } }
+        val replicatedModel = client.getReplicatedModel(branchRef)
+        val branch = runBlocking { replicatedModel.start() }
+
         branch.runRead {
             assertContentEquals(inputModel.root.children, branch.getRootNode().allChildren.map { it.asExported() })
         }
 
-        applyChangesForPullTest(branch)
+        applyChangesForPullTest(replicatedModel)
     }
 
-    private fun applyChangesForPullTest(branch: IBranch) {
+    private fun applyChangesForPullTest(replicatedModel: ReplicatedModel) {
+        val branch = replicatedModel.getBranch()
         branch.runWrite {
             val graphNodes = branch.getRootNode()
                 .getDescendants(false)
@@ -67,5 +67,6 @@ class PushTest {
             graphNodes[2].name = "Z"
         }
         sleep(5000) // changes are pushed asynchronously to the server. wait for the propagation
+        replicatedModel.dispose()
     }
 }
--- a/bulk-model-sync-gradle-test/src/test/kotlin/org/modelix/model/sync/bulk/gradle/test/PushTest.kt
+++ b/bulk-model-sync-gradle-test/src/test/kotlin/org/modelix/model/sync/bulk/gradle/test/PushTest.kt
@@ -12,14 +12,15 @@ import org.modelix.model.ModelFacade
 import org.modelix.model.api.ConceptReference
 import org.modelix.model.api.getDescendants
 import org.modelix.model.api.getRootNode
+import org.modelix.model.client2.IModelClientV2
 import org.modelix.model.client2.ModelClientV2PlatformSpecificBuilder
-import org.modelix.model.client2.ReplicatedModel
 import org.modelix.model.client2.getReplicatedModel
+import org.modelix.model.client2.runWrite
 import org.modelix.model.data.ModelData
 import org.modelix.model.data.NodeData
+import org.modelix.model.lazy.BranchReference
 import org.modelix.model.lazy.RepositoryId
 import org.modelix.model.server.Main
-import org.modelix.model.sleep
 import org.modelix.model.sync.bulk.asExported
 import java.io.File
 import kotlin.test.assertContentEquals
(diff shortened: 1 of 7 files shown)


bulk-model-sync-gradle-test/src/test/kotlin/org/modelix/model/sync/bulk/gradle/test/PushTest.kt
```diff
@@ -40,32 +41,33 @@ class PushTest {
         val repoId = RepositoryId("ci-test")
         val branchName = "master"
         val url = "http://0.0.0.0:${Main.DEFAULT_PORT}/v2"
+
         val branchRef = ModelFacade.createBranchReference(repoId, branchName)
-        val client = ModelClientV2PlatformSpecificBuilder().url(url).build()
+        val client = ModelClientV2PlatformSpecificBuilder().url(url).build().apply { runBlocking { init() } }
+        val replicatedModel = client.getReplicatedModel(branchRef)
+        val branch = runBlocking { replicatedModel.start() }
```
languitar, line 48: Why are these changes necessary? Why are multiple `runBlocking` blocks required instead of only one? Maybe a comment here or in the commit would help understanding this change.

bulk-model-sync-lib/src/commonMain/kotlin/org/modelix/model/sync/bulk/Util.kt
```diff
@@ -16,9 +16,49 @@
...
+private fun countReferences(data: NodeData): Int =
+    data.references.size + data.children.sumOf { countReferences(it) }
+
+private fun countConcepts(data: NodeData): Int {
+    val set = mutableSetOf<String>()
+    countConceptsRec(data, set)
+    return set.size
+}
+
+private fun countConceptsRec(data: NodeData, set: MutableSet<String>) {
```
languitar, line 49:
```suggestion
    collectConcepts(data, set)
    return set.size
}

private fun collectConcepts(data: NodeData, set: MutableSet<String>) {
```

That function actually doesn't count

bulk-model-sync-lib/src/commonMain/kotlin/org/modelix/model/sync/bulk/Util.kt
```diff
@@ -16,9 +16,49 @@
...
+    countSpecificConcept(data, BuiltinLanguages.MPSRepositoryConcepts.Model)
+
+private fun countMpsModules(data: NodeData) =
+    countSpecificConcept(data, BuiltinLanguages.MPSRepositoryConcepts.Module)
+
+private fun countSpecificConcept(data: NodeData, concept: SimpleConcept): Int {
+    var count = if (data.concept == concept.getUID()) 1 else 0
+    count += data.children.sumOf { countSpecificConcept(it, concept) }
+    return count
+}
```
languitar, line 64: This could become a bit expensive because the same tree traversal is performed twice, once for models and once for modules.

model-client/src/commonMain/kotlin/org/modelix/model/client2/ReplicatedModel.kt
```diff
@@ -21,6 +21,9 @@ import org.modelix.model.lazy.CLTree
 import org.modelix.model.lazy.CLVersion
 import org.modelix.model.operations.OTBranch
 
+/**
+ * Dispose should be called on this, as otherwise a regular polling will go on.
+ */
 class ReplicatedModel(
```
languitar, line 27: Can't we make ReplicatedModel implement Closeable for being able to [`use`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.io/use.html) it. That would make resource management a lot safer.

languitar (changes requested):
I am a bit confused about the commit types used in this PR:

* [fix(model-client): add runWrite extension function](https://github.com/modelix/modelix.core/pull/279/commits/f393f196c0a614f6427bb413dbfb0a332f0628d1): what is actually fixed here? From the commit this sounds like a refactoring? Otherwise, the commit subject should explain what is being fixed
* [fix(bulk-model-sync-lib): add logging for import size](https://github.com/modelix/modelix.core/pull/279/commits/a0fd3ec9e53b7ac8b2703a2606f1d70843e32ddf) - why is that a fix? That is adding a feature?
* For the two commits mentioning the ReplicatedModel, I need some explanations on the "why" in the commit messages

languitar:
One more note: a short summary in the PR description here would also help. I still can't see what actually is the fix here.

model-client/src/commonMain/kotlin/org/modelix/model/client2/ReplicatedModel.kt
mhuster23: That's a good idea. I created [MODELIX-568](https://issues.modelix.org/issue/MODELIX-568/) for this. We also did sth similar for the ModelClient itself in https://github.com/modelix/modelix.core/pull/235

languitar:
> Some context for the commit types: [fix(model-client): add runWrite extension function](https://github.com/modelix/modelix.core/pull/279/commits/f393f196c0a614f6427bb413dbfb0a332f0628d1) is part of the actual fix for this PR, it is a new method that is used in the commit after that to do the actual fix. I separated it because it affects a separate component and to correctly scope the co-authorship.

Wouldn't this commit then be a `feat`? You offer a new method for consumers. In any case, code and/or commit require some more explanations why that method is added and for what purpose it should be used.

> [fix(bulk-model-sync-lib): add logging for import size](https://github.com/modelix/modelix.core/pull/279/commits/a0fd3ec9e53b7ac8b2703a2606f1d70843e32ddf) can technically be seen as a feature, since these are new metrics. However, they are in a very fine log level and not intended for everyday use. I did not feel like a minor version bump is justified for this bit of additional logging.

If we see it as something completely irrelevant to the user, `chore` could be used? Otherwise, as a client I'd be happy to see a new feature for more detailed logging. From a fix I always assume it fixes something that was broken. And that's not the case here.

bulk-model-sync-gradle-test/src/test/kotlin/org/modelix/model/sync/bulk/gradle/test/PushTest.kt
mhuster23: The split is necessary to be able to dispose the replicated model. I separated the commits to make it clearer.

bulk-model-sync-lib/src/commonMain/kotlin/org/modelix/model/sync/bulk/Util.kt
languitar: What about this one?

bulk-model-sync-lib/src/jvmMain/kotlin/org/modelix/model/sync/bulk/PlatformSpecific.kt
```diff
@@ -35,6 +35,12 @@ fun ModelImporter.import(jsonFile: File) {
     import(data)
 }
 
+fun ModelImporter.importFilesAsRootChildren(files: Collection<File>) {
+    val models = files.map { ModelData.fromJson(it.readText()) }
+    import(mergeModelData(*models.toTypedArray()))
```
languitar, line 40: Now we got the same problem one level down, right?

bulk-model-sync-lib/src/commonMain/kotlin/org/modelix/model/sync/bulk/Util.kt
mhuster23: This function doesn't exist anymore. The concepts are now also collect in the new `measureImportSize()` function.

languitar: Weird, the thread was still open when I browser through the individual commits.

bulk-model-sync-lib/src/jvmMain/kotlin/org/modelix/model/sync/bulk/PlatformSpecific.kt
languitar: btw, if the array changes are solely related to performance, the angular conventions, to which conventional-commits refer for further commit types, also contain a `perf` commit type: https://github.com/angular/angular/blob/22b96b9/CONTRIBUTING.md#type

languitar (approved):

PR #293, opened 2023-10-23
replace deprecated Project.buildDir usage in Gradle 

odzhychko:
Replace the usage of `Project.buildDir` in our build scripts and in our provided plugins.

See https://docs.gradle.org/current/userguide/upgrading_version_8.html#deprecations_2

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -101,7 +101,7 @@ subprojects {
         reports {
             sarif.required.set(true)
             // This is required for the GitHub upload action to easily find all sarif files in a single directory.
-            sarif.outputLocation.set(parentProject.buildDir.resolve("reports/detekt/${project.name}.sarif"))
+            sarif.outputLocation.set(parentProject.layout.buildDirectory.file("reports/detekt/${project.name}.sarif"))
             html.required.set(true)
         }
     }
@@ -237,7 +237,7 @@ fun MavenPublication.setMetadata() {
 tasks.withType<org.jetbrains.kotlin.gradle.targets.js.npm.tasks.KotlinNpmInstallTask> {
     dependsOn(":ts-model-api:npm_run_build")
 }
-val docsDir = buildDir.resolve("dokka")
+val docsDir = project.layout.buildDirectory.dir("dokka").get().asFile
 
 tasks.dokkaHtmlMultiModule {
     outputDirectory.set(docsDir.resolve("$version"))
--- a/bulk-model-sync-gradle-test/build.gradle.kts
+++ b/bulk-model-sync-gradle-test/build.gradle.kts
@@ -42,8 +42,8 @@ repositories {
 }
 
 val mps by configurations.creating
-val mpsDir = buildDir.resolve("mps").apply { mkdirs() }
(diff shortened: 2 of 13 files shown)


gradlew
```diff
@@ -197,11 +198,15 @@ if "$cygwin" || "$msys" ; then
...
 fi
 
-# Collect all arguments for the java command;
-#   * $DEFAULT_JVM_OPTS, $JAVA_OPTS, and $GRADLE_OPTS can contain fragments of
-#     shell script including quotes and variable substitutions, so put them in
-#     double quotes to make sure that they get re-expanded; and
-#   * put everything else in single quotes, so that it's not re-expanded.
+
+# Add default JVM options here. You can also use JAVA_OPTS and GRADLE_OPTS to pass JVM options to this script.
+DEFAULT_JVM_OPTS='"-Xmx64m" "-Xms64m"'
```
odzhychko, line 203: The maximum heap size of 64 Megabyte (`"-Xms64m"`) seems quite low.
As it was autogenerated, and it works, I did not look further into it.

languitar:
The bump to 8.4 is included here, because it was previously forgotten?

odzhychko:
> The bump to 8.4 is included here, because it was previously forgotten?

Yes. We had different Gradle versions in the different subproject. I updated them consistently to 8.4 once.

I took a brief look into dependabot. Dependabot seems not be able to update a Gradle version.
See. https://github.com/dependabot/dependabot-core/issues/2223

I didn't take a look for now, if we can centrally define the Gradle version for all Gradle projects in the repo.

languitar:
> > The bump to 8.4 is included here, because it was previously forgotten?
> 
> Yes. We had different Gradle versions in the different subproject. I updated them consistently to 8.4 once.

Can you change the commit message to explain this. It was a bit confusing to find a version bump in a PR that talks about removing some deprecated usage inside the Gradle files. The commit could also be scoped properly to bulk-model-sync-gradle

odzhychko:
> > > The bump to 8.4 is included here, because it was previously forgotten?
> > 
> > 
> > Yes. We had different Gradle versions in the different subproject. I updated them consistently to 8.4 once.
> 
> Can you change the commit message to explain this. It was a bit confusing to find a version bump in a PR that talks about removing some deprecated usage inside the Gradle files. The commit could also be scoped properly to bulk-model-sync-gradle


Ok, I create a separate PR and Issue describing the Gradle version update.
See https://github.com/modelix/modelix.core/pull/294

The scope is not `bulk-model-sync-gradle` because the Gradle version is updated in multiple projects in the repo.

odzhychko:
I will rework this PR to distinguish between the changes in our build-scripts and in our plugins.
Also, this build is broken, and I need to take a look into it.

odzhychko:
PR can be reviewed again.

languitar (commented):
I'm not sure if `feat` is the right commit type. User's don't have an benefit from these changes and can't do anything novel afterwards.

languitar:
fix sounds better

languitar (approved):

PR #294, opened 2023-10-23
build: upgrade to Gradle 8.4

odzhychko:
We have multiple projects with their own gradle-wrapper.properties I updated them simultaneously once.

The best practice would be to have only one Gradle wrapper in one umbrella build. The follow-up issue for this is https://issues.modelix.org/issue/MODELIX-576


languitar:
A hacky quickfix could be symlinking the properties files to avoid discrepancies

languitar (approved):

odzhychko:
> A hacky quickfix could be symlinking the properties files to avoid discrepancies

Thanks, I haven't thought of that.
I applied your suggestion.

PR #297, opened 2023-10-26
feat(model-client): Add the possibility to set the userId in model client V2

odzhychko:


model-client/src/commonMain/kotlin/org/modelix/model/client2/ModelClientV2.kt
```diff
@@ -80,6 +80,9 @@ class ModelClientV2(
     }
 
     suspend fun updateUserId() {
+        if (userId !== null) {
```
slisson, line 83: This changes the behavior of `updateUserId()` if no `userId` was provided in the constructor. The constructor parameter should be a separate variable so that it's possible to distinguish whether the initial value was provided in the constructor or by the server.

odzhychko: You are right. I missed, that `updateUserId` is a public function.

odzhychko: Done. It can be reviewed again.

I also added the possibility to change the user ID afterward.

model-client/src/commonMain/kotlin/org/modelix/model/client2/ModelClientV2.kt
```diff
@@ -80,19 +81,30 @@ class ModelClientV2(
...
+    fun setClientProvideUserId(userId: String?) {
+        clientProvidedUserId = userId
+    }
+
     override fun getClientId(): Int = clientId
 
     override fun getIdGenerator(): IIdGenerator = idGenerator
 
-    override fun getUserId(): String? = userId
+    override fun getUserId(): String? = if (clientProvidedUserId != null) clientProvidedUserId else serverProvidedUserId
```
slisson, line 107:
```suggestion
    override fun getUserId(): String? = clientProvidedUserId ?: serverProvidedUserId
```

odzhychko: Done.

slisson (approved):

PR #313, opened 2023-11-06
build: remove unneeded suppress warnings

odzhychko:
The suppression was relevant for false warnings, which are now fixed.

Cf. https://issues.modelix.org/issue/MODELIX-544

### DIFF
--- a/bulk-model-sync-lib/build.gradle.kts
+++ b/bulk-model-sync-lib/build.gradle.kts
@@ -16,7 +16,6 @@ kotlin {
             useJUnitPlatform()
         }
     }
-    @Suppress("UNUSED_VARIABLE", "KotlinRedundantDiagnosticSuppress")
     sourceSets {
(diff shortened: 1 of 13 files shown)


languitar:
Good catch. I probably would have forgotten that this is now possible.

Btw, conventional commits proposes to use the [git trailer](https://git-scm.com/docs/git-interpret-trailers) format for additional information in commits. We could think about standardizing the way we reference tickets in git commit message. I have used something like this before:

```
feat: subject line

some text
here

Fixes: MODELIX-42
```

Or `Related to: MODELIX-42` if not fixing.

languitar (approved):

odzhychko:
> Btw, conventional commits proposes to use the [git trailer](https://git-scm.com/docs/git-interpret-trailers) format for additional information in commits.

Seems nice. I will try it in my next commits. I added it to our weekly, to tell the others.

PR #315, opened 2023-11-07
docs: and link to ModelQL in navigation

odzhychko:


languitar (approved):
I think there's a typo in the commit message ("and" vs. "add")

PR #316, opened 2023-11-07
MODELIX-588 includedModules are not taken into account for the direction model-server -> MPS

mhuster23:


bulk-model-sync-lib/src/commonMain/kotlin/org/modelix/model/sync/bulk/Util.kt
```diff
@@ -21,6 +21,21 @@ import org.modelix.model.api.BuiltinLanguages
 import org.modelix.model.data.ModelData
 import org.modelix.model.data.NodeData
 
+/**
+ * Helper-method used by the gradle plugin and the MPS solution.
+ * Checks if a module is included in the sync.
```
languitar, line 26:
```suggestion
 * Checks if a module is included in the sync.
```

The first line of a docstring should usually state what the method does. The indication, where it is used, will easily become outdated when new usages appear or disappear. And the IDE probably does a better job finding them.

bulk-model-sync-gradle/src/main/kotlin/org/modelix/model/sync/bulk/gradle/tasks/ExportFromModelServer.kt
```diff
@@ -76,13 +85,26 @@ abstract class ExportFromModelServer @Inject constructor(of: ObjectFactory) : De
...
-                val nameRole = BuiltinLanguages.jetbrains_mps_lang_core.INamedConcept.name
-                val fileName = it.getPropertyValue(nameRole)
+
+            getIncludedModules(root).forEach {
+                val fileName = it.getPropertyValue(BuiltinLanguages.jetbrains_mps_lang_core.INamedConcept.name)
                 val outputFile = outputDir.resolve("$fileName.json")
                 ModelExporter(it).export(outputFile)
             }
         }
+        client.close()
```
languitar, line 95: As long as this is not closeable, shouldn't this happen in a try / finally block to be sure that closing always happens?

languitar (changes requested):

languitar (approved):

PR #319, opened 2023-11-10
MODELIX-501 Replace ts-model-api (2)

slisson:


model-api-gen-gradle-test/kotlin-generation/build.gradle.kts
```diff
@@ -1,44 +1,74 @@
...
+    }
 }
 
-tasks.compileKotlin {
-    dependsOn(":metamodel-export:generateMetaModelSources")
+npmPublish {
+    packages {
+        named("js") {
+            packageJson {
+                name.set("@modelix/model-client")
```
odzhychko, line 69: Why does `model-api-gen-gradle-test/kotlin-generation` create a `@modelix/model-client` package?

It seems to be used in `model-api-gen-gradle-test/typescript-generation`, but shouldn't `model-api-gen-gradle-test/typescript-generation` use the model client packaged in `model-client`?

model-api-gen-gradle-test/typescript-generation/build.gradle.kts
```diff
@@ -27,3 +28,17 @@ tasks.register<NpmTask>("packJsPackage") {
     dependsOn("npm_run_build")
     args.set(listOf("pack", "--pack-destination", "build"))
 }
+
+val updateModelClient = tasks.create<NpmTask>("updateModelClient") {
```
odzhychko, line 32: Why is the `model-client` built by `model-api-gen-gradle-test/kotlin-generation`?
Related to https://github.com/modelix/modelix.core/pull/319/files#r1400396250

model-api-gen-runtime/build.gradle.kts
```diff
@@ -59,3 +63,27 @@ kotlin {
 }
 
 description = "Runtime for the meta model generator"
+
+npmPublish {
```
odzhychko, line 67: Why do we need to publish `@modelix/model-api-gen-runtime`?
I did not find its usage.

vue-model-api/build.gradle.kts
```diff
@@ -75,8 +74,8 @@ val createPackageJsonForPublishing = tasks.create("createPackageJsonForPublishin
         // replaces versions which where needed for development and builds,
         // with its published versions
         @Suppress("UNCHECKED_CAST")
-        val dependencies = packageJsonData["dependencies"] as MutableMap<String, String>
-        dependencies["@modelix/model-client"] = "^${project.version}"
+        val peerDependencies = packageJsonData["peerDependencies"] as MutableMap<String, String>
+        // peerDependencies["@modelix/model-client"] = "^${project.version}"
```
odzhychko, line 78: Why don't we set the required version of`@modelix/model-client`?

vue-model-api/.eslintrc.js
```diff
@@ -28,7 +28,7 @@ module.exports = {
   plugins: ["@typescript-eslint"],
   rules: {
     "@typescript-eslint/no-unused-vars": [
-      "error",
+      "warn",
```
odzhychko, line 31: This should be reverted.
(n general `vue-model-api` has some dead code, that we should get rid of before merging.)

model-api-gen/src/main/kotlin/org/modelix/metamodel/generator/TypescriptMMGenerator.kt
```diff
@@ -56,274 +55,40 @@ class TypescriptMMGenerator(val outputDir: Path, val nameConfig: NameConfig = Na
...
-            }
+                ${
+            listOf(
+                language.name.substringBefore("."),
+                "org",
+                "INodeJS",
+                "IConceptJS",
+            ).distinct().sorted().joinToString(",\n")
+        }
+            } from "$runtimeNpmPackage";
```
odzhychko, line 75: That is weird, that one package (`$runtimeNpmPackage`/`@modelix/model-client`) must provide the access to the model client (server connnection, untyped intefaces) and the generated languages.

I thought of `@modelix/model-client` as a library that modelix would provide.
And the generated languages would come from another package and depend upon `@modelix/model-client`.

PR #320, opened 2023-11-13
feat(bulk-model-sync-gradle): add continueOnError option

languitar:
With the new option, the sync can ignore errors encountered during syncing instead of immediately terminating. This is some sort of graceful degradation at the cost of potentially resulting in an inconsistent sync.

I'll add the missing implementation for the direction towards MPS after #312 is merged to avoid unnecessary conflicts.

### DIFF
--- a/bulk-model-sync-gradle/src/main/kotlin/org/modelix/model/sync/bulk/gradle/config/ModelSyncGradleSettings.kt
+++ b/bulk-model-sync-gradle/src/main/kotlin/org/modelix/model/sync/bulk/gradle/config/ModelSyncGradleSettings.kt
@@ -43,6 +43,7 @@ data class SyncDirection(
     internal val registeredLanguages: Set<ILanguage> = mutableSetOf(),
     internal val includedModulePrefixes: Set<String> = mutableSetOf(),
     internal var mpsDebugEnabled: Boolean = false,
+    internal var continueOnError: Boolean = false,
 ) {
     fun fromModelServer(action: Action<ServerSource>) {
         val endpoint = ServerSource()
@@ -78,6 +79,10 @@ data class SyncDirection(
 
     fun registerLanguage(language: ILanguage) {
         (registeredLanguages as MutableSet).add(language)
+    }
+
+    fun enableContinueOnError(state: Boolean) {
+        continueOnError = state
     }
 }
 
--- a/bulk-model-sync-gradle/src/main/kotlin/org/modelix/model/sync/bulk/gradle/tasks/ImportIntoModelServer.kt
+++ b/bulk-model-sync-gradle/src/main/kotlin/org/modelix/model/sync/bulk/gradle/tasks/ImportIntoModelServer.kt
@@ -62,6 +62,9 @@ abstract class ImportIntoModelServer @Inject constructor(of: ObjectFactory) : De
     @Input
     val includedModulePrefixes: SetProperty<String> = of.setProperty(String::class.java)
 
+    @Input
+    val continueOnError: Property<Boolean> = of.property(Boolean::class.java)
+
     @TaskAction
     fun import() {
         registeredLanguages.get().forEach {
@@ -83,7 +86,7 @@ abstract class ImportIntoModelServer @Inject constructor(of: ObjectFactory) : De
             client.runWrite(branchRef) { rootNode ->
                 logger.info("Got root node: {}", rootNode)
                 logger.info("Importing...")
-                ModelImporter(rootNode).importFilesAsRootChildren(files)
+                ModelImporter(rootNode, continueOnError.get()).importFilesAsRootChildren(files)
                 logger.info("Import finished")
             }
         }
--- a/bulk-model-sync-lib/src/commonMain/kotlin/org/modelix/model/sync/bulk/ModelImporter.kt
+++ b/bulk-model-sync-lib/src/commonMain/kotlin/org/modelix/model/sync/bulk/ModelImporter.kt
@@ -38,7 +38,7 @@ import kotlin.jvm.JvmName
  *
  * @param root the root node to be updated
  */
-class ModelImporter(private val root: INode) {
+class ModelImporter(private val root: INode, private val continueOnError: Boolean) {
 
     private val originalIdToExisting: MutableMap<String, INode> = mutableMapOf()
     private val postponedReferences = ArrayList<() -> Unit>()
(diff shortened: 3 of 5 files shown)


bulk-model-sync-gradle/src/main/kotlin/org/modelix/model/sync/bulk/gradle/config/ModelSyncGradleSettings.kt
```diff
@@ -43,6 +43,7 @@ data class SyncDirection(
     internal val registeredLanguages: Set<ILanguage> = mutableSetOf(),
     internal val includedModulePrefixes: Set<String> = mutableSetOf(),
     internal var mpsDebugEnabled: Boolean = false,
+    var continueOnError: Boolean = false,
```
mhuster23, line 46: I wonder, if we should expose this only as a method like `enableContinueOnError()` for consistency as sync direction only used methods for its settings up until now.

docs/global/modules/core/pages/reference/component-bulk-model-sync-gradle.adoc
```diff
@@ -76,6 +76,11 @@ This means that only a minimal amount of write operations is used to update the
 |Includes all modules, whose fully qualified name starts with the given prefix, in the synchronisation process.
 |===
 
+|`continueOnError`
+|Boolean
+|If the sync encounters an error, simply log it and continue instead of terminating. This most likely results in an incomplete sync result. Default to `false`.
+|===
```
mhuster23, line 82: The table is broken because line 77 hasn't been removed.
Also the content does not fit the table header since it is not a method.

bulk-model-sync-lib/src/commonMain/kotlin/org/modelix/model/sync/bulk/ModelImporter.kt
```diff
@@ -47,6 +47,17 @@ class ModelImporter(private val root: INode) {
     private var currentNodeProgress = 0
     private val logger = KotlinLogging.logger {}
 
+    private fun doAndPotentiallyContinueErrors(block: () -> Unit) {
```
mhuster23, line 50:
```suggestion
    private fun doAndPotentiallyContinueOnErrors(block: () -> Unit) {
```
This should make it clearer. Still the method name is a bit long but I have a hard time coming up with a more concise name.

bulk-model-sync-gradle/src/main/kotlin/org/modelix/model/sync/bulk/gradle/tasks/ImportIntoModelServer.kt
```diff
@@ -83,7 +86,7 @@ abstract class ImportIntoModelServer @Inject constructor(of: ObjectFactory) : De
             client.runWrite(branchRef) { rootNode ->
                 logger.info("Got root node: {}", rootNode)
                 logger.info("Importing...")
-                ModelImporter(rootNode).importFilesAsRootChildren(files)
+                ModelImporter(rootNode, continueOnError.get()).importFilesAsRootChildren(files)
```
mhuster23, line 89: `continueOnError` only affects the import into the model server this way. For the import into MPS the solution must be changed as well.

mhuster23 (changes requested):

bulk-model-sync-lib/src/commonMain/kotlin/org/modelix/model/sync/bulk/ModelImporter.kt
languitar: The missing "on" was an oversight.

bulk-model-sync-gradle/src/main/kotlin/org/modelix/model/sync/bulk/gradle/config/ModelSyncGradleSettings.kt
languitar: I had that in my Gradle script and it looked a bit weird to me for just a single boolean.

bulk-model-sync-gradle/src/main/kotlin/org/modelix/model/sync/bulk/gradle/tasks/ImportIntoModelServer.kt
mhuster23: If you want it only for this direction, I suggest moving the `continueOnError` property to the `ServerTarget` so it's clear for the user.
For the other direction you could also add `continueOnError` for `LocalSource` and then pass the value via a system property in the ant script similar to includedModules. But maybe you should wait for #312. Since those changes will make it easier.

docs/global/modules/core/pages/reference/component-bulk-model-sync-gradle.adoc
languitar: fixed

bulk-model-sync-gradle/src/main/kotlin/org/modelix/model/sync/bulk/gradle/tasks/ImportIntoModelServer.kt
languitar: I can rebase my branch on that one and we wait for it to be merged.

languitar: I think I am stuck here. I'd want to implement that also for the direction into MPS. However, due to [#312 being on an outdated state of the sync](https://github.com/modelix/modelix.core/pull/312/files#r1390940876), I can easily built that feature in the current PR. I could log a warning at the moment and implement that once #312 is done.

bulk-model-sync-gradle/src/main/kotlin/org/modelix/model/sync/bulk/gradle/config/ModelSyncGradleSettings.kt
languitar: Updated to that pattern

bulk-model-sync-gradle/src/main/kotlin/org/modelix/model/sync/bulk/gradle/tasks/ImportIntoModelServer.kt
languitar: Is that ok with the message for the moment to unblock this?

bulk-model-sync-gradle/src/main/kotlin/org/modelix/model/sync/bulk/gradle/config/ModelSyncGradleSettings.kt
mhuster23: The name in the docs isn't updated. We could also get rid of the parameter as `enableContinueOnError(false)` wouldn't do anything.

languitar: without the flag, that would be a one-way street. I added the flag so that one is also able to disable this, too.

I'll update the docs

mhuster23 (approved):

PR #323, opened 2023-11-14
fix(mps-model-adapters): handle unresolvable model imports

mhuster23:
If an imported model could not be resolved, it caused a `NullPointerException`. 
This change handles this case gracefully.


languitar (commented):
Would be great to have a test case for these issues. We are starting to hit a lot of corner cases and the implemented fixes are easily lost with the next refactoring if no test ensures this behavior.

mhuster23:
> Would be great to have a test case for these issues. We are starting to hit a lot of corner cases and the implemented fixes are easily lost with the next refactoring if no test ensures this behavior.

I think this is a bigger topic, that we need to discuss in the near future.
Tests would be really nice, but it's not trivial since this is code, that interacts with the MPS API.

languitar:
> > Would be great to have a test case for these issues. We are starting to hit a lot of corner cases and the implemented fixes are easily lost with the next refactoring if no test ensures this behavior.
> 
> I think this is a bigger topic, that we need to discuss in the near future. Tests would be really nice, but it's not trivial since this is code, that interacts with the MPS API.

Do we have a follow up ticket for that on the tracker?

languitar (approved):
Would be nice to at least reword the commit message to include some explanations for this as long as there's no test that visualizes the situation that is fixed easily.

PR #327, opened 2023-11-20
MODELIX-10 Server side rendering

slisson:
Depends on https://github.com/modelix/modelix.core/pull/319


slisson:
Included in https://github.com/modelix/modelix.core/pull/336

PR #331, opened 2023-11-22
build: add workaround for failing download of yarn in CI

odzhychko:


odzhychko:
The underlying issue with GitHub seems to be resolved.

languitar (approved):

PR #332, opened 2023-11-23
MODELIX-608 Health check for warming up the MPS cache

slisson:
New health check `loadModels` implemented. Multiple `moduleNamespace` parameters can be provided to specify the relevant modules.

### DIFF
--- a/docs/global/modules/core/pages/reference/component-mps-model-server-plugin.adoc
+++ b/docs/global/modules/core/pages/reference/component-mps-model-server-plugin.adoc
@@ -9,9 +9,21 @@ https://api.modelix.org/3.12.0/mps-model-server-plugin/index.html[API doc^] | ht
 
 == Health checks
 
-The plugin offers a set of health checks via HTTP on port 48305 and path `/health`.
-Health checks can be enabled adding query parameters with the health check name and the value `true` to the request.
+The plugin offers a set of health checks via HTTP GET on port `48305` and path `/health`.
 
+Health checks can be enabled adding query parameters with the health check name and the value `true` to the request.
+Some health checks require further information that needs to be provided by query parameters.
+
+.Example of combining health check
+[source,text]
+----
+http(s)://<host>:48305/health?indexer=true&loadModels=true&loadModelsModuleNamespacePrefix=foo.bar <1> <2>
+----
+<.> `indexer=true` enables <<indexer>>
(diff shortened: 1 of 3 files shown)


odzhychko (dismissed):

docs/global/modules/core/pages/reference/component-mps-model-server-plugin.adoc
```diff
@@ -30,3 +30,9 @@ Reports an unhealthy system whenever no project is loaded.
 
 Reports an unhealthy system when no virtual folders are available.
 This might also be true in case a project without virtual folders is fully loaded.
+
+=== loadModels
+
+Can be used to eagerly load model data avoiding a slow first query.
+Multiple `moduleNamespace` parameters can be provided to specify the relevant modules.
+All modules starting with the string provided by that parameter are included.
```
languitar, line 38: So an appropriate request would be?

```
/health?loadModels=true&moduleNamespace=foo.bar
```

A bit weird that parameters are now used for two purposes. One time for selecting the check, and then here as a special case for the modules.

In any case, a valid example call given this complexity would be good to have in the docs.

slisson: I solved the part that requires MPS knowledge. Feel free to continue working on this PR.

odzhychko: Ok. Thank you for solving the MPS parts. I can continue with the details.

docs/global/modules/core/pages/reference/component-mps-model-server-plugin.adoc
```diff
@@ -30,3 +42,25 @@ Reports an unhealthy system whenever no project is loaded.
 
 Reports an unhealthy system when no virtual folders are available.
 This might also be true in case a project without virtual folders is fully loaded.
+
+[#loadModels]
+=== loadModels
+
+It can be used to eagerly load model data, avoiding a slow first query.
```
languitar, line 49: Maybe one additional sentence helps here:

```suggestion
Returns after trying to eagerly cache a set of specified modules.
This check can be used to avoiding a slow first ModelQl query after launching an MPS instance running this plugin.
```

languitar (approved):

PR #335, opened 2023-11-24
docs: add image depicting the API generation pipeline

nkoester:
Just adds an overview image to get the idea of the pipeline across


odzhychko (approved):

docs/global/modules/core/images/api-gen-pipeline.png
languitar, line 1: I'd propose to add captions / labels to each item in the graph. These icons look fancy, but don't convey much meaning if one doesn't know them.

PR #345, opened 2023-11-28
MODELIX-640 Handle large modules in bulk sync

mhuster23:
The bulk sync failed for large modules as the json exceeded the string size limit.
Now we use streams instead, to handle this case.

Apart from that this PR includes several significant performance improvements for the bulk-sync.
Additionally, the legacy MPS solution `bulk-model-sync-solution` is removed.

Also fixes [MODELIX-615](https://issues.modelix.org/issue/MODELIX-615/Enable-bulk-model-sync-gradle-to-push-to-non-existing-branches)


bulk-model-sync-mps/src/main/kotlin/org/modelix/mps/model/sync/bulk/MPSBulkSynchronizer.kt
```diff
@@ -0,0 +1,118 @@
...
+        repository.modelAccess.runReadAction {
+            val allModules = repository.modules
+            val includedModules = allModules.filter {
+                isModuleIncluded(it.moduleName!!, includedModuleNames, includedModulePrefixes)
+            }
+            val numIncludedModules = includedModules.count()
+            val outputPath = System.getProperty("modelix.mps.model.sync.bulk.output.path")
+            val counter = AtomicInteger()
+
+            includedModules.toList().parallelStream().forEach { module ->
```
languitar, line 49: Is that toList necessary here?

mhuster23: No, removed it. Good catch.

mhuster23:
@slisson I had to force-push this, since the linter didn't like your `performance` commit prefix.
Your commits look good to me, I only changed the doc a bit.
Seems like the serialization changes broke sth though, could you take a look at that?

bulk-model-sync-gradle/src/main/kotlin/org/modelix/model/sync/bulk/gradle/tasks/ImportIntoModelServer.kt
```diff
@@ -84,11 +87,20 @@ abstract class ImportIntoModelServer @Inject constructor(of: ObjectFactory) : De
...
             }
         }
     }
 }
+
+/**
+ * Memory optimization that doesn't record individual change operations, but only the result.
+ */
+private fun INode.runBulkUpdate(body: () -> Unit) {
+    ((this as PNodeAdapter).branch as OTBranch).runBulkUpdate(body = body)
```
languitar, line 105: Is there a chance that this cast can fail?

bulk-model-sync-lib/src/commonMain/kotlin/org/modelix/model/sync/bulk/ModelImporter.kt
```diff
@@ -51,7 +51,7 @@ class ModelImporter(private val root: INode, private val continueOnError: Boolea
     // updated to the constructor with two arguments.
     constructor(root: INode) : this(root, false)
 
-    private fun doAndPotentiallyContinueOnErrors(block: () -> Unit) {
+    private inline fun doAndPotentiallyContinueOnErrors(block: () -> Unit) {
```
languitar, line 54: Maybe add a comment for why this was inlined?

bulk-model-sync-lib/src/commonMain/kotlin/org/modelix/model/sync/bulk/ModelImporter.kt
```diff
@@ -122,6 +124,17 @@ class ModelImporter(private val root: INode, private val continueOnError: Boolea
...
 
+            // optimization that uses the bulk operation .addNewChildren
+            if (existingNodes.isEmpty() && expectedNodes.all { originalIdToExisting[it.originalId()] == null }) {
+                node.addNewChildren(role, -1, expectedNodes.map { it.concept?.let { ConceptReference(it) } }).zip(expectedNodes).forEach { (newChild, expected) ->
+                    val expectedId = checkNotNull(expected.originalId()) { "Specified node '$expected' has no id" }
+                    newChild.setPropertyValue(NodeData.idPropertyKey, expectedId)
+                    originalIdToExisting[expectedId] = newChild
+                    syncNode(newChild, expected)
+                }
+                continue
```
languitar, line 135: Can you extend the comment on what this optimizes and in which specific case. It takes some time to understand the condition that is checked here and also how it is optimized if true. The next one stumbling upon this code would probably be happy to get some reading aids.

docs/global/modules/core/pages/reference/component-bulk-model-sync-gradle.adoc
```diff
@@ -132,10 +132,12 @@ mpsBuild {
 |`repositoryId`
 |String
 |Id of the target/source model-server repository.
+If the repository does not exist on the model-server, it will be created including a master branch.
 
 |`branchName`
 |String
 |Name of the target/source model-server branch.
+If the target branch does not exist on the model-server, it will be created.
```
languitar, line 140: Can these changes be formed into individual feat commits? That way, they become visible in the changelog. This is important information for us as users as we can drop some workarounds now.

model-datastructure/src/commonMain/kotlin/org/modelix/model/operations/BulkUpdateOp.kt
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    private fun getResultTree(store: IDeserializingKeyValueStore): CLTree = CLTree(resultTreeHash.getValue(store), store)
+
+    override fun toString(): String {
+        return "BulkUpdateOp ${resultTreeHash.getHash()}, ${SerializationUtil.longToHex(subtreeRootId)}"
+    }
+
+    override fun toCode(): String {
+        return """// TODO BulkUpdateOp"""
```
languitar, line 54: Follow-up ticket to reference here?

bulk-model-sync-gradle/src/main/kotlin/org/modelix/model/sync/bulk/gradle/tasks/ImportIntoModelServer.kt
slisson: It's a private method that is only used once on the INode created by `IModelClientV2.runWrite`. If it succeeds once it will always succeed. Unit tests should detect any incompatible changes in the implementation of `IModelClientV2.runWrite`.

bulk-model-sync-lib/src/commonMain/kotlin/org/modelix/model/sync/bulk/ModelImporter.kt
mhuster23: I'm not sure if this needs further explanation, as it is very similar to the [prime example for inline functions](https://kotlinlang.org/docs/inline-functions.html).

model-datastructure/src/commonMain/kotlin/org/modelix/model/operations/BulkUpdateOp.kt
slisson: I just removed `IOperation.toCode()`. It's dead code that was used previously to generate the test cases in [ConflictResolutionTest](https://github.com/modelix/modelix.core/blob/main/model-datastructure/src/commonTest/kotlin/ConflictResolutionTest.kt).

bulk-model-sync-lib/src/commonMain/kotlin/org/modelix/model/sync/bulk/ModelImporter.kt
languitar: We probably have a good reason why we go the special case here for inlining it. It would be bad, if that reasoning gets lost in time and the next one touching the code doesn't know about it anymore.

slisson: I agree that documenting the reasons is unnecessary, because it would just repeat the Kotlin documentation for when to use this keyword. You can expect an experienced Kotlin developer to know this.

nkoester: I understand that documentation is tedious - especially if it is self written code and "obvious". but please consider that even our current teams is not composed of only kotlin experts.

slisson: Maybe I phrased it wrong. It doesn't require any experience to look this keyword up in the Kotlin docs, in case someone is unfamiliar with it. It's just a normal language feature used in it's intended way. It's not about being tedious work to document it, it just noise. Bad document is worse than no documentation.

languitar: We can move a bit more of rational to the commit message maybe to avoid the comment. There's a history of a report that this might have slowed down execution. This information would be lost in time if it's not written down somewhere permanently and well-discoverable from the code base.

mhuster23: The claims of that report has never been verified and I don't want to add possibly false information. 
The rationale was to improve performance no matter if the difference is noticeable to a user. The commit prefix perf conveys sufficient information. If there is an unexperienced dev, who is curious why this improves performance, they can read the documentation for the keyword. We shouldn't overthink this.

docs/global/modules/core/pages/reference/component-bulk-model-sync-gradle.adoc
mhuster23: It's hard to justify attributing these improvements as features to the gradle plugin, since they are a result of one bugfix in the model-client and no code in the plugin was changed for this. This would be one of those cases, where it would be nice to have the possibility to create some (additional) manual release notes.

slisson (approved):

docs/global/modules/core/pages/reference/component-bulk-model-sync-gradle.adoc
languitar: Ah I see. I thought this was a change in this PR.

PR #346, opened 2023-11-28
MODELIX-536 Add tests to extracted sync lib and MPS plugin

slisson:


slisson:
Replaced by https://github.com/modelix/modelix.core/pull/365

PR #353, opened 2023-12-01
MODELIX-646 Use resource based routing in model-server

nkoester:


api/public.yaml
languitar, line 1: The filename is still a bit too generic for the large project. Maybe something like `model-server-public.yaml`?

model-server/build.gradle.kts
```diff
@@ -184,3 +190,81 @@ spotless {
...
+}
+tasks.named("processResources") {
+    dependsOn("openApiGenerate")
+}
+tasks.named("compileKotlin") {
+    dependsOn("openApiGenerate")
+}
+tasks.named("runKtlintCheckOverMainSourceSet") {
+    dependsOn("openApiGenerate")
+}
```
languitar, line 250: Isn't it enough to let processResources depend on the generation? I though build and compile depend on processResources again?

model-server/src/main/kotlin/org/modelix/model/server/Main.kt
```diff
@@ -159,6 +161,8 @@ object Main {
                 install(Routing)
                 installAuthentication(unitTestMode = !KeycloakUtils.isEnabled())
                 install(ForwardedHeaders)
+                install(Resources)
+                install(IgnoreTrailingSlash)
```
languitar, line 165: Maybe link this as the reason for installing the plugin: https://opensource.zalando.com/restful-api-guidelines/#136

model-server/src/main/kotlin/org/modelix/model/server/handlers/ModelReplicationServer.kt
```diff
@@ -68,171 +70,222 @@ class ModelReplicationServer(val repositoriesManager: RepositoriesManager) {
...
+            call.respondDelta(versionHash, baseVersionHash)
+        }
+        get<Paths.v2RepositoriesRepositoryVersionsVersionHashHistoryOldestVersionHashGet> {
+            fun ApplicationCall.repositoryId() = RepositoryId(parameters["repository"]!!)
+            fun PipelineContext<Unit, ApplicationCall>.repositoryId() = call.repositoryId()
+
+            TODO()
+        }
+
+        post<Paths.v2RepositoriesRepositoryBranchesBranchQueryPost> {
```
languitar, line 236: Can these ugly names be made more pretty by setting an `operationId` (cf. https://swagger.io/docs/specification/paths-and-operations/) in the spec?

model-server/src/test/kotlin/org/modelix/model/server/ModelClientV2Test.kt
```diff
@@ -46,6 +48,8 @@ class ModelClientV2Test {
                 json()
             }
             install(WebSockets)
+            install(Resources)
+            install(IgnoreTrailingSlash)
```
languitar, line 52: This is pretty much repeated in many tests. Maybe a common abstraction would be good then.

model-server/build.gradle.kts
nkoester: for me gradle complains when i remove the dependencies for `compileKotlin` and `runKtlintCheckOverMainSourceSet`. the `build` dependency can be removed

api/public.yaml
```diff
@@ -0,0 +1,1365 @@
...
+          content:
+            text/plain:
+              schema:
+                type: string
+        "500":
+          description: "Internal Server Error"
+          content:
+            text/plain:
+              schema:
+                type: string
```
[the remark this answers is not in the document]
nkoester: done for the obvious ones

model-server/src/test/kotlin/org/modelix/model/server/ModelClientV2Test.kt
nkoester: True, but I consider that part of a larger effort after this PR to restructure and extend tests. I will leave it as is for now.

model-server/src/main/kotlin/org/modelix/model/server/Main.kt
nkoester: done

api/public.yaml
nkoester: Will do `model-server.yaml` for now. We need to decide / discuss and further separate the public and internal parts in the future.

model-server/src/main/kotlin/org/modelix/model/server/handlers/ModelReplicationServer.kt
nkoester: Generally, yes. But practically I do not even know the purpose of all available endpoints. This will be a dedicated next step. I will add those I know and can semantically understand. The others will need refinement later on (just like the entire API...)

nkoester:
There are many open tasks, but i consider it a significant change to merge and build upon.

model-server/src/main/kotlin/org/modelix/model/server/handlers/ContentExplorer.kt
```diff
@@ -86,7 +87,7 @@ class ContentExplorer(private val client: IModelClient, private val repoManager:
                     bodyContent { contentPageBody(rootNode, versionHash, emptySet()) }
                 }
             }
-            post("/content/{versionHash}/") {
+            post<Paths.contentVersionHashGet> {
```
odzhychko, line 90:
```suggestion
            post<Paths.contentVersionHashPost> {
```

model-server/build.gradle.kts
```diff
@@ -184,3 +187,77 @@ spotless {
...
 }
+
+// OpenAPI integration
+val basePackage = project.group.toString()
+val openAPIgenerationPath = "$buildDir/generated/openapi"
+
+// Pairs of the different OpenAPI files we use. Each pair must have its own 'category' as first argument as these
+// are used to generate corresponding packages
+val openApiFiles = listOf(
+    Pair("public", "model-server"),
```
odzhychko, line 198: just FYI: You could have used `"public" to "model-server"` to create a `Pair`.

model-server/src/main/kotlin/org/modelix/model/server/handlers/ContentExplorer.kt
nkoester: good catch!

model-server/build.gradle.kts
nkoester: ✅

model-server/build.gradle.kts
```diff
@@ -184,3 +187,77 @@ spotless {
                 '\n'*/
     }
 }
+
+// OpenAPI integration
+val basePackage = project.group.toString()
+val openAPIgenerationPath = "$buildDir/generated/openapi"
```
odzhychko, line 193: nit: `buildDir` is deprecated. Use `project.layout.buildDirectory`

```
w: file:///Users/odzhychko/Documents/arbeit1/modelix/modelix.core/model-server/build.gradle.kts:193:31: 'getter for buildDir: File!' is deprecated. Deprecated in Java
```

odzhychko (changes requested):
I have some minor comments. Everything else looks good to me.

model-server/build.gradle.kts
languitar: ok

model-server/src/main/kotlin/org/modelix/model/server/handlers/ModelReplicationServer.kt
languitar: Can't we just use this PR to sort out good names?

languitar (changes requested):

api/model-server-deprecated.yaml
```diff
@@ -0,0 +1,226 @@
...
+# Manual changes were done for this 'spec' to work in-place with the
+# model-server for now. A lot of changes were done and are still necessary
+# to make this OpenAPI a viable artifact. It will most likely be split
+# into multiple OpenAPI files.
+
+openapi: "3.0.3"
+info:
+  title: "model-server deprecated API"
+  description: "modelix deprecated API"
+  version: "0.0.1"
```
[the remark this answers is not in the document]
nkoester: unfortunately not. ideally we should use the automatic semantic versioning here. but I would realize this in the build step during API publication.

api/model-server-html.yaml
```diff
@@ -0,0 +1,257 @@
...
+# to make this OpenAPI a viable artifact. It will most likely be split
+# into multiple OpenAPI files.
+#
+# This API is currently used in the ContentExplorer
+
+openapi: "3.0.3"
+info:
+  title: "model-server content browser API"
+  description: "modelix content browser API"
+  version: "0.0.1"
```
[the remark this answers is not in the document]
nkoester: see above - the number here is generally meaningless and should rather correspond to the the semantic version.

api/model-server-deprecated.yaml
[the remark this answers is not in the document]
languitar: No, not at all. Just because we do a major increment in the modelix.core version doesn't mean we have a breaking change in the REST API. It would be a real problem with this project structure to go that way. We would constantly indicate to clients that potentially have changed in a breaking fashion.

api/model-server-deprecated.yaml
```diff
@@ -0,0 +1,226 @@
...
+servers:
+  - url: 'http://127.0.0.1:28101'
+    description: local model-server
```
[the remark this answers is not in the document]
nkoester: I did the thing and additionally now serve the swaggerUI at `/swagger`

api/model-server-deprecated.yaml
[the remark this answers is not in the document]
nkoester: makes sense. I put it to `1.0.0` for now

model-server/src/main/kotlin/org/modelix/model/server/Main.kt
```diff
@@ -219,11 +227,21 @@ object Main {
...
                             }
                         }
                         call.respondText("Model Server")
                     }
+                    if (cmdLineArgs.noSwaggerUi) {
+                        get("swagger") {
+                            call.respondText("SwaggerUI is disabled")
+                        }
+                    } else {
+                        swaggerUI(path = "swagger", swaggerFile = "../api/model-server.yaml")
```
languitar, line 243: How does this work with the other definitions?

nkoester: we only expose the public one

languitar:
```suggestion
                        // we only serve the public API to the outside via swagger UI
                        swaggerUI(path = "swagger", swaggerFile = "../api/model-server.yaml")
```

api/model-server.yaml
```diff
@@ -0,0 +1,607 @@
...
+# to make this OpenAPI a viable artifact. It will most likely be split
+# into multiple OpenAPI files.
+
+openapi: "3.0.3"
+info:
+  title: "model-server public API"
+  description: "modelix public API"
+  version: "1.0.0"
+servers:
+  - url: '/'
```
languitar, line 16:
```suggestion
  - url: '/v2'
```

api/model-server.yaml
```diff
@@ -0,0 +1,607 @@
...
+    description: model-server
+paths:
+  /v2/server-id:
```
languitar, line 19:
```suggestion
  /server-id:
```

and so on. That way, we can more easily change our mind about the location of the API

nkoester: those will be breaking changes - I did not intend to have a breaking change in this PR but rather just simply transition to resource based paths. I will add a follow up issue in which we can discuss these breaking changes in advance

languitar: they wouldn't be? Because the base path on top is changed, too?

nkoester: but that change would move all non `/v2` paths, such as `/health` to `/v2/health` and thus create a breaking change?

nkoester: We need to follow up on this. I created an issue here to separate this discussion from this PR.

[MODELIX-683](https://issues.modelix.org/issue/MODELIX-683) Decide and organize model-server endpoints (via OpenAPI definition)

api/model-server.yaml
nkoester: We need to follow up on this. I created an issue here to separate this discussion from this PR.

[MODELIX-683](https://issues.modelix.org/issue/MODELIX-683) Decide and organize model-server endpoints (via OpenAPI definition)

languitar (approved):

PR #354, opened 2023-12-04
fix(model-datastructure): fix LinearHistory

odzhychko:


model-datastructure/src/commonMain/kotlin/org/modelix/model/VersionMerger.kt
```diff
@@ -45,32 +45,36 @@ class VersionMerger(private val storeCache: IDeserializingKeyValueStore, private
...
+            // No merge needed, fast-forward to the right version
+            return rightVersion
+        }
+        val leftNonMerges = leftVersions.filterNot { it.isMerge() }
+        val rightNonMerges = rightVersions.filterNot { it.isMerge() }
+        // We do not have to compare sets.
+        // If the left and right versions have the same versions (excluding merges) in the same order,
+        // then their contents will be the same.
+        // If the left and right versions have different versions (excluding merges) or a different order,
+        // then their content might be different, and we need to decide on a new order in a new merge.
```
slisson, line 70: I think this `if` branch is important to avoid the described endless creation of merge commits. Even if the order of both sides is different I would still expect the new order in `versionsToApply` to be equal to one of the sides, in most cases. We should check that and then return the left/right side instead of creating a new merge commit.

odzhychko: Done.

model-datastructure/src/commonMain/kotlin/org/modelix/model/LinearHistory.kt
```diff
@@ -1,99 +1,58 @@
...
-                    )
+            while (stack.isNotEmpty()) {
+                val version = stack.last()
+                val versionWasVisited = !visited.add(version)
+                if (versionWasVisited) {
+                    stack.removeLast()
+                    yield(version)
+                }
+                val descendants = if (version.isMerge()) {
+                    // Put version 1 last, so that is processed first.
```
slisson, line 37: I already mentioned it before. This algorithm is not allowed to make any decision based on the order in which version appear in the merge. No distinction between MergedVersion1 and MergedVersion2 is allowed for the ordering, because that can differ on different clients. I added a test case that detects this issue and explains it with more details.

odzhychko: > This algorithm is not allowed to make any decision based on the order in which versions appear in the merge.

I thought this only referred to the order in the arguments of  `LinearHistory.load` this would have made the old algorithm produce the same order if the same versions are known.

But this did indeed not prevent the "merge algorithm changing its mind about the result" when a new version was added on one client.

I labeled the property that prevents that "monotonicity". (See comments LinearHistory.load)


> I added a test case that detects this issue and explains it with more details.

Thanks for the great test.

The implemented algorithm is basically the one you implemented (ordering by height and then version ID).
I just removed the recursion to prevent stack overflows, which might happen when use this function to view long histories.

---

For the `UndoTest` is still have an idea to try out.

odzhychko:
I think, I have got something. The idea is to order the merges by ID and distance from the base version, but keep all normal versions between merges linearly as they appeared. bf502cf is a WIP, I'm going to clean up and document the code.

slisson (approved):

PR #361, opened 2023-12-04
Modelix 645 test MPS compatibility

odzhychko:


languitar (approved):
Can you give a few more details in the commit messages on what is happening here? PR and commit messages are a bit empty.

odzhychko:
@languitar Do the updated commit messages explain enough?

PR #365, opened 2023-12-05
Legacy sync plugin

slisson:
Until https://github.com/modelix/modelix.core/pull/237 is done, here is the original implementation of the synchronization plugin between MPS and the model server. It was copied from https://github.com/modelix/modelix/tree/mps/2021.1/mps/org.modelix.model.mpsplugin . Also rescued most of the tests from https://github.com/modelix/modelix/tree/mps/2021.1/integrationtests .

This should make the maintenance for different MPS versions easier.

I was able to get rid of dependencies on [MPS-extensions](https://github.com/JetBrains/MPS-extensions), so the plugin can be installed independently.
Being able to install it as an IntelliJ plugin is another bonus that fixes https://issues.modelix.org/issue/MODELIX-193/Cloud-plugin-can-only-be-used-as-a-global-library .


slisson:
Where else should it go?

slisson:
It's not based on https://github.com/modelix/modelix.core/pull/237, if that was your assumption. I moved it into a new plugin with a new ID which is unrelated to the new sync plugin. It will not cause any merge conflicts with that one.

nkoester:
moved here https://github.com/modelix/modelix.mps-plugins

PR #372, opened 2023-12-13
MODELIX-533 Use repository content filtering

mhuster23:
[MODELIX-533](https://issues.modelix.org/issue/MODELIX-533)

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -168,10 +168,24 @@ subprojects {
 
 allprojects {
     repositories {
-        mavenLocal()
-        maven { url = uri("https://artifacts.itemis.cloud/repository/maven-mps/") }
-        maven { url = uri("https://maven.pkg.jetbrains.space/public/p/kotlinx-html/maven") }
-        mavenCentral()
+        val modelixRegex = "org\\.modelix.*"
+        mavenLocal {
+            content {
+                includeGroupByRegex(modelixRegex)
+            }
+        }
+        maven {
+            url = uri("https://artifacts.itemis.cloud/repository/maven-mps/")
+            content {
+                includeGroupByRegex(modelixRegex)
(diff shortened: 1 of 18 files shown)


settings.gradle.kts
```diff
@@ -1,9 +1,27 @@
 pluginManagement {
     repositories {
-        mavenLocal()
-        gradlePluginPortal()
-        maven { url = uri("https://artifacts.itemis.cloud/repository/maven-mps/") }
-        mavenCentral()
+        val modelixRegex = "org\\.modelix.*"
```
odzhychko, line 3: `excludeGroup` and `includeGroup` could be used instead of the variant with the Regex.

mhuster23: It doesn't work everywhere like that, since `includeGroup` would not include subgroups like `org.modelix.mps`. I used the variant with the regex for consistency.

odzhychko: Ah, ok. Then you could use `includeGroupAndSubgroups` (But Regex is also fine.)

mhuster23: There's also `includeGroupAndSubgroups` but that one is marked as incubating so I avoided it.

odzhychko (approved):

build.gradle.kts
```diff
@@ -168,10 +168,24 @@ subprojects {
 
 allprojects {
     repositories {
-        mavenLocal()
-        maven { url = uri("https://artifacts.itemis.cloud/repository/maven-mps/") }
-        maven { url = uri("https://maven.pkg.jetbrains.space/public/p/kotlinx-html/maven") }
-        mavenCentral()
+        val modelixRegex = "org\\.modelix.*"
```
languitar, line 171: Maybe adding a `$` in front would be even safer?

PR #379, opened 2023-12-20
MODELIX-644 Support writing in generated typed API with ModelQl

mhuster23:
Adds the ability to perform write operations in ModelQL queries.
As the MetaModelGenerator class has organically grown to a significant size, [MODELIX-667](https://issues.modelix.org/issue/MODELIX-667) was created as a followup ticket.

### DIFF
--- a/model-api-gen-gradle-test/kotlin-generation/src/test/kotlin/org/modelix/modelql/typed/TypedModelQLTest.kt
+++ b/model-api-gen-gradle-test/kotlin-generation/src/test/kotlin/org/modelix/modelql/typed/TypedModelQLTest.kt
@@ -17,6 +17,8 @@ import io.ktor.client.HttpClient
 import io.ktor.server.testing.testApplication
 import jetbrains.mps.baseLanguage.C_ClassConcept
 import jetbrains.mps.baseLanguage.C_IntegerType
+import jetbrains.mps.baseLanguage.C_MinusExpression
+import jetbrains.mps.baseLanguage.C_ParameterDeclaration
 import jetbrains.mps.baseLanguage.C_PlusExpression
 import jetbrains.mps.baseLanguage.C_PublicVisibility
 import jetbrains.mps.baseLanguage.C_ReturnStatement
@@ -24,8 +26,12 @@ import jetbrains.mps.baseLanguage.C_StaticMethodDeclaration
 import jetbrains.mps.baseLanguage.C_VariableReference
 import jetbrains.mps.baseLanguage.ClassConcept
 import jetbrains.mps.baseLanguage.StaticMethodDeclaration
+import jetbrains.mps.lang.editor.imageGen.C_ImageGenerator
+import jetbrains.mps.lang.editor.imageGen.ImageGenerator
 import org.modelix.apigen.test.ApigenTestLanguages
+import org.modelix.metamodel.instanceOf
 import org.modelix.metamodel.typed
+import org.modelix.metamodel.untyped
 import org.modelix.model.api.IBranch
 import org.modelix.model.api.PBranch
 import org.modelix.model.api.getRootNode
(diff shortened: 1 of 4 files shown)


modelql-typed/src/commonMain/kotlin/org/modelix/modelql/typed/TypedModelQL.kt
```diff
@@ -140,6 +144,17 @@ object TypedModelQL {
...
+    fun <ParentT : ITypedNode, ChildT : ITypedNode> addNewChild(input: IMonoStep<ParentT>, link: ITypedChildListLink<ChildT>, index: Int = -1, concept: ITypedConcept): IMonoStep<ChildT> {
+        val conceptRef = ConceptReference(concept.untyped().getUID())
+        return input.untyped().addNewChild(link.untyped(), index, conceptRef).typedUnsafe(link.getTypedChildConcept().getInstanceInterface())
+    }
+
+    fun <ParentT : ITypedNode, ChildT : ITypedNode> setChild(input: IMonoStep<ParentT>, link: ITypedSingleChildLink<ChildT>, concept: ITypedConcept): IMonoStep<ChildT> {
+        val conceptRef = ConceptReference(concept.untyped().getUID())
+        input.untyped().children(link.untyped().key()).firstOrNull().mapIfNotNull { it.remove() }
+        return input.untyped().addNewChild(link.untyped(), conceptRef).typedUnsafe(link.getTypedChildConcept().getInstanceInterface())
+    }
```
languitar, line 156: What about tests for these new methods?

model-api-gen/src/main/kotlin/org/modelix/metamodel/generator/MetaModelGenerator.kt
```diff
@@ -463,6 +464,42 @@ class MetaModelGenerator(
...
                                     .build(),
                             )
+                            val returnType = IMonoStep::class.asTypeName().parameterizedBy(targetType)
+                            val receiverType = IMonoStep::class.asTypeName().parameterizedBy(concept.nodeWrapperInterfaceType())
+                            if (feature.multiple) {
+                                addFunction(
+                                    FunSpec.builder(feature.addMethodName())
+                                        .returns(returnType)
+                                        .receiver(receiverType)
+                                        .addParameter("concept", ITypedConcept::class.asTypeName())
```
languitar, line 474: Can we use a default for this parameter when the child type is not abstract?

languitar (commented):
Some initial comments inline. What about a test for the generated methods?

modelql-typed/src/commonMain/kotlin/org/modelix/modelql/typed/TypedModelQL.kt
languitar: Ah, separate commit.

model-api-gen/src/main/kotlin/org/modelix/metamodel/generator/GeneratorInput.kt
```diff
@@ -302,7 +302,10 @@ internal sealed class ProcessedLink(name: String, uid: String?, optional: Boolea
 }
 
 internal class ProcessedChildLink(name: String, uid: String?, optional: Boolean, var multiple: Boolean, type: ProcessedConceptReference, deprecationMessage: String?) :
-    ProcessedLink(name, uid, optional, type, deprecationMessage)
+    ProcessedLink(name, uid, optional, type, deprecationMessage) {
+
+    fun addMethodName() = "add" + generatedName.take(1).uppercase() + generatedName.drop(1)
```
slisson, line 307: Maybe something like `adderMethodName` would be less confusing.

model-api-gen-gradle-test/kotlin-generation/src/test/kotlin/org/modelix/modelql/typed/TypedModelQLTest.kt
```diff
@@ -195,4 +236,167 @@ class TypedModelQLTest {
...
+        assertEquals(C_StaticMethodDeclaration.untyped().getUID(), children.last().untyped().concept?.getUID())
+    }
+
+    @Test
+    fun `add new child - default concept`() = runTest { httpClient ->
+        val client = ModelQLClient.builder().url("http://localhost/query").httpClient(httpClient).build()
+        client.query { root ->
+            root.descendants().ofConcept(C_XmlComment)
+                .first()
+                .addLines()
```
slisson, line 355: The generated method name is a bit confusing in cases like this. Maybe `addTo` would be a better prefix?

mhuster23:
> Do we have automatic disambiguation in case an adder method name would collide with a different accessor name from the model?

No, but I think we can assume that there is no collision. It would be a weird model, if a role `x` and `addToX` existed both on the same concept at the same time.

slisson:
Conflicts between role names are already resolved here: https://github.com/modelix/modelix.core/blob/9f267dfbcb2692cdd9967319c7aab84ea8399284/model-api-gen/src/main/kotlin/org/modelix/metamodel/generator/GeneratorInput.kt#L94-L137
`generatedName` is guaranteed to be unique.

slisson:
I think it's save to assume that there will be no role starting with `addTo`.

slisson (approved):

PR #381, opened 2023-12-21
Configure HTTP timeouts in bulk sync.

languitar:
Makes previously hard-coded HTTP timeouts for the model client configurable and increases the timeout for the bulk sync, where we see timeouts hitting with large models.


mhuster23 (approved):

languitar:
Can I manually trigger a release after this has been built on main to get access to these features?

mhuster23:
Yes, you just have to use the workflow dispatch trigger on main for the Release workflow.

PR #384, opened 2023-12-28
fix: initialize the server ID

odzhychko:
Clients using the v1 API expect server ID to be set.

See https://github.com/modelix/modelix.mps-plugins/blob/70293ee28914c4c5987f0cd67a500806a8a26411/mps-legacy-sync-plugin/src/main/java/org/modelix/model/mpsplugin/ModelServerConnection.kt#L168
See `ModelServerConnection#onConnect` in https://github.com/modelix/modelix.mps/blob/c594b1cc45222a3e8860d1bfb6468cd5bbb1bb77/mps/org.modelix.model.mpsplugin/models/org.modelix.model.mpsplugin.mps

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -24,7 +24,7 @@ repos:
         additional_dependencies: ["@commitlint/config-angular"]
         args: ["--config", "./commitlint.config.js"]
   - repo: https://github.com/macisamuele/language-formatters-pre-commit-hooks
-    rev: v2.10.0
+    rev: v2.12.0
     hooks:
       - id: pretty-format-kotlin
         args:
--- a/model-server/src/main/kotlin/org/modelix/model/server/handlers/ModelReplicationServer.kt
+++ b/model-server/src/main/kotlin/org/modelix/model/server/handlers/ModelReplicationServer.kt
@@ -85,7 +85,14 @@ class ModelReplicationServer(val repositoriesManager: RepositoriesManager) {
             call.respondText(storeClient.generateId("clientId").toString())
         }
         get("server-id") {
-            call.respondText(repositoriesManager.getServerId())
+            // Currently, the server ID is initialized in KeyValueLikeModelServer eagerly on startup.
+            // Should KeyValueLikeModelServer be removed or change,
+            // RepositoriesManager#maybeInitAndGetSeverId will initialize the server ID lazily on the first request.
+            //
+            // Functionally, it does not matter if the server ID is created eagerly or lazily,
+            // as long as the same server ID is returned from the same server.
+            val serverId = repositoriesManager.maybeInitAndGetSeverId()
+            call.respondText(serverId)
         }
         get("user-id") {
             call.respondText(call.getUserName() ?: call.request.origin.remoteHost)
--- a/model-server/src/test/kotlin/org/modelix/model/server/ModelClientTest.kt
+++ b/model-server/src/test/kotlin/org/modelix/model/server/ModelClientTest.kt
@@ -29,6 +29,7 @@ import org.modelix.model.server.store.LocalModelClient
 import java.util.Random
 import kotlin.test.Test
 import kotlin.test.assertEquals
+import kotlin.test.assertNotNull
 import kotlin.test.fail
 import kotlin.time.Duration.Companion.seconds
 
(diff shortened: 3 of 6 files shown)


.pre-commit-config.yaml
```diff
@@ -24,7 +24,7 @@ repos:
         additional_dependencies: ["@commitlint/config-angular"]
         args: ["--config", "./commitlint.config.js"]
   - repo: https://github.com/macisamuele/language-formatters-pre-commit-hooks
-    rev: v2.10.0
+    rev: v2.12.0
```
languitar, line 27: What does this bump fix for you?

model-server/src/main/kotlin/org/modelix/model/server/handlers/ModelReplicationServer.kt
```diff
@@ -85,7 +85,7 @@ class ModelReplicationServer(val repositoriesManager: RepositoriesManager) {
             call.respondText(storeClient.generateId("clientId").toString())
         }
         get("server-id") {
-            call.respondText(repositoriesManager.getServerId())
+            call.respondText(repositoriesManager.maybeInitAndGetSeverId())
```
languitar, line 88: Why is this a lazy operation on first request and not something done on application start?

model-server/src/test/kotlin/org/modelix/model/server/ModelClientTest.kt
```diff
@@ -104,6 +104,18 @@ class ModelClientTest {
         }
     }
 
+    @Test
+    fun `can retrieve server id initially`() = runTest {
+        val modelClient = createModelClient()
+        val serverId = modelClient.get("server-id")
+        assertNotNull(serverId)
```
languitar, line 111:
```suggestion
        val modelClient = createModelClient()
        
        val serverId = modelClient.get("server-id")
        
        assertNotNull(serverId)
```

For the arrange-act-assert structure?

.pre-commit-config.yaml
odzhychko: It fixed missing Python `setuptools` for me.
As it was only a local problem, I did not mention it in the PR.

https://github.com/macisamuele/language-formatters-pre-commit-hooks/commit/e43d17e9db27aebd8d73dff9c2f353a11204bc8e

languitar: Ah, I see. I'd be more in favor of putting such things into separate PRs to avoid polluting the interesting content discussions and diffs with such issues.

model-server/src/main/kotlin/org/modelix/model/server/handlers/ModelReplicationServer.kt
odzhychko: I guess, that there is no specific reason for this.
The requirement seams just to be that the same `server-id` is returned by the server.
This can be done by initializing it eagerly or lazy.

With my change in `KeyValueLikeModelServer::init` I do initialize it eagerly, because I deemed the code more robust.
If I want to do it lazy in `KeyValueLikeModelServer`, I need to create separate endpoints or conditions for the `SERVER_ID_KEY`, `LEGACY_SERVER_ID_KEY`, `LEGACY_SERVER_ID_KEY2`.
And everybody changing `maybeInitAndGetSeverId` would need to update those special cases.

model-server/src/test/kotlin/org/modelix/model/server/ModelClientTest.kt
odzhychko: Done

model-server/src/main/kotlin/org/modelix/model/server/handlers/ModelReplicationServer.kt
languitar: Maybe it would be good to put these explanations into comments or the commit message so the next one touching this piece of code knows why this way was chosen and not another one.

.pre-commit-config.yaml
odzhychko: If it helps, I'll do it that way in the future  😄

model-server/src/main/kotlin/org/modelix/model/server/handlers/ModelReplicationServer.kt
odzhychko: Done

languitar (approved):

model-server/src/main/kotlin/org/modelix/model/server/handlers/ModelReplicationServer.kt
languitar: That's a lot better to understand later on this way!

PR #385, opened 2023-12-28
fix(ts-model-api): publish the contents of the `dist` directory

odzhychko:
The old pattern only included the directory `dist` without its content.


languitar (commented):
Is the commit type correct for this fix? Were previous releases broken due to this? Or is this only a `build` commit?

odzhychko:
> Is the commit type correct for this fix? Were previous releases broken due to this? Or is this only a `build` commit?

Yes, the previous release was broken due to this.

I updated the commit message now to make it more clear.
> Because of that, the published ts-model-api package had no code and was broken.

languitar (approved):

PR #389, opened 2024-01-03
ci: set log level to INFO in the publish workflow

odzhychko:
We try to see if a retry is related to the failing publish workflows. https://issues.modelix.org/issue/MODELIX-612/Stabilize-publishing-pipeline

A retry would be logged at log level INFO.
https://github.com/gradle/gradle/blob/master/platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/repositories/transport/NetworkOperationBackOffAndRetry.java#L64

We do not use DEBUG because the DEBUG log level can expose security sensitive information to the console. https://docs.gradle.org/current/userguide/logging.html

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -31,6 +31,7 @@ jobs:
       - name: Build and Publish Artifacts
         run: >-
           ./gradlew --build-cache build publish
+          --info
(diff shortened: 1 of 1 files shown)


odzhychko:
Close in favor of #388

PR #390, opened 2024-01-03
fix(model-server): use one large transaction when putting many values

languitar:
So far, when putting multiple values via objects/put, a database transaction was created and committed for every value. That is unbelievably slow with a real database. This commit changes the implementation of the endpoint to use `putAll` instead of multiple `put` calls, effectively reducing the whole operation to a single transaction.

This is a `fix` commit and not a `perf` one as the previous implementation completely broke some known use cases due to ending in timeouts.

### DIFF
--- a/model-server/src/main/kotlin/org/modelix/model/server/handlers/ModelReplicationServer.kt
+++ b/model-server/src/main/kotlin/org/modelix/model/server/handlers/ModelReplicationServer.kt
@@ -225,26 +225,26 @@ class ModelReplicationServer(val repositoriesManager: RepositoriesManager) {
                 }
                 route("objects") {
                     put {
-                        var writtenEntries = 0
-                        withContext(Dispatchers.IO) {
-                            var isKey = true
-                            var key = ""
-                            call.receiveStream().bufferedReader().lineSequence().forEach { line ->
-                                if (isKey) {
-                                    key = line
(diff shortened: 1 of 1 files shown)


mhuster23 (approved):
This will use more memory as everything needs to be stored in the map first, but it's a tradeoff, that we probably have to make.

languitar:
We could try to handle everything as a sequence as one further optimization. But that would also require extending the `IStoreClient` interface with a `putAll` version that accepts a stream of `Pair<String, String?>`.

languitar:
That optimization won't work easily as the used IgniteCache only has an eager putAll signature.

odzhychko:
> This will use more memory as everything needs to be stored in the map first, but it's a tradeoff, that we probably have to make.

> We could try to handle everything as a stream as one further optimization. But that would also require extending the IStoreClient interface with a putAll version that accepts a stream of Pair<String, String?>.

Have you looked into`IStoreClient.runTransaction`?
It does not work out of the box in the couroutine scope. But in general, that should be something doable.

languitar:
`runTransaction` is used internally inside `put` and `putAll`. Just wrapping a `runTransaction` around many `put` calls already gave a speedup, but using `putAll` was still a lot faster and saturated the CPU more than just keeping the transaction alive.

PR #394, opened 2024-01-08
MODELIX-678 model-server content explorer raises an error when trying to unfold the root node of an empty repository

mhuster23:


languitar (commented):
This should probably be covered by a test in the end.

model-server/src/main/kotlin/org/modelix/model/server/handlers/ContentExplorer.kt
```diff
@@ -112,19 +98,24 @@ class ContentExplorer(private val client: IModelClient, private val repoManager:
...
-                    val candidate = PNodeAdapter(id, node.branch).takeIf { it.isValid }
-                    if (candidate != null) {
-                        found = candidate
-                        break
-                    }
+                val id = call.parameters["nodeId"]?.toLongOrNull()
+                    ?: return@get call.respondText("node id not found", status = HttpStatusCode.NotFound)
+
+                val versionHash = call.parameters["versionHash"]
+                    ?: return@get call.respondText("version hash not found", status = HttpStatusCode.NotFound)
```
languitar, line 105: Another solution to the ugly `return@` syntax would be installing the ErrorPages plugin and then mapping custom Exceptions to error responses. This would result in being able to raise specific exceptions here.

languitar (approved):

PR #395, opened 2024-01-08
MODELIX-680 Transfer model data as streams instead of JSON

slisson:
With the changes in this PR I can successfully run the bulk import on a large model also after the initial import.

### DIFF
--- a/model-client/src/commonMain/kotlin/org/modelix/model/client2/ModelClientV2.kt
+++ b/model-client/src/commonMain/kotlin/org/modelix/model/client2/ModelClientV2.kt
@@ -21,21 +21,28 @@ import io.ktor.client.plugins.HttpTimeout
 import io.ktor.client.plugins.ResponseException
 import io.ktor.client.plugins.contentnegotiation.ContentNegotiation
 import io.ktor.client.plugins.expectSuccess
+import io.ktor.client.request.HttpRequestBuilder
 import io.ktor.client.request.get
 import io.ktor.client.request.post
 import io.ktor.client.request.put
 import io.ktor.client.request.setBody
+import io.ktor.client.statement.HttpResponse
+import io.ktor.client.statement.bodyAsChannel
 import io.ktor.client.statement.bodyAsText
 import io.ktor.http.ContentType
+import io.ktor.http.HttpHeaders
 import io.ktor.http.HttpStatusCode
 import io.ktor.http.URLBuilder
 import io.ktor.http.appendPathSegments
 import io.ktor.http.contentType
 import io.ktor.http.takeFrom
 import io.ktor.serialization.kotlinx.json.json
+import io.ktor.utils.io.readUTF8Line
 import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.emptyFlow
+import kotlinx.coroutines.flow.flow
 import kotlinx.coroutines.launch
-import kotlinx.serialization.json.Json
 import org.modelix.kotlin.utils.DeprecationInfo
 import org.modelix.model.IVersion
 import org.modelix.model.api.IIdGenerator
@@ -54,6 +61,8 @@ import org.modelix.model.operations.OTBranch
 import org.modelix.model.persistent.HashUtil
 import org.modelix.model.persistent.MapBasedStore
 import org.modelix.model.server.api.v2.VersionDelta
+import org.modelix.model.server.api.v2.VersionDeltaStream
+import org.modelix.model.server.api.v2.asStream
 import org.modelix.modelql.client.ModelQLClient
 import org.modelix.modelql.core.IMonoStep
 import kotlin.time.Duration
@@ -117,9 +126,9 @@ class ModelClientV2(
                 takeFrom(baseUrl)
                 appendPathSegmentsEncodingSlash("repositories", repository.id, "init")
             }
+            useVersionStreamFormat()
         }
-        val delta = response.body<VersionDelta>()
-        return createVersion(null, delta)
+        return createVersion(null, response.readVersionDelta())
     }
 
     override suspend fun listRepositories(): List<RepositoryId> {
@@ -151,9 +160,9 @@ class ModelClientV2(
                     parameters["lastKnown"] = (baseVersion as CLVersion).getContentHash()
                 }
             }
+            useVersionStreamFormat()
         }
-        val delta = Json.decodeFromString<VersionDelta>(response.bodyAsText())
-        return createVersion(baseVersion as CLVersion?, delta)
+        return createVersion(baseVersion as CLVersion?, response.readVersionDelta())
     }
 
     override suspend fun loadVersion(
@@ -169,9 +178,9 @@ class ModelClientV2(
                     parameters["lastKnown"] = (baseVersion as CLVersion).getContentHash()
                 }
             }
+            useVersionStreamFormat()
         }
-        val delta = Json.decodeFromString<VersionDelta>(response.bodyAsText())
-        return createVersion(baseVersion as CLVersion?, delta)
+        return createVersion(baseVersion as CLVersion?, response.readVersionDelta())
     }
 
     override suspend fun push(branch: BranchReference, version: IVersion, baseVersion: IVersion?): IVersion {
@@ -193,11 +202,11 @@ class ModelClientV2(
                 takeFrom(baseUrl)
                 appendPathSegmentsEncodingSlash("repositories", branch.repositoryId.id, "branches", branch.branchName)
             }
+            useVersionStreamFormat()
             contentType(ContentType.Application.Json)
             setBody(delta)
         }
-        val mergedVersionDelta = response.body<VersionDelta>()
-        return createVersion(version, mergedVersionDelta)
+        return createVersion(version, response.readVersionDelta())
     }
 
     private suspend fun uploadObjects(repository: RepositoryId, objects: Sequence<Pair<String, String>>) {
@@ -224,8 +233,9 @@ class ModelClientV2(
                     parameters["lastKnown"] = lastKnownVersion.hash
                 }
             }
+            useVersionStreamFormat()
         }
-        val receivedVersion = createVersion(lastKnownVersion, response.body())
+        val receivedVersion = createVersion(lastKnownVersion, response.readVersionDelta())
         LOG.debug { "${clientId.toString(16)}.pull($branch, $lastKnownVersion) -> $receivedVersion" }
         return receivedVersion
     }
@@ -237,11 +247,12 @@ class ModelClientV2(
                 takeFrom(baseUrl)
                 appendPathSegmentsEncodingSlash("repositories", branch.repositoryId.id, "branches", branch.branchName)
             }
+            useVersionStreamFormat()
         }
 
         val receivedVersion = when (response.status) {
             HttpStatusCode.NotFound -> null
-            HttpStatusCode.OK -> createVersion(null, response.body())
+            HttpStatusCode.OK -> createVersion(null, response.readVersionDelta())
             else -> throw ResponseException(response, response.bodyAsText())
         }
         LOG.debug { "${clientId.toString(16)}.pullIfExists($branch) -> $receivedVersion" }
@@ -284,8 +295,9 @@ class ModelClientV2(
                     parameters["lastKnown"] = lastKnownVersion.hash
                 }
             }
+            useVersionStreamFormat()
         }
-        val receivedVersion = createVersion(lastKnownVersion, response.body())
+        val receivedVersion = createVersion(lastKnownVersion, response.readVersionDelta())
         LOG.debug { "${clientId.toString(16)}.poll($branch, $lastKnownVersion) -> $receivedVersion" }
         return receivedVersion
     }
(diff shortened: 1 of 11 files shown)


model-server/src/main/kotlin/org/modelix/model/server/CmdLineArgs.kt
```diff
@@ -21,6 +21,9 @@ internal class CmdLineArgs {
     @Parameter(names = ["-inmemory"], description = "Use in-memory storage", converter = BooleanConverter::class)
     var inmemory = false
 
+    @Parameter(names = ["--local-persistence"], description = "Use ignite with local disk persistence", converter = BooleanConverter::class)
+    var localPersistence = false
+
```
languitar, line 26: Does the command line parsing have some feature like mutual exclusion to prevent users from providing `inmemory` and `--local-persistence` at the same time?

Why does this use POSIX double dashes instead of the single dash used for all other options?

model-client/src/commonMain/kotlin/org/modelix/model/client2/ModelClientV2.kt
```diff
@@ -192,12 +197,12 @@ class ModelClientV2(
             url {
                 takeFrom(baseUrl)
                 appendPathSegmentsEncodingSlash("repositories", branch.repositoryId.id, "branches", branch.branchName)
+                parameters["format"] = "plain"
```
languitar, line 200: Some constant / enum for "plain" etc.?

model-client/src/commonMain/kotlin/org/modelix/model/client2/ModelClientV2.kt
```diff
@@ -328,6 +336,34 @@ class ModelClientV2(
...
+            if (isHash) {
+                lastHash = it
+                if (firstHash == null) {
+                    firstHash = it
+                }
+            } else {
+                val value = it
+                store.keyValueStore.put(lastHash!!, value)
+            }
+            isHash = !isHash
```
languitar, line 353: Is `windowed` available on flows? The mutability and the swapping of the boolean can then maybe be expressed in a way that is a bit easier to read.

model-client/src/commonMain/kotlin/org/modelix/model/client2/ModelClientV2.kt
```diff
@@ -451,3 +487,14 @@ suspend fun <T> IModelClientV2.runWrite(branchRef: BranchReference, body: (INode
...
 }
+
+private suspend fun HttpResponse.bodyLinesAsFlow(): Flow<String> {
+    return flow {
+        val channel = bodyAsChannel()
+        var line: String? = channel.readUTF8Line()
+        while (line != null) {
+            emit(line)
+            line = channel.readUTF8Line()
+        }
```
languitar, line 498: Maybe this can be realized without mutability when using `toInputStream().forEachLine`.

model-datastructure/src/commonMain/kotlin/org/modelix/model/lazy/CLVersion.kt
```diff
@@ -398,7 +398,7 @@ private fun trackAccessedEntries(store: IKeyValueStore, body: (IDeserializingKey
 }
 
 private class AccessTrackingStore(val store: IKeyValueStore) : IKeyValueStore {
-    val accessedEntries: MutableMap<String, String?> = HashMap()
+    val accessedEntries: MutableMap<String, String?> = LinkedHashMap()
```
languitar, line 401: Can you add a comment on why this implementation was chosen. Seems important if it changed and the next one reading the code should probably know why this decision was made.

model-server/src/main/kotlin/org/modelix/model/server/handlers/ModelReplicationServer.kt
```diff
@@ -275,12 +277,48 @@ class ModelReplicationServer(val repositoriesManager: RepositoriesManager) {
...
+                        } else {
+                            appendLine()
+                            append(it.first)
+                        }
+                        appendLine()
+                        append(it.second)
+                    }
+                }
+            }
+            null, "", "json" -> {
```
languitar, line 302: Could also use the constants / enum that's then used in the client, too

model-server/src/main/kotlin/org/modelix/model/server/handlers/ModelReplicationServer.kt
```diff
@@ -275,12 +277,48 @@ class ModelReplicationServer(val repositoriesManager: RepositoriesManager) {
...
 
     private suspend fun ApplicationCall.respondDelta(versionHash: String, baseVersionHash: String?) {
-        val delta = VersionDelta(
-            versionHash,
-            baseVersionHash,
-            objectsMap = repositoriesManager.computeDelta(versionHash, baseVersionHash),
-        )
-        delta.checkObjectHashes()
-        respond(delta)
+        when (val format = request.queryParameters["format"]) {
```
languitar, line 280: What about using content negotiation instead to use more standard features?

model-server/src/main/kotlin/org/modelix/model/server/handlers/ModelReplicationServer.kt
```diff
@@ -275,12 +277,48 @@ class ModelReplicationServer(val repositoriesManager: RepositoriesManager) {
...
+                            first = false
+                            check(it.first == versionHash) { "First object should be the version" }
+                        } else {
+                            appendLine()
+                            append(it.first)
+                        }
+                        appendLine()
+                        append(it.second)
+                    }
+                }
```
languitar, line 300: I'd factor this into a separate method to avoid this endpoint from growing.

model-server/src/main/kotlin/org/modelix/model/server/handlers/ModelReplicationServer.kt
```diff
@@ -275,12 +277,48 @@ class ModelReplicationServer(val repositoriesManager: RepositoriesManager) {
...
-            baseVersionHash,
-            objectsMap = repositoriesManager.computeDelta(versionHash, baseVersionHash),
-        )
-        delta.checkObjectHashes()
-        respond(delta)
+        when (val format = request.queryParameters["format"]) {
+            "versionHash" -> {
+                respondText(versionHash)
+            }
+            "plain" -> {
```
languitar, line 284: What about a test for the new format?

model-server/src/main/kotlin/org/modelix/model/server/handlers/RepositoriesManager.kt
```diff
@@ -200,18 +207,41 @@ class RepositoriesManager(val client: LocalModelClient) {
             ?: throw IllegalStateException("No version found for branch '${branch.branchName}' in repository '${branch.repositoryId}'")
     }
 
-    private val deltaCache = LRUMap<Pair<String, String?>, Lazy<Map<String, String>>>(10)
-    fun computeDelta(versionHash: String, baseVersionHash: String?): Map<String, String> {
+    private val deltaCache = LRUMap<Pair<String, String?>, SoftReference<Lazy<Map<String, String>>>>(10)
```
languitar, line 210: Would it make sense to make the cache size configurable?

model-server/src/main/kotlin/org/modelix/model/server/handlers/RepositoriesManager.kt
```diff
@@ -200,18 +207,41 @@ class RepositoriesManager(val client: LocalModelClient) {
...
+                val version = CLVersion(versionHash, client.storeCache)
+                emitObjects(version.treeHash!!)
+            }
+        }
+
         return runSynchronized(deltaCache) {
-            deltaCache.getOrPut(versionHash to baseVersionHash) {
+            var cachedResult: Lazy<Map<String, String>>?
+            val key = versionHash to baseVersionHash
+            cachedResult = deltaCache.get(key)?.get()
```
languitar, line 232: Somehow GitHub doesn't let me add suggestions... but why not:

```kotlin
val cachedResult = deltaCache.get(key)?.get() ?: {
   ...
}
```

Avoids the mutability and reduces the scope of the variable.

model-client/src/commonMain/kotlin/org/modelix/model/client2/ModelClientV2.kt
slisson: No, it's not.

model-datastructure/src/commonMain/kotlin/org/modelix/model/lazy/CLVersion.kt
slisson: If you change it back there will be an exception thrown in some tests that clearly tell you the reason. It's not strongly necessary that this is map is ordered it's just the easiest way to fix the tests. It would also be OK to use an unordered map and fix the failing tests in some other way. Since this is private code I think it's OK not to specify this behavior explicitly.

model-server/src/main/kotlin/org/modelix/model/server/handlers/ModelReplicationServer.kt
slisson: Not sure how that would look like.

model-server/src/main/kotlin/org/modelix/model/server/handlers/ModelReplicationServer.kt
slisson: The new format is the default one for new versions of the client. The old "JSON" format just exists for not breaking old versions of the client, but should be removed in the future. This means "plain" is already. Testing the deprecated "JSON" format would require adding support for switching the format in the client, which means adding additional deprecated code.

model-server/src/main/kotlin/org/modelix/model/server/handlers/RepositoriesManager.kt
slisson: This will soon also be replaced by a more efficient stream based implementation as part of [MODELIX-681](https://issues.modelix.org/issue/MODELIX-681/Explicit-partial-loading).

model-client/src/commonMain/kotlin/org/modelix/model/client2/ModelClientV2.kt
slisson: Extracted it into a new function to remove the duplication of the string literals.

model-client/src/commonMain/kotlin/org/modelix/model/client2/ModelClientV2.kt
slisson: Not creating a new list for each key/value pair (as windowed does) is probably also relevant for the performance.

model-server/src/main/kotlin/org/modelix/model/server/handlers/ModelReplicationServer.kt
slisson: It's already a separate method.

model-client/src/commonMain/kotlin/org/modelix/model/client2/ModelClientV2.kt
languitar: No idea how much the compiler could/would optimize here. But as it's not available, there's nothing to be done here.

model-server/src/main/kotlin/org/modelix/model/server/handlers/ModelReplicationServer.kt
languitar: Basically, we could defined different MIME types and request them using the "Accept" HTTP header instead of the query parameter.

model-server/src/main/kotlin/org/modelix/model/server/handlers/ModelReplicationServer.kt
languitar: Could also be tested with a plain HTTP test towards the server.

model-server/src/main/kotlin/org/modelix/model/server/handlers/ModelReplicationServer.kt
languitar: I see all three formats being implemented in the same route function in ktor?

languitar:
Any chance to resolve the conflicts and remaining discussion topics today? Then I can give that a shot tomorrow in the real environment.

model-server/src/main/kotlin/org/modelix/model/server/CmdLineArgs.kt
slisson: The other parameters should probably also use double dashes. I just dropped the commit as it is out of scope for this PR.

model-server/src/main/kotlin/org/modelix/model/server/handlers/ModelReplicationServer.kt
slisson: I refactored it a bit, but the method is now just 30 lines. Splitting it wouldn't improve the readability.

languitar (commented):
https://github.com/modelix/modelix.core/blob/main/api/model-server.yaml would need changes to reflect the added query parameter and the variability in response content types.

model-server/src/main/kotlin/org/modelix/model/server/CmdLineArgs.kt
slisson: Added it back, but also changed the names of the other parameters to use double dashes to fix the inconsistency.

model-server/src/main/kotlin/org/modelix/model/server/handlers/ModelReplicationServer.kt
slisson: There is now a new MIME type `application/x-modelix-objects`. The client sends an `Accept: application/x-modelix-objects` header to the server and the server responds with a `Content-Type: application/x-modelix-objects` header.

If the client accepts `application/json`, but not `application/x-modelix-objects` (which old client versions do), the server responds with as it did before.
If the client doesn't accept `application/json` (the user probably opened the URL in the browser) the server responds with the new format.

languitar (changes requested):
Looks fine now. Apart from one thing: the new content types and response formats need to be declared in the OpenAPI specification.

PR #399, opened 2024-01-10
MODELIX-685 Resource based routing breaks docker container of model-server 

nkoester:
Fixes regression introduced by #353

### DIFF
--- a/model-server/build.gradle.kts
+++ b/model-server/build.gradle.kts
@@ -122,8 +122,8 @@ val cucumber = task("cucumber") {
 tasks.register<Copy>("copyApis") {
     from("../api/")
     include("*.yaml")
-    into(project.layout.projectDirectory.dir("build/openapi/src/main/resources/api"))
-    sourceSets["main"].resources.srcDir("build/openapi/src/main/resources/")
+    into(project.layout.buildDirectory.dir("openapi/src/main/resources/api"))
+    sourceSets["main"].resources.srcDir(project.layout.buildDirectory.dir("openapi/src/main/resources/"))
 }
 
 tasks.named("runKtlintCheckOverMainSourceSet") {
--- a/model-server/build.gradle.kts
+++ b/model-server/build.gradle.kts
@@ -120,11 +120,15 @@ val cucumber = task("cucumber") {
 tasks.register<Copy>("copyApis") {
     from("../api/")
     include("*.yaml")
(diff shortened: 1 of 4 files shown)


model-server/build.gradle.kts
```diff
@@ -117,8 +117,15 @@ val cucumber = task("cucumber") {
     }
 }
 
+tasks.register<Copy>("copyApis") {
+    from("../api/")
+    include("*.yaml")
+    into(project.layout.projectDirectory.dir("api"))
```
languitar, line 123: Why not putting this into the resource folder?

nkoester: makes more sense and also reduces complexity of the dockefile

model-server/build.gradle.kts
```diff
@@ -117,8 +117,19 @@ val cucumber = task("cucumber") {
     }
 }
 
+tasks.register<Copy>("copyApis") {
+    from("../api/")
+    include("*.yaml")
+    into(project.layout.projectDirectory.dir("src/main/resources/api"))
```
languitar, line 123: This should go under the build directory to avoid accidentally committing the generated files. The srcSets need to be extended then with the new source root that is created there.

model-server/build.gradle.kts
```diff
@@ -117,8 +117,30 @@ val cucumber = task("cucumber") {
     }
 }
 
+// copies the openAPI specifications from the api folder into a resource
+// folder so that they are packaged and deployed with the model-server
+tasks.register<Copy>("copyApis") {
+    from("../api/")
+    include("*.yaml")
+    into(project.layout.projectDirectory.dir("build/openapi/src/main/resources/api"))
```
languitar, line 125:
```suggestion
    into(project.layout.buildDirectory.dir("openapi/src/main/resources/api"))
```

model-server/build.gradle.kts
```diff
@@ -117,8 +117,30 @@ val cucumber = task("cucumber") {
...
+    sourceSets["main"].resources.srcDir("build/openapi/src/main/resources/")
```
languitar, line 126:
```suggestion
    sourceSets["main"].resources.srcDir(project.layout.buildDirectory.dir("openapi/src/main/resources/"))
```

languitar (dismissed):
One more idea to get rid of the explicit "build" string

languitar (approved):

PR #401, opened 2024-01-15
MODELIX-701 Content explorer shows no content

mhuster23:
The route to post the expanded nodes was not found after #353.

### DIFF
--- a/model-server/src/test/kotlin/org/modelix/model/server/handlers/ContentExplorerTest.kt
+++ b/model-server/src/test/kotlin/org/modelix/model/server/handlers/ContentExplorerTest.kt
@@ -19,6 +19,10 @@ package org.modelix.model.server.handlers
 import io.ktor.client.call.body
 import io.ktor.client.request.get
 import io.ktor.client.request.post
+import io.ktor.client.request.setBody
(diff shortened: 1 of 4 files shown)


model-server/src/test/kotlin/org/modelix/model/server/handlers/ContentExplorerTest.kt
```diff
@@ -71,4 +77,31 @@ class ContentExplorerTest {
...
+        val nodeId = checkNotNull(version.getTree().root?.id)
+
+        val expandedNodes = ContentExplorerExpandedNodes(setOf(nodeId.toString()), false)
+
+        val response = client.post("/content/$versionHash/") {
+            contentType(ContentType.Application.Json)
+            setBody(expandedNodes)
+        }
+
+        val html = Input.fromString(response.bodyAsText().replace("<br>", "")).build() // br leads to parser exception
```
languitar, line 100: The BR thing probably happens because you use a strict XML parse and not a parser specifically tailored to HTML The next change to the HTML structure can result in the next error of this kind. It would be better to switch to dedicated HTML parsing to avoid these issues: https://stackoverflow.com/a/28307130/283649

PR #403, opened 2024-01-16
MODELIX-572 Upgrade model-server Docker dependencies with dependabot

mhuster23:
This will keep our model-server image up-to-date.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,3 +5,8 @@ updates:
     directory: "/"
     schedule:
       interval: "daily"
+  - package-ecosystem: "docker"
+    open-pull-requests-limit: 20
+    directory: "/"
+    schedule:
+      interval: "daily"


.github/dependabot.yml
```diff
@@ -5,3 +5,8 @@ updates:
     directory: "/"
     schedule:
       interval: "daily"
+  - package-ecosystem: "docker"
+    open-pull-requests-limit: 20
+    directory: "/"
```
odzhychko, line 10: I'm not sure if `/` will detect `model-server/Dockerfile`.
Maybe `directory` needs to be `/model-server`.

I guess we can just merge in try it out.

odzhychko (approved):

.github/dependabot.yml
mhuster23: I was wondering the same thing. But there currently is [no way to test this before merging](https://github.com/dependabot/dependabot-core/issues/4605), so we'll have to see.

PR #404, opened 2024-01-16
ci: fix dependabot docker directory

mhuster23:
followup for #403 as dependabot cannot find the Dockerfile


languitar (approved):

mhuster23:
Also nice coincidence that PR 404 is about sth not found 😜

PR #408, opened 2024-01-17
ci: test the publish workflow in PRs if changed

languitar:
This PR runs the GitHub Actions publish workflow in case we have changed it in a PR. Without this, testing changes to this workflow is guess work until a release appeared. The workflow has been modified to only publish to maven local in PR runs to avoid spamming repositories with artificial testing releases. That way, we don't test everything about publishing, but a complete misconfiguration of the publishing tasks or pipeline should become visible. Docker publishing is not tested and skipped here as I couldn't find an easy way to not push things to our repos.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -5,6 +5,10 @@ on:
     tags:
       - '[0-9]+.[0-9]+.[0-9]+*'
   workflow_dispatch:
+  # Test this workflow in PRs in case it changed
+  pull_request:
+    paths:
+      - .github/workflows/publish.yml
 
 jobs:
   newRelease:
@@ -26,8 +30,16 @@ jobs:
           java-version: '11'
       - name: Set up Gradle
         uses: gradle/gradle-build-action@v2
-      - name: Use tag as version
-        run: echo "${GITHUB_REF#refs/*/}" > version.txt
+      - name: Configure the project version
+        env:
+          IS_PR: ${{ github.event_name == 'pull_request' }}
+        run: |-
+          if [[ "$IS_PR" = 'true' ]]; then
+            echo "0.0.1-SNAPSHOT" > version.txt
+          else
+            echo "${GITHUB_REF#refs/*/}" > version.txt
+          fi
+          cat version.txt
         # Perform the build in a separate call to avoid trying to publish
         # something where the build already failed partially. This could happen
         # due to the use of the --continue flag in the publish step.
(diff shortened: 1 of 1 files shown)


mhuster23 (commented):
I'm not sure if this PR is actually beneficial.
We already publish to maven local for the plugin tests so a misconfiguration of the tasks would become apparent there.
On top of that, due to the added if statements, it makes it harder for a human to parse, what the workflow does 99% of the time.

languitar:
This CI workflow would not be running at all before a release. And correctly defining a GitHub Actions workflow is often not easy and requires many iteration with GitHub actually executing the workflow to detect issues. We would only see the misconfigurations of the workflow by triggering many releases, thereby polluting the public changelog and forcing users into more updates. With the experience I have so far with the complexity of reconfiguring such workflows I'd say it's worth running this here, even if it is not a 100% reproduction of the actual release workflow.

.github/workflows/publish.yml
```diff
@@ -27,7 +31,13 @@ jobs:
       - name: Set up Gradle
         uses: gradle/gradle-build-action@v2
       - name: Use tag as version
+        # Set a production version only when running on tags
+        if: ${{ github.event_name != 'pull_request' }}
         run: echo "${GITHUB_REF#refs/*/}" > version.txt
+      - name: Use dummy version for PR
+        # Set a dummy version that doesn't hurt if accidentally released
+        if: ${{ github.event_name == 'pull_request' }}
+        run: echo "0.0.1-SNAPSHOT" > version.txt
```
mhuster23, line 40: I think it would be nicer to have only one step with an `if` inside of the `run`, so we don't have at least one skipped step for every execution.

languitar:
Any other opinions on whether this is worth it?

odzhychko:
> Any other opinions on whether this is worth it?

Not a strong opinion, but I favor of the change.
(I broke accidentally broke publishing myself a few times.)
The logic in the Workflow file is not too complicated.

.github/workflows/publish.yml
```diff
@@ -46,27 +56,36 @@ jobs:
...
-          -Pgpr.user=${{ github.actor }}
-          -Pgpr.key=${{ secrets.GITHUB_TOKEN }}
-          -Pgpr.universalkey=${{ secrets.GHP_UNIVERSAL_PUBLISH_TOKEN }}
-          -Porg.gradle.internal.http.connectionTimeout=180000
-          -Porg.gradle.internal.http.socketTimeout=180000
+        run: |-
+          if [[ "$IS_PR" = 'true' ]]; then
+            TARGET=publishToMavenLocal
+          else
+            TARGET=publish
```
odzhychko, line 63: If we publish locally on each PR we do not need the "build.yaml" workflow, rigth?
Should we merge `build.yaml` and `publish.yaml`?

languitar: That makes the trigger conditions a bit more complicated as the build also runs on main.

mhuster23: With these changes `publish.yaml`won't trigger on each PR. Only if the file itself was changed.

languitar: So what shall we do about this? I'd say dropping this workflow and putting it into the build workflow has a higher potential for accidentally publishing undesired stuff to the Maven repos when fiddling with the build workflow.

odzhychko: > That makes the trigger conditions a bit more complicated as the build also runs on main.

> With these changes publish.yamlwon't trigger on each PR. Only if the file itself was changed.

> I'd say dropping this workflow and putting it into the build workflow has a higher potential for accidentally publishing undesired stuff to the Maven repos when fiddling with the build workflow.

I agree with your points. Let's keep `build.yaml` and `publish.yaml` separate.

odzhychko (approved):

PR #409, opened 2024-01-17
MODELIX-709 Use dependabot for npm and github-actions

mhuster23:
Unfortunately dependabot currently does not support adding multiple directories, so the config is a bit repetitive.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,3 +10,8 @@ updates:
     directory: "/model-server"
     schedule:
       interval: "daily"
+  - package-ecosystem: "github-actions"
+    open-pull-requests-limit: 20
+    directory: "/"
+    schedule:
+      interval: "daily"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -15,3 +15,30 @@ updates:
     directory: "/"
     schedule:
       interval: "daily"
+  # We use the default versioning strategy for npm (increase for apps and widen for libraries),
+  # so that clients are free to select an appropriate version from the allowed range, when they use one of the libraries.
+  - package-ecosystem: "npm"
+    open-pull-requests-limit: 20
+    directory: "/"
+    schedule:
+      interval: "daily"
+  - package-ecosystem: "npm"
+    open-pull-requests-limit: 20
+    directory: "/ts-model-api"
+    schedule:
+      interval: "daily"
+  - package-ecosystem: "npm"
+    open-pull-requests-limit: 20
+    directory: "/vue-model-api"
+    schedule:
+      interval: "daily"
+  - package-ecosystem: "npm"
+    open-pull-requests-limit: 20
+    directory: "/model-api-gen-gradle-test/typescript-generation"
+    schedule:
+      interval: "daily"
+  - package-ecosystem: "npm"
+    open-pull-requests-limit: 20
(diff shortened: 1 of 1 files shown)


languitar:
Yet another reason for switching to renovate. Shouldn't we give that a try here?

mhuster23:
I think we're going to try it in other repositories, since we require stuff there, that isn't covered by dependabot (e.g. docker compose). (see [MODELIX-704](https://issues.modelix.org/issue/MODELIX-704/Automatically-update-dependencies-in-modelix.workspaces-and-modelix.kubernetes)) 
If it works well for us there, we can change it here too. For now I would stick to dependabot.

.github/dependabot.yml
```diff
@@ -10,3 +10,38 @@ updates:
...
+    open-pull-requests-limit: 20
+    directory: "/"
+    schedule:
+      interval: "daily"
+  - package-ecosystem: "npm"
+    open-pull-requests-limit: 20
+    directory: "/"
+    schedule:
+      interval: "daily"
+    versioning-strategy: "increase"
```
languitar, line 23: Some explanation on why this strategy was chosen would be good. The default `auto` would actually use `widen` for libraries, which we provide, so that clients are free to select an appropriate version from the allowed range when they use one of the libraries.

languitar (changes requested):

.github/dependabot.yml
mhuster23: I don't have a strong opinion on the versioning-strategy. If you think another one is more appropriate, I will change it. @odzhychko what do you think, since this affects components you developed?

odzhychko: > `auto`	Try to differentiate between apps and libraries. Use increase for apps and widen for libraries.

> so that clients are free to select an appropriate version from the allowed range when they use one of the libraries.

I would use `auto`, because of what languitar said about libraries.

In the upcoming automatically created PRs we should check, if Dependabot appropriately differentiated between libraries and applications. If not, we would need to upgrade the version appropriately by ourselves. I did not see that we could override the `versioning-strategy` for specific libraries.

languitar (approved):
Should we somehow preserve the decision for the versioning strategy in some comment? The next time someone looks at the config, that information is evident.

odzhychko:
> Should we somehow preserve the decision for the versioning strategy in some comment? The next time someone looks at the config, that information is evident.

Definitly. I will write an ADR about it.

languitar:
Maybe a simple comment in the dependabot file is also enough in this case? ;-)

.github/dependabot.yml
```diff
@@ -10,3 +10,35 @@ updates:
     directory: "/model-server"
     schedule:
       interval: "daily"
+  - package-ecosystem: "github-actions"
+    open-pull-requests-limit: 20
+    directory: "/"
+    schedule:
+      interval: "daily"
+  # We use the default versioning strategy for npm (increase for apps and widen) for libraries,
```
languitar, line 18:
```suggestion
  # We use the default versioning strategy for npm (increase for apps and widen for libraries),
```

languitar (approved):

languitar (approved):

odzhychko:
> Maybe a simple comment in the dependabot file is also enough in this case? ;-)

Even better ;D

PR #412, opened 2024-01-19
MODELIX-714 ModelQl query on model-server fails with "issue getting concept"

mhuster23:
These edge cases were not caught by the tests, because the client and server share the same `TypedLanguagesRegistry` as it is a a singleton.
In production the model-server would have no knowledge about the used languages, but it has them in our test.
If we wanted to test these cases automatically, we would probably have to test this with a real model-server, which would be quite expensive.

### DIFF
--- a/model-api-gen-gradle-test/ci.sh
+++ b/model-api-gen-gradle-test/ci.sh
@@ -2,6 +2,19 @@
 
 set -e
 (
-  cd "$(dirname "$0")"
-  ./gradlew build
+  TEST_DIR="$(dirname "$(readlink -f "$0")")"
+  cd "$TEST_DIR/../"
+  if [ "${CI}" != "true" ]; then
+    trap cleanup INT TERM EXIT
+    cleanup () {
+      kill "${MODEL_SERVER_PID}"
+      exit
+    }
+  fi
+  ./gradlew :model-server:run --console=plain --args="-inmemory -port 28102" &
+  MODEL_SERVER_PID=$!
+
+  curl -X GET --retry 30 --retry-connrefused --retry-delay 1 http://localhost:28102/health
+  cd "$TEST_DIR"
+  ./gradlew build --console=plain --stacktrace
 )
--- a/model-api/src/commonMain/kotlin/org/modelix/model/api/INode.kt
+++ b/model-api/src/commonMain/kotlin/org/modelix/model/api/INode.kt
@@ -323,8 +323,17 @@ fun INode.resolveProperty(role: String): IProperty {
         ?: throw RuntimeException("Property '$role' not found in concept ${c.getLongName()}")
 }
 
+/**
+ * Attempts to resolve the child link.
+ * @return resolved child link
+ *         or null, if this concept has no child link or an exception was thrown during concept resolution
+ */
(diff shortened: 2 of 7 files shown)


model-api/src/commonMain/kotlin/org/modelix/model/api/INode.kt
```diff
@@ -324,7 +324,11 @@ fun INode.resolveProperty(role: String): IProperty {
 }
 
 fun INode.tryResolveChildLink(role: String): IChildLink? {
-    val c = this.concept ?: return null
+    val c = try {
+        this.concept ?: return null
+    } catch (e: RuntimeException) {
+        return null
```
slisson, line 330: This doesn't look like a proper fix. It just makes it harder to debug if there is a missing language. The relation between MODELIX-714 and this change is also not clear, so it's hard to propose an alternative solution.

languitar: To avoid such design discussion in the future, it would probably be good to specify what the `try` in the method names of the patched methods is intended to do. With this patch, `try` is extended to include the cases where a concept is unknown.

model-api-gen-gradle-test/ci.sh
languitar, line 1: It's a bit unfortunate that a bug in the model-server is now effectively tested as part of the api-gen tests. Should we open up a follow-up ticket to create some real e2e tests?

model-client/src/commonMain/kotlin/org/modelix/model/client2/ModelClientV2.kt
```diff
@@ -137,7 +137,7 @@ class ModelClientV2(
                 takeFrom(baseUrl)
                 appendPathSegments("repositories")
             }
-        }.bodyAsText().lines().map { RepositoryId(it) }
+        }.bodyAsText().lines().filter { it.isNotEmpty() }.map { RepositoryId(it) }
```
languitar, line 140: Why can empty lines happen?

languitar (approved):

model-client/src/commonMain/kotlin/org/modelix/model/client2/ModelClientV2.kt
mhuster23: I explained it in the commit message.

model-api-gen-gradle-test/ci.sh
mhuster23: [MODELIX-723](https://issues.modelix.org/issue/MODELIX-723/Create-end-to-end-tests-for-model-server)

model-client/src/commonMain/kotlin/org/modelix/model/client2/ModelClientV2.kt
languitar: So the response includes a single newline and lines emits one empty line in that case?

mhuster23: I don't remember exactly anymore but I think the response body was empty. So `lines` returns the empty string.

PR #439, opened 2024-01-22
ci: configure dev deps prefix for dependabot

mhuster23:
The default has a different commit scope for dev dependencies, we want to have these unified.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -22,6 +22,8 @@ updates:
     directory: "/"
     schedule:
       interval: "daily"
+    commit-message:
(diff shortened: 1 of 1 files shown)


mhuster23:
> The [examples in the dependabot documentation](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#commit-message) do not include a trailing space. Therefore I suspect that dependabot would add that space on its own.

It says that it will add a colon, if there is no trailing whitespace. Some examples do have whitespace.

languitar (approved):

PR #450, opened 2024-01-23
Introduce delete repository endpoint

benedekh:
This MR proposes a /delete endpoint on the model server that could remove a repository. The endpoint is also available in the ModelClientV2. 

The feature was originally proposed by @nkoester in #237 (see its removal in [f97d2e8](https://github.com/modelix/modelix.core/pull/237/commits/f97d2e8b699089ec3d8e5d4dd50fa0e19e4b9225)).

### DIFF
--- a/api/model-server.yaml
+++ b/api/model-server.yaml
@@ -314,6 +314,18 @@ paths:
           $ref: '#/components/responses/500'
         "200":
           $ref: '#/components/responses/versionDelta'
+  /v2/repositories/{repository}/delete:
+    post:
+      operationId: deleteRepository
+      parameters:
+        - name: repository
+          in: "path"
+          required: true
+          schema:
+            type: string
+      responses:
+        "200":
+          $ref: '#/components/responses/200'
   /v2/repositories/{repository}/versions/{versionHash}:
     get:
       operationId: getRepositoryVersionHash
--- a/api/model-server.yaml
+++ b/api/model-server.yaml
@@ -324,8 +324,10 @@ paths:
           schema:
             type: string
       responses:
-        "200":
-          $ref: '#/components/responses/200'
+        "204":
+          $ref: '#/components/responses/204'
+        "404":
+          $ref: '#/components/responses/404'
   /v2/repositories/{repository}/versions/{versionHash}:
     get:
       operationId: getRepositoryVersionHash
--- a/model-server/src/main/kotlin/org/modelix/model/server/handlers/ModelReplicationServer.kt
+++ b/model-server/src/main/kotlin/org/modelix/model/server/handlers/ModelReplicationServer.kt
@@ -169,8 +169,13 @@ class ModelReplicationServer(val repositoriesManager: RepositoriesManager) {
             fun ApplicationCall.repositoryId() = RepositoryId(parameters["repository"]!!)
             fun PipelineContext<Unit, ApplicationCall>.repositoryId() = call.repositoryId()
 
-            repositoriesManager.removeRepository(repositoryId())
-            call.respond(HttpStatusCode.OK)
+            val repositoryId = repositoryId()
+            if (!repositoriesManager.repositoryExists(repositoryId)) {
+                call.respond(HttpStatusCode.NotFound)
+            } else {
+                repositoriesManager.removeRepository(repositoryId)
+                call.respond(HttpStatusCode.NoContent)
+            }
         }
 
         post<Paths.postRepositoryBranch> {
--- a/model-server/src/main/kotlin/org/modelix/model/server/handlers/ModelReplicationServer.kt
+++ b/model-server/src/main/kotlin/org/modelix/model/server/handlers/ModelReplicationServer.kt
@@ -170,11 +170,11 @@ class ModelReplicationServer(val repositoriesManager: RepositoriesManager) {
             fun PipelineContext<Unit, ApplicationCall>.repositoryId() = call.repositoryId()
 
             val repositoryId = repositoryId()
-            if (!repositoriesManager.repositoryExists(repositoryId)) {
-                call.respond(HttpStatusCode.NotFound)
-            } else {
-                repositoriesManager.removeRepository(repositoryId)
+            val foundAndDeleted = repositoriesManager.removeRepository(repositoryId)
+            if (foundAndDeleted) {
                 call.respond(HttpStatusCode.NoContent)
+            } else {
+                call.respond(HttpStatusCode.NotFound)
             }
         }
 
--- a/model-server/src/main/kotlin/org/modelix/model/server/handlers/ModelReplicationServer.kt
+++ b/model-server/src/main/kotlin/org/modelix/model/server/handlers/ModelReplicationServer.kt
@@ -165,6 +165,14 @@ class ModelReplicationServer(val repositoriesManager: RepositoriesManager) {
             call.respondDelta(initialVersion.getContentHash(), null)
         }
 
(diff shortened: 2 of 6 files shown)


model-server/src/main/kotlin/org/modelix/model/server/handlers/ModelReplicationServer.kt
```diff
@@ -165,6 +165,14 @@ class ModelReplicationServer(val repositoriesManager: RepositoriesManager) {
             call.respondDelta(initialVersion.getContentHash(), null)
         }
 
+        post<Paths.deleteRepository> {
+            fun ApplicationCall.repositoryId() = RepositoryId(parameters["repository"]!!)
+            fun PipelineContext<Unit, ApplicationCall>.repositoryId() = call.repositoryId()
+
+            repositoriesManager.removeRepository(repositoryId())
+            call.respond(HttpStatusCode.OK)
```
languitar, line 173: HTTP No Content would probably be the more appropriate response code to use here.

api/model-server.yaml
```diff
@@ -314,6 +314,18 @@ paths:
...
+  /v2/repositories/{repository}/delete:
+    post:
+      operationId: deleteRepository
+      parameters:
+        - name: repository
+          in: "path"
+          required: true
+          schema:
+            type: string
+      responses:
```
languitar, line 326: Which response is given when a repository is to be deleted that does not exist? Should be specified here.

model-server/src/main/kotlin/org/modelix/model/server/handlers/ModelReplicationServer.kt
benedekh: fixed in [81d8615](https://github.com/modelix/modelix.core/pull/450/commits/81d8615efaf13c61e4af2ee6dc3a83ec979293a5)

languitar (changes requested):
From the spec that looks fine to me now. But a test for the new endpoint would be good.

benedekh:
> From the spec that looks fine to me now. But a test for the new endpoint would be good.

I've implemented two tests for ModelClientV2 in [da30f52](https://github.com/modelix/modelix.core/pull/450/commits/da30f5213807ba03ed49c564dff0c2262b371647), which test the deleteRepository endpoint. Should I create separate tests that directly call the corresponding handler method on the server side or is it enough? (These tests implicitly test that method as well, since they are integration tests.)

model-server/src/test/kotlin/org/modelix/model/server/ModelClientV2Test.kt
```diff
@@ -154,4 +155,30 @@ class ModelClientV2Test {
...
+
+    @Test
+    fun `non-existing repository cannot be removed`() = runTest {
+        val url = "http://localhost/v2"
+        val client = ModelClientV2.builder().url(url).client(client).build().also { it.init() }
+        val repositoryId = RepositoryId(UUID.randomUUID().toString())
+        client.initRepository(repositoryId)
+
+        client.deleteRepository(repositoryId)
+        // repository should not exist here
```
languitar, line 179: Can't this be dropped. If a truly random repo name was never created as a repo it should most likely not exist.

languitar:
> > From the spec that looks fine to me now. But a test for the new endpoint would be good.
> 
> I've implemented two tests for ModelClientV2 in [da30f52](https://github.com/modelix/modelix.core/pull/450/commits/da30f5213807ba03ed49c564dff0c2262b371647), which test the deleteRepository endpoint. Should I create separate tests that directly call the corresponding handler method on the server side or is it enough? (These tests implicitly test that method as well, since they are integration tests.)

I think the indirect test through the client was the usual pattern right now. Maybe someone else has a different opinion on this?

model-server/src/test/kotlin/org/modelix/model/server/ModelClientV2Test.kt
benedekh: fixed in [c212413](https://github.com/modelix/modelix.core/pull/450/commits/c212413f2b2c28a560af40327abf88e0abdd42e0)

model-server/src/test/kotlin/org/modelix/model/server/ModelClientV2Test.kt
```diff
@@ -154,4 +155,29 @@ class ModelClientV2Test {
...
+        assertEquals(true, success)
+    }
+
+    @Test
+    fun `non-existing repository cannot be removed`() = runTest {
+        val url = "http://localhost/v2"
+        val client = ModelClientV2.builder().url(url).client(client).build().also { it.init() }
+        val repositoryId = RepositoryId(UUID.randomUUID().toString())
+        client.initRepository(repositoryId)
+        client.deleteRepository(repositoryId)
```
languitar, line 177:
```suggestion
```

That's what I meant. The repository shouldn't exist if it was never created.

model-server/src/main/kotlin/org/modelix/model/server/handlers/ModelReplicationServer.kt
```diff
@@ -165,6 +165,19 @@ class ModelReplicationServer(val repositoriesManager: RepositoriesManager) {
...
+            fun ApplicationCall.repositoryId() = RepositoryId(parameters["repository"]!!)
+            fun PipelineContext<Unit, ApplicationCall>.repositoryId() = call.repositoryId()
+
+            val repositoryId = repositoryId()
+            if (!repositoriesManager.repositoryExists(repositoryId)) {
+                call.respond(HttpStatusCode.NotFound)
+            } else {
+                repositoriesManager.removeRepository(repositoryId)
+                call.respond(HttpStatusCode.NoContent)
+            }
```
languitar, line 178: One more thing I didn't notice before: This is probably not deterministic. Between the test for `repositoryExists` and the `removeRepositoryCall`, some other call might have already removed the repo. This implementation here would need to ensure that the 404 response is also sent in that case.

benedekh: I moved those operations into the transaction to make them atomic. Fixed in [dc41ece](https://github.com/modelix/modelix.core/pull/450/commits/dc41ece986a8ee4fc698762d99dc5163ec832336).

languitar (approved):

languitar:
Looks fine to me now. I've pinged Michael and Sascha for a second review in case I missed something.

mhuster23 (approved):

model-server/src/test/kotlin/org/modelix/model/server/ModelClientV2Test.kt
```diff
@@ -154,4 +155,27 @@ class ModelClientV2Test {
...
+    @Test
+    fun `newly created repository can be removed`() = runTest {
+        val url = "http://localhost/v2"
+        val client = ModelClientV2.builder().url(url).client(client).build().also { it.init() }
+        val repositoryId = RepositoryId(UUID.randomUUID().toString())
+        client.initRepository(repositoryId)
+
+        val success = client.deleteRepository(repositoryId)
+
+        assertEquals(true, success)
```
slisson, line 168: I would also check if it's removed from the result of `client.listRepositories()`.

benedekh: Fixed in [f7bd1a0](https://github.com/modelix/modelix.core/pull/459/commits/f7bd1a0bf6390602ee5f7759a2c35fbd2f4e4625), see #459.

PR #452, opened 2024-01-24
MODELIX-720 Add ktor Micrometer monitoring endpoint

nkoester:
Add minimal metrics endpoint to the model-server


api/model-server.yaml
```diff
@@ -539,6 +539,12 @@ paths:
           $ref: '#/components/responses/200'
         "404":
           $ref: '#/components/responses/404'
+  /metrics:
+    get:
+      operationId: getMetrics
+      responses:
+        "200":
+          $ref: '#/components/responses/200'
```
languitar, line 547: Not sure if we should include that in the spec here. This is a different concern then our core content API.

model-server/src/main/kotlin/org/modelix/model/server/handlers/MetricsHandler.kt
```diff
@@ -0,0 +1,82 @@
...
+
+    val appMicrometerRegistry = PrometheusMeterRegistry(PrometheusConfig.DEFAULT)
+
+    fun init(application: Application) {
+        application.apply {
+            install(MicrometerMetrics) {
+                registry = appMicrometerRegistry
+                distributionStatisticConfig = DistributionStatisticConfig.Builder()
+                    .percentilesHistogram(true)
+                    .maximumExpectedValue(Duration.ofSeconds(20).toNanos().toDouble())
```
languitar, line 51: That maximum most likely isn't true in our use case. With streaming support, requests can take pretty long.

api/model-server.yaml
nkoester: There is no real structure whatsoever. I would decide this when dealing with https://issues.modelix.org/issue/MODELIX-683

languitar: If we don't add the route now, we don't have a breaking change later on ;-) The information here is in any case close to nothing.

nkoester: So you vote to not use the openAPI for the metric endpoint? I would not want to do that. If not: What is your preference here?

model-server/src/main/kotlin/org/modelix/model/server/handlers/MetricsHandler.kt
```diff
@@ -0,0 +1,82 @@
...
+                    .serviceLevelObjectives(
+                        Duration.ofMillis(100).toNanos().toDouble(),
+                        Duration.ofMillis(500).toNanos().toDouble(),
+                    )
```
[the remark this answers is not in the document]
nkoester: They are based of the ktor documentation over here. 

https://ktor.io/docs/micrometer-metrics.html#distribution_statistics

I do not have better SLO estimates.

model-server/src/main/kotlin/org/modelix/model/server/handlers/MetricsHandler.kt
languitar: I suspect histogram makes most sense for endpoints that are not streaming or downloading large files. For those, 20 is actually already pretty long. So we could say we search for the expected/desired response time on those endpoints and just ignore the long-running one. But I'd add a comment making that decision clear.

api/model-server.yaml
languitar: I would say the whole metrics stuff is a separate concern. And we don't even have the format under our control. I'd see that as orthogonal, maybe with an additional spec if we really want to. In the end, the whole metrics reporting format is standardized by someone else and we can just refer to that spec.

model-server/src/main/kotlin/org/modelix/model/server/handlers/MetricsHandler.kt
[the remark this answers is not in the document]
languitar: Maybe we can just drop that for the moment as long as we don't think about SLAs and SLOs?

api/model-server.yaml
nkoester: I will add a dedicated yaml for this for now. We can use that as a basis for MODELIX-683 and decide then how to proceed further.

languitar (commented):
I'm not sure if this project actually makes sense. If we have in the future multiple independent components with different APIs, we would tie them together with this project more than probably desired / needed.

api/build.gradle.kts
languitar, line 1: I'm not sure if this project actually makes sense. If we have in the future multiple independent components with different APIs, we would tie them together with this project more than probably desired / needed. Moreover, different users of the API definition can have different generator configuration needs.

nkoester: They simply depend on the project. No need to implement the not needed parts. If a supbproject needs individual generator configurations, it can adapt them manually as before. I think this subproject makes sense.

nkoester: I reverted the change back to its original state

languitar (approved):

PR #454, opened 2024-01-25
chore(deps): use the build(deps) prefix for all NPM dev updates

languitar:
The settings didn't configure the prefix for development dependencies and the settings were not applied to all NPM locations declared in the dependabot config.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -24,11 +24,15 @@ updates:
       interval: "daily"
     commit-message:
       prefix: "build(deps)"
+      prefix-development: "build(deps)"
   - package-ecosystem: "npm"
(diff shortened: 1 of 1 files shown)


mhuster23 (approved):
As far as I understood the documentation, `prefix-development` only needs to be specified, if it differs from `prefix`. The changes in line 33-35 are probably enough, but it doesn't hurt to specify `prefix-development` explicitly.

languitar:
Ah, what a weird configuration logic. NPM has a default for a separate dev prefix, but setting the general one to something non-default also erases the other default... :upside_down_face:  In any case, to avoid future confusion with these weird rules, being explicit here probably doesn't hurt.

PR #480, opened 2024-02-05
MODELIX-730 buildMonoQuery + callQuery does work on the model-server if buildMonoQuery is called outside of running query operation

mhuster23:


model-api-gen-gradle-test/kotlin-generation/src/test/kotlin/org/modelix/modelql/typed/TypedModelQLTest.kt
```diff
@@ -360,4 +361,18 @@ abstract class TypedModelQLTest {
 
         assertTrue { actual.instanceOf(C_XmlDocument) }
     }
+
+    @Test
+    fun `buildQuery and callQuery can be separated`() = runTest { client ->
```
languitar, line 366:
```suggestion
    fun `buildQuery can be used outside of an ongoing query call`() = runTest { client ->
```

Maybe that's a bit more specific than "separate" (in time? in source code location?...)

languitar (commented):
I'm not sure if I am happy with this solution. So far, we can construct queries ahead of time without knowledge of the actual client. So it's easy to separate reusable query fragments into isolated source code files. With this change, we would have to change all existing queries to pass in the client. Moreover, for recursive queries, there's usually a pair of "assign buildMonoQuery result" to variable and "callQuery from inside query function". As this is usually a top-level variable, initializing it based on a client is a lot harder than it is right now.

languitar:
> An alternative solution would be to turn it into a free function in `modelql-untyped` and hardcode the flag to `true` since it's true for all ModelQLNodes. That would mean there is no dependency on the client, but users would need to be careful to use the `buildMonoQuery` function from `modelql-untyped` and not `modelql-core`. Does that sound better?

But then I can also use the deprecated workaround that you have proposed once and put that in my own wrapper function where I can make the decision for or against role IDs actually a configuration property. I'm not sure if there's much benefit for now in this approach instead of really letting the server decide on what to do.

mhuster23:
We will use the workaround until we fully migrate to using roleIds everywhere.

PR #482, opened 2024-02-06
MODELIX-480: Migrate sync relevant features to kotlin lib

benedekh:
PR will be moved to https://github.com/modelix/modelix.mps-plugins/pull/66

# Related YouTrack epics and issues

- https://issues.modelix.org/issue/MODELIX-658
- https://issues.modelix.org/issue/MODELIX-480
   - https://issues.modelix.org/issue/MODELIX-637
   - https://issues.modelix.org/issue/MODELIX-628
   - https://issues.modelix.org/issue/MODELIX-635
   - https://issues.modelix.org/issue/MODELIX-634
   - https://issues.modelix.org/issue/MODELIX-629
   - https://issues.modelix.org/issue/MODELIX-625
   - https://issues.modelix.org/issue/MODELIX-805

PR #487, opened 2024-02-06
Enable concept lookup in the MPSLanguageRepository

benedekh:
Extracts changes made to the MPSLanguageRepository class in https://github.com/modelix/modelix.core/pull/482.

### DIFF
--- a/mps-model-adapters/src/main/kotlin/org/modelix/model/mpsadapters/MPSLanguageRepository.kt
+++ b/mps-model-adapters/src/main/kotlin/org/modelix/model/mpsadapters/MPSLanguageRepository.kt
@@ -28,9 +28,9 @@ import org.modelix.model.api.ILanguageRepository
 
 data class MPSLanguageRepository(private val repository: SRepository) : ILanguageRepository {
 
-    fun getConcept(uid: String): SAbstractConcept? = (resolveConcept(uid) as? MPSConcept)?.concept
+    fun resolveMPSConcept(uid: String): SAbstractConcept? = resolveConcept(uid)?.concept
 
-    override fun resolveConcept(uid: String): IConcept? {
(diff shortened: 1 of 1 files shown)


mps-model-adapters/src/main/kotlin/org/modelix/model/mpsadapters/MPSLanguageRepository.kt
```diff
@@ -21,11 +21,15 @@ import jetbrains.mps.smodel.adapter.structure.MetaAdapterFactory
...
 import jetbrains.mps.smodel.runtime.illegal.IllegalConceptDescriptor
+import org.jetbrains.mps.openapi.language.SAbstractConcept
 import org.jetbrains.mps.openapi.module.SRepository
 import org.modelix.model.api.IConcept
 import org.modelix.model.api.ILanguageRepository
 
-class MPSLanguageRepository(private val repository: SRepository) : ILanguageRepository {
+data class MPSLanguageRepository(private val repository: SRepository) : ILanguageRepository {
+
+    fun getConcept(uid: String): SAbstractConcept? = (resolveConcept(uid) as? MPSConcept)?.concept
```
slisson, line 31: If we change the return type of `MPSLanguageRepository.resolveConcept` to `MPSConcept?` then `getConcept` can be simplified to `resolveConcept(uid)?.concept`. An additional function is then probably not needed anymore. At least a  more distinctive name (`resolveMPSConcept`?) would be useful.

benedekh: fixed in [c6daac3](https://github.com/modelix/modelix.core/pull/487/commits/c6daac3191ab2b85a596b3725f6e4e1e265c47ac)

slisson (approved):

PR #488, opened 2024-02-06
MODELIX-718 Store Git metadata when syncing

mhuster23:
This PR adds the option for users to set arbitrary custom properties for the rootNode during the sync.
For example, users could store git metadata this way.

### DIFF
--- a/bulk-model-sync-gradle-test/ci.sh
+++ b/bulk-model-sync-gradle-test/ci.sh
@@ -3,7 +3,8 @@
 set -e
 set -x
 
-cd "$(dirname "$0")"
+TEST_DIR="$(dirname "$(readlink -f "$0")")"
+cd "${TEST_DIR}"
 
 (
   cd graph-lang-api
@@ -20,11 +21,11 @@ if [ "${CI}" != "true" ]; then
   }
 fi
 
-cd ..
+cd "${TEST_DIR}/.."
 ./gradlew :model-server:run --console=plain --args="-inmemory -port 28309" &
 MODEL_SERVER_PID=$!
 
-cd "$(dirname "$0")"
+cd "${TEST_DIR}"
 
 curl -X GET --retry 30 --retry-connrefused --retry-delay 1 http://localhost:28309/health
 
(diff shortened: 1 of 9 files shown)


bulk-model-sync-gradle/src/main/kotlin/org/modelix/model/sync/bulk/gradle/tasks/ImportIntoModelServer.kt
```diff
@@ -91,6 +96,11 @@ abstract class ImportIntoModelServer @Inject constructor(of: ObjectFactory) : De
                     logger.info("Got root node: {}", rootNode)
                     logger.info("Calculating diff...")
                     ModelImporter(rootNode, continueOnError.get()).importFilesAsRootChildren(files)
+
+                    logger.info("Setting meta properties...")
+                    for ((key, value) in metaProperties.get()) {
+                        rootNode.setPropertyValue(IProperty.fromName(key), value)
+                    }
```
languitar, line 103: Are there any reserved properties already on the root node the plugin implementation should shield against from blindly accepting them?

bulk-model-sync-gradle-test/ci.sh
languitar, line 1: What's the essence of the changes here? Pretty hard to understand. And why could the test run before?

languitar (dismissed):
Generally fine.

It's a bit hard to understand the changes to the test CI bash script and the reasons for the performed refactorings to the tests. A few more explanations in the commit messages would make these changes easier to understand and to reason about.

bulk-model-sync-gradle/src/main/kotlin/org/modelix/model/sync/bulk/gradle/tasks/ImportIntoModelServer.kt
mhuster23: As far as I know, the root node doesn't contain any properties by default.

bulk-model-sync-gradle-test/ci.sh
mhuster23: It depended on how you invoked the script. The old version used relative paths. The new one uses absolute paths.

mhuster23:
Reworded some commits to improve clarity.

languitar (approved):

PR #489, opened 2024-02-07
fix(mps-model-adapters): do not fail when moving children in model

odzhychko:
This is the specific fix with rudimentary testing.

I am working on setting up proper integration tests.
To achieve that, I am doing some exploration, and it will take longer than this fix.
See https://github.com/modelix/modelix.core/compare/main...MODELIX-742-integration-tests

I assume that `moveChild` might also cause issues in other implementations of `IDefaultNodeAdapter`.
Also, `addNewChild` seems broken and could affect the sync.
This also something I will test.

### DIFF
--- a/bulk-model-sync-lib/src/commonMain/kotlin/org/modelix/model/sync/bulk/ModelImporter.kt
+++ b/bulk-model-sync-lib/src/commonMain/kotlin/org/modelix/model/sync/bulk/ModelImporter.kt
@@ -22,8 +22,10 @@ import org.modelix.model.api.INode
 import org.modelix.model.api.INodeReference
 import org.modelix.model.api.INodeResolutionScope
 import org.modelix.model.api.SerializedNodeReference
+import org.modelix.model.api.getConcept
 import org.modelix.model.api.getDescendants
 import org.modelix.model.api.remove
+import org.modelix.model.api.tryResolveChildLink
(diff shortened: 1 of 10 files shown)


slisson:
IChildLink should specify whether the nodes in that role are ordered or not. The sync algorithm then can use this information to avoid unnecessary move operations, which is also useful in cases where the INode implementation would be able to move the node. The moveChild method should never just ignore the call.

odzhychko:
> IChildLink should specify whether the nodes in that role are ordered or not. The sync algorithm then can use this information to avoid unnecessary move operations, which is also useful in cases where the INode implementation would be able to move the node. The moveChild method should never just ignore the call.

Thanks for the review.

By ignoring the moves, my change would have [caused removing wrong nodes](https://github.com/modelix/modelix.core/blob/2ccd7b3ec63ce2597f056b313b19d6522ce82b26/bulk-model-sync-lib/src/commonMain/kotlin/org/modelix/model/sync/bulk/ModelImporter.kt#L169).

Also your thought helped me discover   [SContainmentLinkAdapter.isUnordered](https://github.com/JetBrains/MPS/blob/master/core/openapi/source/org/jetbrains/mps/openapi/language/SContainmentLink.java#L37)

odzhychko:
I will reopen a new PR, when I am ready.

PR #493, opened 2024-02-07
chore: openapi restructuring proposal

languitar:


model-server-openapi/specifications/model-server.yaml
```diff
@@ -13,481 +13,700 @@ info:
...
-  - url: '/'
+  - url: '/v2'
     description: model-server
+
+tags:
+  - name: IDs
+  - name: repository management
+  - name: branch management
+  - name: versions
+    description: TODO what is this?
```
slisson, line 24: The `.../branches/...` endpoints operate on the latest version on a branch, which is fine if that is what you want, but sometimes you may want to be more specific.

model-server-openapi/specifications/model-server.yaml
```diff
@@ -13,481 +13,700 @@ info:
...
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/RepositoryList"
+        default:
+          $ref: "#/components/responses/Problem"
     post:
-      operationId: postGenerateClientId
+      operationId: initializeRepository
+      summary: Creates a new repository
```
slisson, line 88: For me it's not intuitive that this endpoint creates a new repository. Also specifying the repository name in the JSON body while all other endpoint specify it in the URL seems inconsistent. I would prefer a `POST /repositories/{repository}` endpoint.

model-server-openapi/specifications/model-server.yaml
```diff
@@ -13,481 +13,700 @@ info:
...
+      tags:
+        - repository management
+      parameters:
+        - name: useRoleIds
```
slisson, line 92: Shouldn't this be part of the JSON body?

benedekh:
**Remarks:**

- (blocking) We have to create follow-up issues to refactor the implementation behind the API. E.g. the `deleteRepository` endpoint used to accept POST requests, but now it shall accept DELETE requests.
- (non-blocking) We have to provide examples to each API endpoint or to none at all. Somewhere in-between could be annyoing from the user's perspective.
- (blocking) What happened to those API endpoints in `model-server.yaml` that were not behind the `/v2` URL? E.g., `/headers`, `/get/{key}`, `/getAll`, `/put/{key}`, `/putAll`,

languitar:
> * [blocking] What happened to those API endpoints in `model-server.yaml` that were not behind the `/v2` URL? E.g., `/headers`, `/get/{key}`, `/getAll`, `/put/{key}`, `/putAll`.

We removed them because they circumvent per-repository access control.

model-server-openapi/specifications/model-server.yaml
languitar: > POST requests are idiomatically used to create single resources on a collection resource endpoint,

https://opensource.zalando.com/restful-api-guidelines/#post

model-server-openapi/specifications/model-server.yaml
languitar: I think we just forgot removing this TODO. That was already clear to us

model-server-openapi/specifications/model-server.yaml
```diff
@@ -13,481 +13,700 @@ info:
...
-        "500":
-          $ref: '#/components/responses/500'
         "200":
-          $ref: '#/components/responses/versionDelta'
-  /v2/repositories/{repository}/branches/{branch}/pollHash:
+          $ref: '#/components/responses/200json'
+        default:
+          $ref: '#/components/responses/Problem'
+
+  /repositories/{repository}/versions/{versionHash}/delta:
```
mhuster23, line 378: I'm not sure yet, how I feel about including the `repository` path parameter in version related routes.
It makes sense structurally, but our versionHashes are unique across repositories, so given a `versionHash` one would first have to determine which `repository` it belongs to, to be able to query it.
This somewhat feels like a step backwards.

model-server-openapi/specifications/model-server-operative.yaml
```diff
@@ -15,18 +15,56 @@ info:
...
           $ref: '#/components/responses/200'
 
+  /health:
+    get:
+      operationId: getHealth
+      responses:
+        "200":
+          description: healthy
+          content:
+            text/plain:
```
odzhychko, line 46: We should make the responses (healthy and not healthy) an object to make it extensible.
Extensions to the health endpoint could be detailed statuses of used services.

https://opensource.zalando.com/restful-api-guidelines/#110

odzhychko: Examples from Spring Boot are: https://docs.spring.io/spring-boot/docs/1.5.14.RELEASE/reference/html/production-ready-monitoring.html#production-ready-health-access-restrictions

model-server-openapi/specifications/model-server-operative.yaml
```diff
@@ -15,18 +15,56 @@ info:
...
+              schema:
+                type: string
+                enum: ["healthy"]
+        "500":
```
odzhychko, line 50: We could make it a 503 to be more expressive.

model-server-openapi/specifications/model-server-operative.yaml
languitar: We actually haven't touched these much. There's also a ktor extension for realizing these endpoints which would simply bring a reasonable format.

model-server-openapi/specifications/model-server.yaml
```diff
@@ -13,481 +13,700 @@ info:
...
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/BranchList"
+        "404":
+          $ref: "#/components/responses/NoSuchRepository"
+        default:
+          $ref: "#/components/responses/Problem"
+
+    post:
```
odzhychko, line 173: `POST /repositories/{repository}/branches`, `PUT /repositories/{repository}/branches/{branch}` should additionally be able to create a branch from another branch name without knowing the version.

model-server-openapi/specifications/model-server.yaml
```diff
@@ -13,481 +13,700 @@ info:
...
+      summary: Updates an existing branch
+      description: Can be used to let an existing branch point to a new hash.
+      tags:
+        - branch management
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: "#/components/schemas/Branch"
```
odzhychko, line 241: `PUT /repositories/{repository}/branches/{branch}`: The request should not include `name`, only `current_hash`.

odzhychko: I see, you thought about it by responing `400 Branch name cannot be changed`.
Is it common for APIs to always take and return the same object even, if the data can not be changed?

languitar: The typical pattern would be to read and post/put the same type of resource. We can mark some of the properties as readOnly to avoid making them mutable but that doesn't account for all possible conditions.

model-server-openapi/specifications/model-server.yaml
```diff
@@ -13,481 +13,700 @@ info:
...
-            type: string
-        - name: branch
-          in: "path"
-          required: true
-          schema:
-            type: string
+      operationId: queryRepositoryVersionHash
+      tags:
+        - ModelQl access
+        - versions
```
odzhychko, line 413: `/repositories/{repository}/versions/{versionHash}/query` does not need `versions`. `ModelQL` is enough.

model-server-openapi/specifications/model-server.yaml
```diff
@@ -13,481 +13,700 @@ info:
...
+    parameters:
+      - $ref: "#/components/parameters/RepositoryPath"
+      - $ref: "#/components/parameters/VersionHashPath"
     get:
-      operationId: pollRepositoryBranchHash
+      operationId: getRepositoryVersionHash
+      description: |-
+        If `from` is not specified, this **returns the full version data**.
+      tags:
+        - versions
```
odzhychko, line 387: `/repositories/{repository}/versions/{versionHash}/delta` does not need `versions` tag. `delta access` is enough. I think the `versions` tag is not needed to cluster endpionts that operate on a version, because those endpoints operate very differntly on the versions. (e.g. delta-access, model-access, Model-QL access)

languitar: That tag was intended to collect everything that operates on a single version. Doesn't that make sense?

model-server-openapi/specifications/model-server.yaml
languitar: See above

model-server-openapi/specifications/model-server.yaml
odzhychko: It is odd because `version` is the only tag that tags endpoints from different mechanisms.

model-server-openapi/specifications/model-server.yaml
```diff
@@ -13,481 +13,700 @@ info:
...
       responses:
         "200":
           $ref: '#/components/responses/200'
+        default:
+          $ref: "#/components/responses/Problem"
+
   /getRecursively/{key}:
     get:
+      tags:
+        - TODO remove this?
```
odzhychko, line 709: This would still fit with `delta access`.

model-server-openapi/specifications/model-server.yaml
```diff
@@ -497,8 +716,15 @@ paths:
...
           $ref: '#/components/responses/200json'
+        default:
+          $ref: "#/components/responses/Problem"
+
   /poll/{key}:
     get:
+      tags:
+        - TODO remove this?
+      description: |-
+        TODO: could not find any user in our code
```
odzhychko, line 727: This is used in ModelClientV2 as part of the `delta access` for long polling.

model-server-openapi/specifications/model-server.yaml
```diff
@@ -13,481 +13,700 @@ info:
   description: "modelix public API"
   version: "1.0.0"
 servers:
-  - url: '/'
+  - url: '/v2'
     description: model-server
+
+tags:
```
odzhychko, line 19: Users showed interest in a whole new king of API that can be used to read the version history.
e.g. Who made when a change. What is the hash of the change. What are previous versions? How to a normalized linear history look? How does the history graph looks?

model-server-openapi/specifications/model-server.yaml
```diff
@@ -13,481 +13,700 @@ info:
...
+        default:
+          $ref: "#/components/responses/Problem"
+
+  # TODO what is this?
+  /counter/{key}:
+    post:
+      tags:
+        - TODO remove this?
+      description: |-
+        TODO only used by RestWebModelClient.
```
odzhychko, line 667: This was used to get accessending IDs for clientId RestWebModelClient.

model-server-openapi/specifications/model-server-operative.yaml
nkoester: https://github.com/zensum/ktor-health-check

I think we talked about this extension in this context

model-server-openapi/specifications/model-server-operative.yaml
```diff
@@ -15,18 +15,56 @@ info:
   version: "1.0.0"
 
 servers:
-  - url: '/'
+  - url: '/meta'
```
odzhychko, line 18: `/meta` is very ambiguous in our context. Maybe would should use something linke `administration`/`introspection`/`control`/`server`.

model-server-openapi/specifications/model-server.yaml
```diff
@@ -13,481 +13,700 @@ info:
   description: "modelix public API"
   version: "1.0.0"
 servers:
-  - url: '/'
+  - url: '/v2'
```
odzhychko, line 16: Do we want version throug an URL?
How to we evolve the API?
Do we want to use MediaType-Versioning?
https://opensource.zalando.com/restful-api-guidelines/#114

languitar:
This will be split into smaller PRs now that we have decided on a process to handle the API.

PR #502, opened 2024-02-09
feat(model-server): Add healthcheck to model-server image

### LINKED ISSUE #501: Add a health check to the model-server image
Currently we have to specify the health_check in each compose file manually. If the model-server image was making use of the [HEALTHCHECK](https://docs.docker.com/engine/reference/builder/#healthcheck) we could leave that responsibility to modelix.

drik98:
Closes #501

### DIFF
--- a/model-server/Dockerfile
+++ b/model-server/Dockerfile
@@ -1,6 +1,7 @@
 FROM registry.access.redhat.com/ubi8/openjdk-11:1.18-2.1705602259
 WORKDIR /usr/modelix-model
 EXPOSE 28101
+HEALTHCHECK CMD curl --fail http://localhost:28101/health || exit 1
 COPY run-model-server.sh /usr/modelix-model/
 COPY build/dependency-libs/ /usr/modelix-model/model-server/build/libs/
 COPY build/libs/ /usr/modelix-model/model-server/build/libs/


drik98:
As not configured differently the defaults will be 

interval - DURATION (default: 30s)
timeout - DURATION (default: 30s)
start-period - DURATION (default: 0s)
retries - DURATION (default: 3)

which seems sensible

odzhychko (dismissed):
HEALTHCHECK seems helpful.

Even if HEALTHCHECK does additional checks at start up, I cannot think of a case where this might break a realistic uses case of the image.

odzhychko:
@drik98 Your commit is not signed.

> The base branch requires all commits to be signed. [Learn more about signing commits.](https://docs.github.com/articles/about-gpg/)

slisson (changes requested):
There is a `/health` endpoint that should be used. See https://github.com/modelix/modelix.kubernetes/blob/1f62139d76482d0428a86c9fa167b7f38b441fec/helm/modelix/templates/common/model-deployment.yaml#L45-L58

drik98:
> There is a `/health` endpoint that should be used. See https://github.com/modelix/modelix.kubernetes/blob/1f62139d76482d0428a86c9fa167b7f38b441fec/helm/modelix/templates/common/model-deployment.yaml#L45-L58

I meant to use that but forgot in the implementation. Good catch.

model-server/Dockerfile
```diff
@@ -1,6 +1,7 @@
 FROM registry.access.redhat.com/ubi8/openjdk-11:1.18-2.1705602259
 WORKDIR /usr/modelix-model
 EXPOSE 28101
+HEALTHCHECK CMD curl --fail http://localhost:28101 || exit 1
```
languitar, line 4: We have a dedicated health endpoint that can be used here. Cf. https://github.com/modelix/modelix.core/blob/aa16a542008905c49cc465a535a2557aead9369d/docs/global/modules/core/pages/howto/usage-model-server.adoc?plain=1#L70

```suggestion
HEALTHCHECK CMD curl --fail http://localhost:28101/health
```

What was the reason to add `|| exit 1`? Curl would fails with a non-zero exit code in any case.

languitar (changes requested):

model-server/Dockerfile
languitar: Oops, now the remaining comments pointing to that loaded :see_no_evil:

drik98: > What was the reason to add || exit 1? Curl would fails with a non-zero exit code in any case.

The reason is that curl may exit with any code https://timi.eu/docs/anatella/5_1_9_1_list_of_curl_error_codes.html but the health check should exit with either 0, 1 or 2. But it should not even be exited with 2 :D 


image

odzhychko (changes requested):
Please change your three commits into one.

odzhychko (approved):

model-server/Dockerfile
languitar: Ah right, didn't know that Docker healthchecks are that picky.

languitar (approved):
Thanks for the PR. Looks fine to me.

slisson (approved):

PR #451, opened 2024-01-23
MODELIX-719 Performance of IModelClient2.pull

slisson:
All objects of the model were loaded individually from the database. These many small requests resulted in a poor performance when running a bulk sync on an existing large model. Batch requests are now used to reduce the overhead.

PR #503, opened 2024-02-12
Add an option to eagerly warm up the Ignite cache on model-server start

languitar:
Adds a command line option for controlling whether the ignite cache is populated eagerly on model-server start. When enabled, the model-server only becomes responsive once the cache is populated from the database.

This option is implemented as populating the cache this way is a lot faster than pulling a branch at the moment. For instance, in one project, populating the cache using the loadCache method takes only a about 20 seconds whereas pulling the data using the REST API as a means of warming up the cache takes about 13 minutes.


odzhychko:
Is this PR needed after #451 was merged?

languitar:
No, not needed anymore.

languitar:
I'll reopen this temporarily to build an image with the pull fix from #519 but without the query performance optimizations

PR #504, opened 2024-02-12
Introduce ordered property for child links

odzhychko:
Child links that are unordered are handled specially when importing using the bulk-model-sync. 
 Such unordered children are not being reorder in when imported.

### DIFF
--- a/bulk-model-sync-lib/src/commonMain/kotlin/org/modelix/model/sync/bulk/ModelImporter.kt
+++ b/bulk-model-sync-lib/src/commonMain/kotlin/org/modelix/model/sync/bulk/ModelImporter.kt
@@ -154,7 +154,12 @@ class ModelImporter(private val root: INode, private val continueOnError: Boolea
                         originalIdToExisting[expectedId] = newChild
                         newChild
                     } else {
+                        // The existing child node is not only moved to a new index,
+                        // it is potentially moved to a new parent and role.
                         node.moveChild(role, index, existingNode)
+                        // If the old parent and old role synchronized before the move operation,
+                        // the existing child node would have been marked as to be deleted.
+                        // Now that it is used, it should not be deleted.
                         nodesToRemove.remove(existingNode)
                         existingNode
                     }
@@ -166,6 +171,9 @@ class ModelImporter(private val root: INode, private val continueOnError: Boolea
                 syncNode(childNode, expected, progressReporter)
             }
 
+            // At this point, all n expected children for this role are created and correctly sorted.
+            // This means the first n children are correct.
+            // Any child beyond that is an unexpected child has to be marked for removal.
             nodesToRemove += node.getChildren(role).drop(expectedNodes.size)
         }
     }
--- a/bulk-model-sync-lib/src/commonMain/kotlin/org/modelix/model/sync/bulk/ModelImporter.kt
+++ b/bulk-model-sync-lib/src/commonMain/kotlin/org/modelix/model/sync/bulk/ModelImporter.kt
@@ -23,6 +23,7 @@ import org.modelix.model.api.INodeReference
 import org.modelix.model.api.INodeResolutionScope
 import org.modelix.model.api.SerializedNodeReference
 import org.modelix.model.api.getDescendants
+import org.modelix.model.api.isChildRoleOrdered
 import org.modelix.model.api.remove
 import org.modelix.model.data.ModelData
 import org.modelix.model.data.NodeData
@@ -118,15 +119,15 @@ class ModelImporter(private val root: INode, private val continueOnError: Boolea
         }
     }
 
-    private fun syncChildren(node: INode, data: NodeData, progressReporter: ProgressReporter) {
-        val allRoles = (data.children.map { it.role } + node.allChildren.map { it.roleInParent }).distinct()
+    private fun syncChildren(existingParent: INode, expectedParent: NodeData, progressReporter: ProgressReporter) {
+        val allRoles = (expectedParent.children.map { it.role } + existingParent.allChildren.map { it.roleInParent }).distinct()
         for (role in allRoles) {
-            val expectedNodes = data.children.filter { it.role == role }
-            val existingNodes = node.getChildren(role).toList()
+            val expectedNodes = expectedParent.children.filter { it.role == role }
+            val existingNodes = existingParent.getChildren(role).toList()
 
             // optimization that uses the bulk operation .addNewChildren
             if (existingNodes.isEmpty() && expectedNodes.all { originalIdToExisting[it.originalId()] == null }) {
-                node.addNewChildren(role, -1, expectedNodes.map { it.concept?.let { ConceptReference(it) } }).zip(expectedNodes).forEach { (newChild, expected) ->
+                existingParent.addNewChildren(role, -1, expectedNodes.map { it.concept?.let { ConceptReference(it) } }).zip(expectedNodes).forEach { (newChild, expected) ->
                     val expectedId = checkNotNull(expected.originalId()) { "Specified node '$expected' has no id" }
                     newChild.setPropertyValue(NodeData.idPropertyKey, expectedId)
                     originalIdToExisting[expectedId] = newChild
@@ -142,21 +143,38 @@ class ModelImporter(private val root: INode, private val continueOnError: Boolea
                 continue
             }
 
-            expectedNodes.forEachIndexed { index, expected ->
-                val nodeAtIndex = node.getChildren(role).toList().getOrNull(index)
+            val isOrdered = existingParent.isChildRoleOrdered(role)
+
+            expectedNodes.forEachIndexed { indexInImport, expected ->
+                val existingChildren = existingParent.getChildren(role).toList()
                 val expectedId = checkNotNull(expected.originalId()) { "Specified node '$expected' has no id" }
+                // newIndex is the index on which to import the expected child.
+                // It might be -1 if the child does not exist and should be added at the end.
+                val newIndex = if (isOrdered) {
+                    indexInImport
+                } else {
+                    // The `existingChildren` are only searched once for the expected element before changing.
+                    // Therefore, indexing existing children will not be more efficient than iterating once.
+                    // (For the moment, this is fine because as we expect unordered children to be the exception,
+                    // Reusable indexing would be possible if we switch from
+                    // a depth-first import to a breadth-first import.)
+                    existingChildren
+                        .indexOfFirst { existingChild -> existingChild.originalId() == expected.originalId() }
+                }
+                // existingChildren.getOrNull handles `-1` as needed by returning `null`.
+                val nodeAtIndex = existingChildren.getOrNull(newIndex)
                 val expectedConcept = expected.concept?.let { s -> ConceptReference(s) }
                 val childNode = if (nodeAtIndex?.originalId() != expectedId) {
                     val existingNode = originalIdToExisting[expectedId]
                     if (existingNode == null) {
-                        val newChild = node.addNewChild(role, index, expectedConcept)
+                        val newChild = existingParent.addNewChild(role, newIndex, expectedConcept)
                         newChild.setPropertyValue(NodeData.idPropertyKey, expectedId)
                         originalIdToExisting[expectedId] = newChild
                         newChild
                     } else {
                         // The existing child node is not only moved to a new index,
                         // it is potentially moved to a new parent and role.
-                        node.moveChild(role, index, existingNode)
+                        existingParent.moveChild(role, newIndex, existingNode)
                         // If the old parent and old role synchronized before the move operation,
                         // the existing child node would have been marked as to be deleted.
                         // Now that it is used, it should not be deleted.
@@ -171,10 +189,10 @@ class ModelImporter(private val root: INode, private val continueOnError: Boolea
                 syncNode(childNode, expected, progressReporter)
             }
 
-            // At this point, all n expected children for this role are created and correctly sorted.
-            // This means the first n children are correct.
-            // Any child beyond that is an unexpected child has to be marked for removal.
-            nodesToRemove += node.getChildren(role).drop(expectedNodes.size)
+            val expectedNodesIds = expectedNodes.map(NodeData::originalId).toSet()
+            // Do not use existingNodes, but call node.getChildren(role) because
+            // the recursive synchronization in the meantime already removed some nodes from node.getChildren(role).
+            nodesToRemove += existingParent.getChildren(role).filterNot { existingNode -> expectedNodesIds.contains(existingNode.originalId()) }
         }
     }
 
--- a/bulk-model-sync-lib/src/jvmTest/kotlin/org/modelix/model/sync/bulk/ModelImporterOrderPropertyTest.kt
+++ b/bulk-model-sync-lib/src/jvmTest/kotlin/org/modelix/model/sync/bulk/ModelImporterOrderPropertyTest.kt
@@ -1,0 +1,366 @@
+/*
+ * Copyright (c) 2023.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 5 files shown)


bulk-model-sync-lib/src/commonMain/kotlin/org/modelix/model/sync/bulk/ModelImporter.kt
```diff
@@ -121,52 +122,127 @@ class ModelImporter(private val root: INode, private val continueOnError: Boolea
...
+        }
+    }
 
-            // optimization that uses the bulk operation .addNewChildren
-            if (existingNodes.isEmpty() && expectedNodes.all { originalIdToExisting[it.originalId()] == null }) {
-                node.addNewChildren(role, -1, expectedNodes.map { it.concept?.let { ConceptReference(it) } }).zip(expectedNodes).forEach { (newChild, expected) ->
+    private fun syncChildRole(
+        data: NodeData,
+        role: String?,
+        node: INode,
```
languitar, line 132: Would this be easier to read of that variables was called `parent` or `host` or something alike?

model-api/src/commonMain/kotlin/org/modelix/model/api/INode.kt
```diff
@@ -380,6 +380,20 @@ fun INode.tryResolveProperty(role: String): IProperty? {
         ?: allLinks.find { it.getSimpleName() == role }
         ?: allLinks.find { it.getUID() == role }
 }
+
+/**
+ * Resolves whether the child link is ordered or not.
+ */
+fun INode.isChildRoleOrdered(role: String?): Boolean {
+    // Assume children to be ordered by default.
+    // Unordered children are the special case.
```
languitar, line 389: Maybe that could be lifted to the kdoc. That's information potentially interesting for users of this method.

model-api/src/commonMain/kotlin/org/modelix/model/api/IChildLink.kt
```diff
@@ -30,6 +30,15 @@ interface IChildLink : ILink {
     companion object {
         fun fromName(name: String): IChildLink = ChildLinkFromName(name)
     }
+
+    /**
+     * Whether children with this role can be ordered.
+     *
+     * If a child role is not ordered implementations of [[INode.moveChild]] are allowed to fail
```
languitar, line 37:
```suggestion
     * If a child role is not ordered, implementations of [[INode.moveChild]] are allowed to fail
```

model-api/src/commonMain/kotlin/org/modelix/model/api/IChildLink.kt
```diff
@@ -30,6 +30,15 @@ interface IChildLink : ILink {
     companion object {
         fun fromName(name: String): IChildLink = ChildLinkFromName(name)
     }
+
+    /**
+     * Whether children with this role can be ordered.
```
languitar, line 35: Shouldn't this be more in the direction of whether ordering is significant?

It is probably also good to mention potential consequences if this false. for instance that multiple requests to children of a role are not guaranteed to return elements in the same order.

bulk-model-sync-lib/src/commonMain/kotlin/org/modelix/model/sync/bulk/ModelImporter.kt
```diff
@@ -121,52 +122,127 @@ class ModelImporter(private val root: INode, private val continueOnError: Boolea
...
+        progressReporter: ProgressReporter,
+    ) {
+        val expectedNodes = data.children.filter { it.role == role }
```
languitar, line 135: The code here is somewhat performance critical. Now that I see this (already existing) code fragment: The list of children is iterated once to compute all roles and then again for each role. I suspect there is a more efficient version that builds up a map first to avoid the repeated full iteration through all children.

bulk-model-sync-lib/src/commonMain/kotlin/org/modelix/model/sync/bulk/ModelImporter.kt
```diff
@@ -121,52 +122,127 @@ class ModelImporter(private val root: INode, private val continueOnError: Boolea
...
+            syncNode(childNode, expected, progressReporter)
+        }
+
+        // At this point, all n expected children for this role are created and correctly sorted.
+        // This means the first n children are correct.
+        // Any child beyond that is an unexpected child has to be marked for removal.
+        nodesToRemove += node.getChildren(role).drop(expectedNodes.size)
+    }
+
+    private fun syncChildRoleUnordered(
```
languitar, line 208: This method has a large overlap with syncChildRoleOrdered. Wouldn't it make more sense to only factor out the index computation into two distinct paths. Right now, property assignment etc is duplicated. That means we have to remember to fix both paths in case there is a bug etc.

bulk-model-sync-lib/src/jvmTest/kotlin/org/modelix/model/sync/bulk/ModelImporterOrderPropertyTest.kt
```diff
@@ -0,0 +1,508 @@
...
+
+        pBranch.runWrite {
+            modelData.load(pBranch)
+        }
+
+        return pBranch.withAutoTransactions()
+    }
+
+    @Test
+    fun `unordered children are not reorder in import`() {
```
languitar, line 97:
```suggestion
    fun `unordered children are not reorder`() {
```

I think the import context is already clear from the test class and that way I don't have to parse the somewhat complex grammar of the test method names ;-)

bulk-model-sync-lib/src/jvmTest/kotlin/org/modelix/model/sync/bulk/ModelImporterOrderPropertyTest.kt
```diff
@@ -0,0 +1,508 @@
...
+
+        val children = branch.getRootNode().allChildren.first().allChildren.toList()
+        // child1 and child2 should not be reordered because they have an unordered role
+        // child3 is removed
+        // child4 is added at the end
+        assertEquals(listOf("child1", "child2", "child4"), children.map(INode::originalId))
+    }
+
+    @Test
+    fun `unordered children with changing parents in import are appended`() {
```
languitar, line 175:
```suggestion
    fun `unordered children with changing parents are appended`() {
```

bulk-model-sync-lib/src/jvmTest/kotlin/org/modelix/model/sync/bulk/ModelImporterOrderPropertyTest.kt
```diff
@@ -0,0 +1,508 @@
...
+
+        val children = branch.getRootNode().allChildren.first().allChildren.toList()
+        // child1 moved with because of the role change from "uid2" to "uid3".
+        assertEquals(listOf("child2", "child1"), children.map(INode::originalId))
+        // child1 changes its role from "uid2" to "uid3".
+        assertEquals("uid3", children[1].getContainmentLink()!!.getUID())
+    }
+
+    @Test
+    fun `unordered children with changing parent in import are appended`() {
```
languitar, line 241:
```suggestion
    fun `unordered children with changing parent are appended`() {
```

bulk-model-sync-lib/src/jvmTest/kotlin/org/modelix/model/sync/bulk/ModelImporterOrderPropertyTest.kt
```diff
@@ -0,0 +1,508 @@
...
+        val dataToImport = ModelData.fromJson(dataToImportJson)
+        modelImporter.import(dataToImport)
+
+        val children = branch.getRootNode().allChildren.first().allChildren.toList()
+        // child1 is added after existing child2 because it was moved from another node
+        assertEquals(listOf("child2", "child1"), children.map(INode::originalId))
+    }
+
+    @Test
+    fun `unordered child can not change concept`() {
```
languitar, line 311:
```suggestion
    fun `an unordered child cannot change concept`() {
```

Why is this actually the case?

bulk-model-sync-lib/src/jvmTest/kotlin/org/modelix/model/sync/bulk/ModelImporterOrderPropertyTest.kt
```diff
@@ -0,0 +1,508 @@
...
+        val dataToImport = ModelData.fromJson(dataToImportJson)
+        modelImporter.import(dataToImport)
+
+        val children = branch.getRootNode().allChildren.toList()
+        // child1 and child 2 are reordered because they have no role.
+        assertEquals(listOf("child2", "child1"), children.map(INode::originalId))
+    }
+
+    @Test
+    fun `unordered children are synced recursively`() {
```
languitar, line 457: Do we also need this test case for ordered children?

languitar (changes requested):
Some initial thoughts after a first round on the draft PR

bulk-model-sync-lib/src/commonMain/kotlin/org/modelix/model/sync/bulk/ModelImporter.kt
odzhychko: That is an interesting thought.
I can probably reuse the old algorithm as is, if I manipulate the provided expected order of children to match the order of the existing children for roles that have no significant order.
I will give it a try and see if this works.
(Luckily, I have automated tests to validate such a refactoring 😁)

bulk-model-sync-lib/src/commonMain/kotlin/org/modelix/model/sync/bulk/ModelImporter.kt
odzhychko: This method does not exist anymore.
But I changed it for `fun syncChildren(existingParent: INode, expectedParent: NodeData, ...)

model-api/src/commonMain/kotlin/org/modelix/model/api/INode.kt
odzhychko: Done. Also I extended the explanation further.

bulk-model-sync-lib/src/commonMain/kotlin/org/modelix/model/sync/bulk/ModelImporter.kt
odzhychko: For now this is out of scope.
The needed change is more complicated, because `expectedParent.children` changes might change in between iterations.

model-api/src/commonMain/kotlin/org/modelix/model/api/IChildLink.kt
odzhychko: Done. I extended KDoc to include how this impacts requesting children subsequently.

bulk-model-sync-lib/src/commonMain/kotlin/org/modelix/model/sync/bulk/ModelImporter.kt
odzhychko: Done. I found a solution that has less duplication, but is less straightforward.

bulk-model-sync-lib/src/jvmTest/kotlin/org/modelix/model/sync/bulk/ModelImporterOrderPropertyTest.kt
odzhychko: For the ordered children the test already existed in `ModelImporterTest.kt`.
As I now use the same code path as for ordered children, I removed this seperate test.

bulk-model-sync-lib/src/jvmTest/kotlin/org/modelix/model/sync/bulk/ModelImporterOrderPropertyTest.kt
odzhychko: This is the case because it is generally not intended for nodes in MPS and our `INode`s to change concepts.
So in the importer, we cannot change the concept.
If we are requested to but cannot change the concept, we fail not to produce garbage data.
(Though this check is not done not properly in all existing path, but this is a change for some other PR.) 

---

I removed this test becaus it is not specific to ordering anymore.

languitar (dismissed):

odzhychko:
@slisson This change is an extension of the very basic `IChildLink`. Even if this PR is checked and approved by someone else, I will wait for your feedback/veto.

bulk-model-sync-lib/src/jvmTest/kotlin/org/modelix/model/sync/bulk/ModelImporterOrderPropertyTest.kt
```diff
@@ -0,0 +1,355 @@
...
+        }
+        """.trimIndent()
+        val dataToImport = ModelData.fromJson(dataToImportJson)
+        modelImporter.import(dataToImport)
+
+        val children = branch.getRootNode().allChildren.first().allChildren.toList()
+        // child1 and child2 should not be reordered because they have an unordered role
+        // child3 is removed
+        // child4 is added at the end
+        assertEquals(listOf("child1", "child2", "child4"), children.map(INode::originalId))
```
slisson, line 169: If `IChildLink.isOrdered` is false then we can't expect any specific order. `assertEquals(setOf("child1", "child2", "child4"), children.map(INode::originalId).toSet())` would be correct.

bulk-model-sync-lib/src/jvmTest/kotlin/org/modelix/model/sync/bulk/ModelImporterOrderPropertyTest.kt
```diff
@@ -0,0 +1,355 @@
...
+            ]
+          }
+        }
+        """.trimIndent()
+        val dataToImport = ModelData.fromJson(dataToImportJson)
+        modelImporter.import(dataToImport)
+
+        val children = branch.getRootNode().allChildren.first().allChildren.toList()
+        // child1 moved with because of the role change from "uid2" to "uid3".
+        assertEquals(listOf("child2", "child1"), children.map(INode::originalId))
```
slisson, line 233: It's not valid to assume that `.allChildren` returns nodes in different roles are in any specific order. These tests probably shouldn't use `.allChildren` at all.

bulk-model-sync-lib/src/jvmTest/kotlin/org/modelix/model/sync/bulk/ModelImporterOrderPropertyTest.kt
```diff
@@ -0,0 +1,355 @@
...
+            ]
+          }
+        }
+        """.trimIndent()
+        val dataToImport = ModelData.fromJson(dataToImportJson)
+        modelImporter.import(dataToImport)
+
+        val children = branch.getRootNode().allChildren.first().allChildren.toList()
+        // child1 is added after existing child2 because it was moved from another node
+        assertEquals(listOf("child2", "child1"), children.map(INode::originalId))
```
slisson, line 305: Same here. If `IChildLink.isOrders == false` then we can't expect a specific order.

bulk-model-sync-lib/src/jvmTest/kotlin/org/modelix/model/sync/bulk/ModelImporterOrderPropertyTest.kt
odzhychko: Done. (`.allChildren` is now only used to navigate from the root to the relevant parents.)

bulk-model-sync-lib/src/jvmTest/kotlin/org/modelix/model/sync/bulk/ModelImporterOrderPropertyTest.kt
odzhychko: Done

languitar (approved):

PR #505, opened 2024-02-12
ci: allow publishing SNAPSHOT release via workflow_dispatch

languitar:
This allows pushing out PR states for testing under a snapshot release identifier.

This commit migrates the publish GitHub workflow from using custom bash scripts for publishing towards the commonly provided GitHub actions. That way, the decision when to publish which type of release is completely contained inside the publishing workflow. Without this change, the workflow and the scripts would often need simultaneous changes for reflecting desired publishing results.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -4,6 +4,7 @@ on:
   push:
     tags:
       - '[0-9]+.[0-9]+.[0-9]+*'
+  # Allows building SNAPSHOT releases with the commit SHA inlcuded for testing purposes
   workflow_dispatch:
   # Test this workflow in PRs in case it changed
   pull_request:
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yml
```diff
@@ -30,16 +31,31 @@ jobs:
...
+          echo "VERSION=${version}" >> $GITHUB_OUTPUT
+      - name: Determine Docker metadata
+        id: meta
+        uses: docker/metadata-action@v5
+        with:
+          images: |
+            modelix/model-server
+            modelix/modelix-model
+          tags: |
+            type=raw,value=${{ steps.version.outputs.VERSION }},enable=true
```
odzhychko, line 56: > /usr/bin/docker buildx build --file ./model-server/Dockerfile --iidfile /home/runner/work/_temp/docker-actions-toolkit-0IDqvj/iidfile --label org.opencontainers.image.created=2024-02-12T13:13:22.169Z --label org.opencontainers.image.description=Mostly model related components --label org.opencontainers.image.licenses=Apache-2.0 --label org.opencontainers.image.revision=869a63a6c43458e09a9e7edefe30ddfce6ac3ef7 --label org.opencontainers.image.source=https://github.com/modelix/modelix.core --label org.opencontainers.image.title=modelix.core --label org.opencontainers.image.url=https://github.com/modelix/modelix.core --label org.opencontainers.image.version=0.0.1-869a63a6c43458e09a9e7edefe30ddfce6ac3ef7-SNAPSHOT --platform linux/amd64,linux/arm64 --provenance mode=max,builder-id=https://github.com/modelix/modelix.core/actions/runs/7872295160 --tag modelix/model-server:0.0.1-869a63a6c43458e09a9e7edefe30ddfce6ac3ef7-SNAPSHOT --tag modelix/modelix-model:0.0.1-869a63a6c43458e09a9e7edefe30ddfce6ac3ef7-SNAPSHOT --metadata-file /home/runner/work/_temp/docker-actions-toolkit-0IDqvj/metadata-file ./model-server

> modelix/modelix-model:0.0.1-869a63a6c43458e09a9e7edefe30ddfce6ac3ef7-SNAPSHOT

Should we use a shorter commit hash?

On a release tag we should also publish the tag `latest` like in the script.

languitar: I can shorten that

languitar: Now we would get `modelix/model-server:0.0.1-bf016f5-SNAPSHOT`

odzhychko: That looks better.

What about the `latest` tag for releases?

languitar: Yes, missed that.

languitar: updated to include that, too

odzhychko: Now the SNAPSHOT builds are also published as `latest`. 
I think we need to distinguish proper release and SNAPSHOT releases.

odzhychko (approved):

.github/workflows/publish.yml
languitar: I think we should have another look on this once the next release is due. Might be that something's still of by then.

PR #511, opened 2024-02-14
Fix/properties and children in builtin languages

odzhychko:


model-api/src/commonTest/kotlin/org/modelix/model/api/BuiltinLanguagesTest.kt
```diff
@@ -0,0 +1,30 @@
...
+
+package org.modelix.model.api
+
+import kotlin.test.Test
+import kotlin.test.assertEquals
+
+class BuiltinLanguagesTest {
+
+    @Test
+    fun propertiesAreAdded() {
```
languitar, line 25: Maybe something like "allPropertiesAreAvailableWithoutPriorAccess"?

languitar (dismissed):
For the two newly added tests, an explanatory comment for why this seemingly simple condition is being tested could help

model-api/src/commonTest/kotlin/org/modelix/model/api/BuiltinLanguagesTest.kt
odzhychko: Done. Also added more detailed comments in the test code.

languitar:
Is it intended that this PR now includes a few of @mhuster23 's commits from his draft PR?

odzhychko:
> Is it intended that this PR now includes a few of @mhuster23 's commits from his draft PR?

No that was an accident. I already removed them.

languitar (approved):

PR #517, opened 2024-02-15
MODELIX-752 Improve model-server query performance by caching the model in memory

slisson:
The versioning feature of our model data structure has the cost of a slower performance. To achieve a comparable performance to the light-model-server the whole model is now loaded into memory using a more efficient map implementation.

The model is update incrementally whenever the branch changes.

In the future we may be able to improve the performance by introducing indices and parallelizing the query execution (using IBulkQuery) and then might not need this in-memory model anymore.

### DIFF
--- a/model-datastructure/src/commonMain/kotlin/org/modelix/model/lazy/NonCachingObjectStore.kt
+++ b/model-datastructure/src/commonMain/kotlin/org/modelix/model/lazy/NonCachingObjectStore.kt
@@ -1,0 +1,43 @@
+/*
+ * Copyright (c) 2023.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.modelix.model.lazy
+
+import org.modelix.model.IKeyValueStore
+
+class NonCachingObjectStore(override val keyValueStore: IKeyValueStore) : IDeserializingKeyValueStore {
+
+    override fun <T> getAll(hashes_: Iterable<String>, deserializer: (String, String) -> T): Iterable<T> {
+        val hashes = hashes_.toList()
+        val serialized: Map<String, String?> = keyValueStore.getAll(hashes_)
+        return hashes.map { hash ->
+            val value = checkNotNull(serialized[hash]) { "Entry not found: $hash" }
+            deserializer(hash, value)
+        }
+    }
+
+    override fun <T> get(hash: String, deserializer: (String) -> T): T? {
+        return keyValueStore.get(hash)?.let(deserializer)
+    }
+
+    override fun put(hash: String, deserialized: Any, serialized: String) {
+        keyValueStore.put(hash, serialized)
+    }
+
+    override fun prefetch(hash: String) {
+        keyValueStore.prefetch(hash)
+    }
+}
--- a/model-datastructure/src/jvmMain/kotlin/org/modelix/model/InMemoryModel.kt
+++ b/model-datastructure/src/jvmMain/kotlin/org/modelix/model/InMemoryModel.kt
@@ -1,0 +1,293 @@
+/*
+ * Copyright (c) 2024.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.modelix.model
+
+import gnu.trove.map.TLongObjectMap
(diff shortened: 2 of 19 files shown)


model-server-openapi/specifications/model-server.yaml
```diff
@@ -291,6 +291,25 @@ paths:
           $ref: '#/components/responses/200json'
         "500":
           $ref: '#/components/responses/500'
+  /v2/repositories/{repository}/branches/{branch}/query/warmup:
+    get:
+      operationId: postRepositoryBranchQueryWarmup
```
languitar, line 296: post in the operation ID for a get route does not make much sense

model-server/src/main/kotlin/org/modelix/model/server/handlers/ModelReplicationServer.kt
```diff
@@ -290,6 +301,15 @@ class ModelReplicationServer(val repositoriesManager: RepositoriesManager) {
...
         }
 
+        get<Paths.postRepositoryBranchQueryWarmup> { parameters ->
+            val branchRef = RepositoryId(parameters.repository).getBranchReference(parameters.branch)
+            val version = repositoriesManager.getVersion(branchRef)
+            withContext(Dispatchers.IO) {
+                inMemoryModel.getModel(version!!.getTree())
+            }
+            call.respondText("Version $version loaded into memory")
+        }
```
languitar, line 311: Depending on how long this might take, especially with a potentially IO-limited database from the cloud, we might get into issues with exceeding the default 30 seconds timeout again.

slisson: The `inMemoryModel.getModel` call continues even if the client disconnects, because there are no coroutine usages inside of it that would allow the dispatcher to interrupt it. This warmup endpoint can be used in a readiness probe and eventually it will succeed without a delay.

model-server-openapi/specifications/model-server.yaml
slisson: Changed to `getRepositoryBranchQueryWarmup`.

model-server/src/main/kotlin/org/modelix/model/server/handlers/ModelReplicationServer.kt
languitar: Even if that's not a problem for the effect of this operation, having to deal with timeouts is tedious. What about splitting this into two operations. One that starts loading and one that can be used to query if loading finished?

slisson: The endpoint is intended to be used by a readiness probe. I don't understand how splitting it into two endpoint would make that easier to use.

model-datastructure/src/jvmMain/kotlin/org/modelix/model/InMemoryModel.kt
```diff
@@ -0,0 +1,293 @@
...
+private val LOG = mu.KotlinLogging.logger { }
+
+class IncrementalInMemoryModel {
+    private var lastModel: InMemoryModel? = null
+
+    @Synchronized
+    fun getModel(tree: CLTree): InMemoryModel {
+        val reusable = lastModel?.takeIf { it.branchId == tree.getId() }
+        val newModel = if (reusable == null) {
+            InMemoryModel.load(tree)
```
odzhychko, line 54: Isn't the `tree.getId()` actually the (internal) repository ID?

Does this mean, that we evict and reload the `lastModel` continuously, when the server servers two repositories by ModelQL?

(Still better than before, but I want to understand the limitations.)

model-server/src/main/kotlin/org/modelix/model/server/handlers/ModelReplicationServer.kt
languitar: We already have a health endpoint. In k8s, we can only define a single http endpoint for probes and I would also want the health endpoint be usable in addition to fetching the model into memory. So maybe then this should use a similar pattern for specifying checks to execute via query parameters?

model-datastructure/src/jvmMain/kotlin/org/modelix/model/InMemoryModel.kt
slisson: I didn't want to have an unlimited amount of models in memory, but I changed it.

languitar:
This needs a rebase now.

model-datastructure/src/commonMain/kotlin/org/modelix/model/lazy/NonCachingObjectStore.kt
```diff
@@ -0,0 +1,40 @@
...
+ */
+
+package org.modelix.model.lazy
+
+import org.modelix.model.IKeyValueStore
+
+class NonCachingObjectStore(override val keyValueStore: IKeyValueStore) : IDeserializingKeyValueStore {
+
+    override fun <T> getAll(hashes_: Iterable<String>, deserializer: (String, String) -> T): Iterable<T> {
+        val hashes = hashes_.toList()
```
odzhychko, line 24: Can we iterate over the result map?

```kt
    override fun <T> getAll(hashes: Iterable<String>, deserializer: (String, String) -> T): Iterable<T> {
        val serialized: Map<String, String?> = keyValueStore.getAll(hashes)
        return serialized.mapNotNull { (hash, value) -> value?.let { deserializer(hash, it) } }
```

The cast to `as T` would be bad, if  any requested `hash` is not in `serialized.values` or if any of `serialized.keys` is `null`.

odzhychko (commented):
Aside from the [this one question](https://github.com/modelix/modelix.core/pull/517#discussion_r1495636141) every think else looks fine to me.

model-datastructure/src/commonMain/kotlin/org/modelix/model/lazy/NonCachingObjectStore.kt
slisson: The result is expected to be in the same order as the `hashes`. That `IDeserializingKeyValueStore` API is poorly designed and may be removed completely in the future.

I added a not null check.

odzhychko (approved):

PR #519, opened 2024-02-15
fix(model-server): respond with all objects if no base version was given

odzhychko:
In an optimization, only some of the referenced objects were emitted.

Making this this optimization correct emits more objects.

Also for not seending duplicate objects we have to keep a set of send object hashes while streaming objects.

### DIFF
--- a/model-server/src/main/kotlin/org/modelix/model/server/handlers/RepositoriesManager.kt
+++ b/model-server/src/main/kotlin/org/modelix/model/server/handlers/RepositoriesManager.kt
@@ -37,6 +37,7 @@ import org.modelix.model.lazy.KVEntryReference
 import org.modelix.model.lazy.RepositoryId
 import org.modelix.model.lazy.computeDelta
 import org.modelix.model.metameta.MetaModelBranch
+import org.modelix.model.persistent.CPVersion
 import org.modelix.model.server.store.IStoreClient
 import org.modelix.model.server.store.LocalModelClient
 import org.modelix.model.server.store.pollEntry
@@ -236,16 +237,23 @@ class RepositoriesManager(val client: LocalModelClient) {
                 val version = CLVersion(versionHash, objectStore)
                 // Use a bulk query to make as few request to the underlying store as possible.
                 val bulkQuery = objectStore.newBulkQuery()
+                // It is unsatisfactory that we have to keep already emitted hashes in memory.
+                // But without changing the underlying model,
+                // we have to do this to not emit objects more than once.
+                val seenHashes = mutableSetOf<String>()
                 fun emitObjects(entry: KVEntryReference<*>) {
                     bulkQuery.get(entry).onSuccess {
                         channel.trySend(entry.getHash() to it!!.serialize())
-                        for (referencedEntry in it!!.getReferencedEntries()) {
-                            emitObjects(referencedEntry)
+                        for (referencedEntry in it.getReferencedEntries()) {
+                            val wasSeenBefore = !seenHashes.add(referencedEntry.getHash())
+                            // Do not emit the object if we already emitted it.
+                            if (!wasSeenBefore) {
+                                emitObjects(referencedEntry)
+                            }
                         }
                     }
                 }
-                channel.send(versionHash to kvStore.get(versionHash)!!)
-                emitObjects(version.treeHash!!)
+                emitObjects(KVEntryReference(versionHash, CPVersion.DESERIALIZER))
                 (bulkQuery as? BulkQuery)?.process()
             }
         }
--- a/model-server/src/test/kotlin/org/modelix/model/server/ModelClientV2Test.kt
+++ b/model-server/src/test/kotlin/org/modelix/model/server/ModelClientV2Test.kt
@@ -28,10 +28,12 @@ import org.modelix.model.api.IConceptReference
 import org.modelix.model.api.ITree
 import org.modelix.model.api.PBranch
 import org.modelix.model.client2.ModelClientV2
+import org.modelix.model.client2.runWrite
 import org.modelix.model.lazy.CLTree
 import org.modelix.model.lazy.CLVersion
 import org.modelix.model.lazy.RepositoryId
 import org.modelix.model.operations.OTBranch
+import org.modelix.model.persistent.IKVValue
 import org.modelix.model.server.handlers.ModelReplicationServer
 import org.modelix.model.server.store.InMemoryStoreClient
 import org.modelix.modelql.core.count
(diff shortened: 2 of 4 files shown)


model-server/src/main/kotlin/org/modelix/model/server/handlers/RepositoriesManager.kt
```diff
@@ -227,16 +228,22 @@ class RepositoriesManager(val client: LocalModelClient) {
         if (baseVersionHash == null) {
             // no need to cache anything if there is no delta computation happening
             return flow {
+                // It is unsatisfactory that we have to keep already emitted hashes in memory.
+                // But without changing the underlying model,
+                // we have to do this to not emit objects more the once.
```
languitar, line 233:
```suggestion
                // we have to do this to not emit objects more than once.
```

model-server/src/test/kotlin/org/modelix/model/server/handlers/ModelReplicationServerTest.kt
```diff
@@ -0,0 +1,98 @@
...
+        }
+
+        // Assert
+        val response = client.get {
+            url {
+                takeFrom(url)
+                appendPathSegments("repositories", repositoryId.id, "branches", branchId.branchName)
+            }
+            useVersionStreamFormat()
+        }
```
languitar, line 87: I think the grouping is off here. The actions that's being tested is pulling. Writing some test content to the store is still arranging?

model-server/src/test/kotlin/org/modelix/model/server/handlers/ModelReplicationServerTest.kt
```diff
@@ -0,0 +1,98 @@
...
+        val repositoryId = RepositoryId("repo1")
+        val branchId = repositoryId.getBranchReference("my-branch")
+
+        // Act
+        modelClient.runWrite(branchId) { root ->
+            root.addNewChild("aChild", -1, null as IConceptReference?)
+        }
+        modelClient.runWrite(branchId) { root ->
+            root.addNewChild("aChild", -1, null as IConceptReference?)
+        }
```
languitar, line 78: Why are two entries needed and in two separate transactions?

model-server/src/test/kotlin/org/modelix/model/server/handlers/ModelReplicationServerTest.kt
```diff
@@ -0,0 +1,98 @@
...
+                takeFrom(url)
+                appendPathSegments("repositories", repositoryId.id, "branches", branchId.branchName)
+            }
+            useVersionStreamFormat()
+        }
+        val versionDelta = response.readVersionDelta()
+        val seenHashes = mutableSetOf<String>()
+        versionDelta.objectsFlow!!.collect { (hash, _) ->
+            val wasNotSeenYet = seenHashes.add(hash)
+            if (!wasNotSeenYet) {
```
languitar, line 92:
```suggestion
            val wasSeenBefore = seenHashes.add(hash)!
            if (wasSeenBefore) {
```

Maybe this version of duplicate negations is a bit easier to read?

model-server/src/test/kotlin/org/modelix/model/server/handlers/ModelReplicationServerTest.kt
```diff
@@ -0,0 +1,98 @@
...
+                // We should not send the same object (with the same hash more than once)
+                fail("Hash $hash send more than once.")
```
languitar, line 94:
```suggestion
                fail("Hash $hash sent more than once.")
```

languitar (commented):
I can't judge everything here. I understand from the diff why duplicate sends are avoided but (maybe due to reordering) I don't see how the missing objects are fixed.

model-server/src/main/kotlin/org/modelix/model/server/handlers/RepositoriesManager.kt
odzhychko: Done

model-server/src/test/kotlin/org/modelix/model/server/handlers/ModelReplicationServerTest.kt
odzhychko: You are right. I changed it.

model-server/src/test/kotlin/org/modelix/model/server/handlers/ModelReplicationServerTest.kt
odzhychko: I added a comment in the code.

model-server/src/test/kotlin/org/modelix/model/server/handlers/ModelReplicationServerTest.kt
odzhychko: Done

model-server/src/test/kotlin/org/modelix/model/server/handlers/ModelReplicationServerTest.kt
odzhychko: Done

model-server/src/main/kotlin/org/modelix/model/server/handlers/RepositoriesManager.kt
```diff
@@ -236,16 +237,23 @@ class RepositoriesManager(val client: LocalModelClient) {
...
+                            // Do not emit the object, if we already emitted id.
+                            if (!wasSeenBefore) {
+                                emitObjects(referencedEntry)
+                            }
                         }
                     }
                 }
-                channel.send(versionHash to kvStore.get(versionHash)!!)
-                emitObjects(version.treeHash!!)
+                emitObjects(KVEntryReference(versionHash, CPVersion.DESERIALIZER))
```
odzhychko, line 256: @languitar This line is the main fix.

In the previous version, we only sent data for `treeHash`.
By using `KVEntryReference(versionHash, CPVersion.DESERIALIZER)` we send everything returned by `CLVersion.getReferencedEntries`.


```kt
    // CLVersion.getReferencedEntries
    override fun getReferencedEntries(): List<KVEntryReference<IKVValue>> {
        return listOfNotNull(
            treeHash,
            previousVersion,
            originalVersion,
            baseVersion,
            mergedVersion1,
            mergedVersion2,
            operationsHash,
        ) + (operations ?: arrayOf()).map { it.getReferencedEntries() }.flatten()
    }
```

model-server/src/main/kotlin/org/modelix/model/server/handlers/RepositoriesManager.kt
```diff
@@ -236,16 +237,23 @@ class RepositoriesManager(val client: LocalModelClient) {
...
+                // But without changing the underlying model,
+                // we have to do this to not emit objects more than once.
+                val seenHashes = mutableSetOf<String>()
                 fun emitObjects(entry: KVEntryReference<*>) {
                     bulkQuery.get(entry).onSuccess {
                         channel.trySend(entry.getHash() to it!!.serialize())
                         for (referencedEntry in it!!.getReferencedEntries()) {
-                            emitObjects(referencedEntry)
+                            val wasSeenBefore = !seenHashes.add(referencedEntry.getHash())
+                            // Do not emit the object, if we already emitted id.
```
slisson, line 249:
```suggestion
                            // Do not emit the object, if we already emitted it.
```

model-server/src/test/kotlin/org/modelix/model/server/ModelClientV2Test.kt
```diff
@@ -184,4 +185,28 @@ class ModelClientV2Test {
...
+
+        // Act
+        val versionPulled = modelClientForAssert.pullIfExists(branchId)!! as CLVersion
+
+        // Assert
+        val referencedEntriesInPulledVersion = versionPulled.data!!.getReferencedEntries()
+        for (referencedEntry in referencedEntriesInPulledVersion) {
+            // Check that the store also provides each referenced KVEntry.
+            // `getValue` would fail if this is not the case.
+            referencedEntry.getValue(versionPulled.store)
```
slisson, line 209: Maybe we should do the same check recursively on the object returned by `getValue`.

odzhychko: Thanks for the tip. Done.

languitar (approved):
Fine for me. Maybe @slisson could have a look as well as the last thread was his request.

slisson (approved):

PR #520, opened 2024-02-16
 fix(vue-model-api): do not declare development dependencies as library dependencies

odzhychko:


.pre-commit-config.yaml
```diff
@@ -43,9 +43,9 @@ repos:
...
       - id: eslint
         additional_dependencies:
-          - "eslint@8.5.0"
-          - "eslint-config-prettier@9.0.0"
-          - "@typescript-eslint/eslint-plugin@6.7.2"
-          - "@typescript-eslint/parser@6.7.2"
+          - "eslint@8.56.0"
+          - "eslint-config-prettier@9.1.0"
+          - "@typescript-eslint/eslint-plugin@7.0.1"
+          - "@typescript-eslint/parser@7.0.1"
```
languitar, line 49: How annoying that you can't just use what's declared in package.json.

languitar (approved):

.pre-commit-config.yaml
odzhychko: Yeah, I was thinking, if it would make more sens to call `npm run lint <CHANGED_FILES_ARG>` as from pre-commit instead of using the ESLint hook in the future.

languitar: Sounds reasonable.

PR #523, opened 2024-02-19
MODELIX-637: SyncQueue fixes

benedekh:
### Changes in this pull request

- extensive (manual) testing the SyncQueue in various scenarios:
  -  SyncTasks starting new SyncTasks
  - using the async execution semantics extensively by the tasks waiting for each other
  - consecutive tasks are cancelled if a predecessor throws an exception
  - simplifying lock acquisiton (MPS read/write, modelix read/write) in some cases. However, there might be some corner-cases still wherer MPS might get frozen. This is can be also due to some missing dependencies between the SyncTasks. But we'll debug and improve it further as we go on with the plugin.
  - Model/ModuleBinding.deactivate is more thread-safe due to explicit synchronizations
  - Collections that might be accessed from different threads are synchronized. E.g., inside the BindingsRegistry, MpsToModelixMap
  - ModelixTreeChangeVisitor is refactored: 
      - the event-handlers are moved to the Node/Model/ModuleTransformers where they semantically belong (depending on the type of change and the target of the change)
      - errors are logged explicitly, especially if we missed a case that should have been handled. E.g. a Modelix Node is neither an MPS Node/Model/Module.
  - new changes are handled in ModelixTreeChangeVisitor:
      - property change in Models/Module, 
      - deletion of ModelImports, Language/DevKit Dependencies, ModuleDependencies
      - model/module rename. However, module rename still does not work due to [MODELIX-726](https://issues.modelix.org/issue/MODELIX-726).
  - some other refactorings, bugfixes and TODO comments I forgot to mention

### DIFF
--- a/mps-sync-plugin-lib/src/main/kotlin/org/modelix/mps/sync/SyncService.kt
+++ b/mps-sync-plugin-lib/src/main/kotlin/org/modelix/mps/sync/SyncService.kt
@@ -8,6 +8,7 @@ import org.modelix.model.lazy.BranchReference
 import org.modelix.mps.sync.bindings.ModelBinding
 import org.modelix.mps.sync.bindings.ModuleBinding
 import java.net.URL
+import java.util.concurrent.CompletableFuture
 
 @UnstableModelixFeature(reason = "The new modelix MPS plugin is under construction", intendedFinalization = "2024.1")
 interface SyncService {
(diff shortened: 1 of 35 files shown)


KarstenDiekhoff:
The code looks good as far as I can tell. There are a lot of findings from the bot though, which we might want to take a look at.

As for a functional review, I am not quite sure how to create a setup to test these things. You mentioned already testing quite a bit. Maybe that is enough? Otherwise maybe someone else can run some tests, since I would just ask you and therefore run the same tests as you again.

PR #528, opened 2024-02-22
fix(model-server): handle empty MODELIX_SERVER_MODELQL_WARMUP_BRANCH

languitar:
Don't fail the health check in case the user did not define the new env var MODELIX_SERVER_MODELQL_WARMUP_BRANCH. Before this fix, the health checked returned 500 due to a NullPointerException.


languitar:
This needs clarifications whether the whole logic should only run when both variables are defined by the user or if we want to implicitly assume master if the branch is not specified in the environment.

odzhychko:
I guess, that `(!repositoriesManager.inMemoryModels.loadModelAsync(version!!.getTree()))` throws a NPE when the repository or branch does not exist.
If the repository does not exist we could either fail to startup or create it and start tracking it.

odzhychko:
However we specify the behavior to be, let us please add tests.

languitar:
Here's the issue for the underlying problem when a repo does not yet exist: https://issues.modelix.org/issue/MODELIX-775/model-server-cannot-be-launched-if-warmup-is-requested-for-a-to-be-created-repository

languitar:
> I guess, that `(!repositoriesManager.inMemoryModels.loadModelAsync(version!!.getTree()))` throws a NPE when the repository or branch does not exist. If the repository does not exist we could either fail to startup or create it and start tracking it.

Failing startup is not an ideal option as that implies users have to change model-server configurations for bootstrapping a new repo, which is cumbersome.

PR #532, opened 2024-02-26
fix(model-server): use role IDs in InMemoryNode

languitar:
Otherwise, properties of built-in languages cannot be looked up via ModelQl as the role IDs retranslated into the role name during query execution.


model-datastructure/src/jvmMain/kotlin/org/modelix/model/InMemoryModel.kt
```diff
@@ -253,6 +253,8 @@ class InMemoryModel private constructor(
 }
 
 data class InMemoryNode(val model: InMemoryModel, val nodeId: Long) : INode, INodeReference {
+    override fun usesRoleIds(): Boolean = true
```
slisson, line 256: The `InMemoryModel` should load the correct value from `CLTree.usesRoleIds`.

slisson (approved):

PR #533, opened 2024-02-26
fix(model-server): some objects were missing when pulling a version

slisson:
Channel.trySend fails when the buffer is full, but that failure was ignored. Using runBlocking together with Channel.send to wait for space in the buffer.

This was detected by https://github.com/modelix/modelix.mps-plugins/pull/38. The tests in modelix.mps-plugins succeed with this change.

It was broken since version 4.8.1 which includes PR https://github.com/modelix/modelix.core/pull/451. There was already a comment about it https://github.com/modelix/modelix.core/pull/451#issuecomment-1933696866.


languitar (changes requested):
Please add some explanations to the code how runBlocking together with send fixes a bug that's not visible from any kind of test. The fix is otherwise lost easily with the next refactoring in this code area. I suspect the condition is hard to unit-test and therefore there's no test included?

PR #534, opened 2024-02-26
fix(model-server): RepositoriesManager.computeDelta should wait for channel to have capacity

odzhychko:
Previously, sending objects failed quietly when the underlying channel was full. Now, the execution of the bulk query is blocked when the underlying channel is full.

### DIFF
--- a/model-server/src/main/kotlin/org/modelix/model/server/handlers/RepositoriesManager.kt
+++ b/model-server/src/main/kotlin/org/modelix/model/server/handlers/RepositoriesManager.kt
@@ -18,6 +18,7 @@ import kotlinx.coroutines.flow.asFlow
 import kotlinx.coroutines.flow.channelFlow
 import kotlinx.coroutines.flow.emptyFlow
 import kotlinx.coroutines.flow.map
+import kotlinx.coroutines.runBlocking
 import kotlinx.datetime.Clock
 import org.apache.commons.collections4.map.LRUMap
 import org.modelix.model.IKeyValueStore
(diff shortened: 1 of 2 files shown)


model-server/src/test/kotlin/org/modelix/model/server/handlers/RepositoriesManagerTest.kt
```diff
@@ -0,0 +1,78 @@
...
+        }
+        val latestVersionHash = repositoriesManager.getVersionHash(branchRef)!!
+
+        // Act / Assert
+        val flow = repositoriesManager.computeDelta(latestVersionHash, null)
+        // `count` is an example for consuming the flow.
+        // The flow should not fail when the consumer is slow.
+        // With the given data, the consumption is slower than the production of values.
+        // Still, the following code should run successfully.
+        val count = flow.count()
```
slisson, line 57: When I run the test without the fix it still succeeds. `.count()` is not sufficient.

slisson (changes requested):

model-server/src/test/kotlin/org/modelix/model/server/handlers/RepositoriesManagerTest.kt
odzhychko: Hm, ok.
I extended the existing `ModelClientV2Test.kt` instead.
It is one test less and tests less internal things.

slisson: The modified test fails with without the fix.

slisson (approved):

languitar:
Could this be the reason for error conditions such as: java.lang.RuntimeException: Entry vSaGx*t5TjgBGJSPlycXh5QWKQ7ZDCWb4hkHPd8cTPy0 not found

odzhychko:
> Could this be the reason for error conditions such as: java.lang.RuntimeException: Entry vSaGx*t5TjgBGJSPlycXh5QWKQ7ZDCWb4hkHPd8cTPy0 not found

Yes, this is the primary reason for such errors.

PR #553, opened 2024-03-01
MODELIX-789 bulk-model-sync from model-server to MPS results in many renamed files without content changes

mhuster23:
The renamings happen because the project used FilePerRootPersistence and MPS could not resolve the concept, as it was not loaded. As a fallback the filename would default to the id of the root node of a changed model.

However changes were made where none should have been. This PR fixes those cases.

### DIFF
--- a/model-api/src/commonMain/kotlin/org/modelix/model/api/INode.kt
+++ b/model-api/src/commonMain/kotlin/org/modelix/model/api/INode.kt
@@ -211,6 +211,7 @@ interface INode {
      * @return the original reference of this node
      */
     fun getOriginalReference(): String? = getPropertyValue(IProperty.fromName(NodeData.ID_PROPERTY_KEY))
+        ?: getPropertyValue(IProperty.fromName("#mpsNodeID#")) // for backwards compatibility
 
     // <editor-fold desc="non-string based API">
     fun usesRoleIds(): Boolean = false
--- a/model-api/src/commonMain/kotlin/org/modelix/model/api/INode.kt
+++ b/model-api/src/commonMain/kotlin/org/modelix/model/api/INode.kt
@@ -22,6 +22,7 @@ import kotlinx.coroutines.flow.flatMapConcat
 import kotlinx.coroutines.flow.flattenConcat
 import kotlinx.coroutines.flow.flowOf
 import org.modelix.model.area.IArea
+import org.modelix.model.data.NodeData
 
 /**
  * Representation of a model element.
@@ -205,6 +206,11 @@ interface INode {
      */
     @Deprecated("use getReferenceLinks()")
     fun getReferenceRoles(): List<String>
+
+    /**
+     * @return the original reference of this node
+     */
+    fun getOriginalReference(): String? = getPropertyValue(IProperty.fromName(NodeData.ID_PROPERTY_KEY))
 
     // <editor-fold desc="non-string based API">
     fun usesRoleIds(): Boolean = false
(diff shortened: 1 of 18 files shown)


mps-model-adapters/src/main/kotlin/org/modelix/model/mpsadapters/MPSNode.kt
```diff
@@ -160,7 +160,7 @@ data class MPSNode(val node: SNode) : IDeprecatedNodeDefaults {
 
     override fun getPropertyValue(property: IProperty): String? {
         if (property.isIdProperty()) {
-            return node.nodeId.toString()
+            return node.reference.toString()
```
languitar, line 163: If that effectively changes the content of the #mpsNodeId# property, we should probably rename it and have to mark this as a breaking change.

mhuster23: We also stored the reference in the past. This is a bug, that was introduced in September of last year.
Technically it is a breaking change, but also not really.
I'm not sure if we should declare it breaking.

languitar: Our client code at least expects a node id there.

languitar: We could add a second property called "node reference" and put the full reference there. That would also be more consistent.

mhuster23: That would add more redundancy though. Since the reference ends with the nodeId.
If a user needs the node id they can parse it from the reference.

slisson: The bulk-sync should store `INode.reference.serialize()` and the code for that belongs into the synchronizer and not into `MPSNode`. `MPSNode` doesn't need to know anything about that special property.

`#mpsNodeId#` is not a suitable name for that value. It was originally used by the legacy-sync-plugin. I already changed it to store the reference into a property named `$orignalId` (see https://github.com/modelix/modelix.mps-plugins/blob/d9bcc94cfe0bcbc02a0b61ffb88168d33b6502cd/mps-legacy-sync-plugin/src/main/java/org/modelix/model/mpsplugin/MPSNodeMapping.kt#L12). Might still not be a good name, but we can choose one now.

Client code shouldn't directly access that property to avoid breaking changes. We need an API that returns the "original reference".

I also just saw that MPSNodeReference.serialize is incorrectly implemented as `"$PREFIX:$ref"`. It should be `"$PREFIX:${ref.modelReference?.moduleReference?.moduleId}/${ref.modelReference?.modelId}/${ref.nodeId}"` instead. The reason is, that

mhuster23: I called it `#mpsNodeRef#` for now. Maybe we should name it `#originalRef#` since we might have other original node sources in the future.

model-api/src/commonMain/kotlin/org/modelix/model/api/INode.kt
```diff
@@ -206,6 +207,12 @@ interface INode {
     @Deprecated("use getReferenceLinks()")
     fun getReferenceRoles(): List<String>
 
+    /**
+     * @return the original reference of this node
```
slisson, line 211:
```suggestion
     * @return the serialized reference of the source node, if this one was created during an import
```

slisson (approved):

PR #558, opened 2024-03-05
MODELIX-801 Parallel queries to different branches seem to create a deadlock

slisson:
fix(model-server): InMemoryModel was blocking the ktor request handler thread pool

The model was loaded by a synchronized method and all HTTP requests were blocking the thread while waiting for the result. Since the default pool is constrained in size it was running out of thread for handling other request such as the /health endpoint.

Replaced the synchronized method by coroutine features that don't block the threads.


model-datastructure/src/jvmMain/kotlin/org/modelix/model/InMemoryModel.kt
```diff
@@ -42,52 +44,66 @@ import org.modelix.model.lazy.NonCachingObjectStore
...
 
-    @Synchronized
-    fun getModel(id: String) = models.getOrPut(id) { InMemoryModelLoader(IncrementalInMemoryModel()) }
+    fun dispose() {
+        coroutineScope.cancel("disposed")
+    }
 
-    fun getModel(tree: CLTree) = getModel(tree.getId()).model.getModel(tree)
+    private fun getModelLoader(branchId: String): InMemoryModelLoader {
+        return synchronized(branchId2modelLoader) {
```
odzhychko, line 94: `branchId2modelLoader` is already a synchronized map.

model-server/src/main/kotlin/org/modelix/model/server/handlers/RepositoriesManager.kt
```diff
@@ -56,6 +56,10 @@ class RepositoriesManager(val client: LocalModelClient) {
     private val objectStore: IDeserializingKeyValueStore get() = client.storeCache
     val inMemoryModels = InMemoryModels()
 
+    fun dispose() {
```
odzhychko, line 59: This dispose is never called. Do you intend it to call in tests or Main.kt on shutdown?

odzhychko:
Looks good to me.
@languitar I tested this version with our data as we did with #557.

odzhychko (approved):

model-datastructure/src/jvmMain/kotlin/org/modelix/model/InMemoryModel.kt
slisson: `getOrPut` is an extension function and not atomic.

model-server/src/main/kotlin/org/modelix/model/server/handlers/RepositoriesManager.kt
slisson: The method should be called when the RepositoriesManager instance is not used anymore. Tests would be one place. We have to search for all places where an instance is created and add a `dispose()` call.

languitar: Is the idea to do this in this PR or later?

slisson: There are no cases where it's really necessary. I added a TODO in the code. It's save to ignore for now. Not even worth a ticket.

languitar (approved):

PR #563, opened 2024-03-07
fix(model-server): do not execute long-running repository operation on request threads

odzhychko:
Execute them on Dispatchers.IO because they are mostly blocked by waiting for data from store. Long-running operations are `computeDelta` and `mergeChanges`.

### DIFF
--- a/model-server/src/main/kotlin/org/modelix/model/server/handlers/KeyValueLikeModelServer.kt
+++ b/model-server/src/main/kotlin/org/modelix/model/server/handlers/KeyValueLikeModelServer.kt
@@ -112,7 +112,11 @@ class KeyValueLikeModelServer(val repositoriesManager: RepositoriesManager) {
                 if (isHealthy()) {
                     call.respondText(text = "healthy", contentType = ContentType.Text.Plain, status = HttpStatusCode.OK)
                 } else {
-                    call.respondText(text = "not healthy", contentType = ContentType.Text.Plain, status = HttpStatusCode.InternalServerError)
+                    call.respondText(
+                        text = "not healthy",
+                        contentType = ContentType.Text.Plain,
+                        status = HttpStatusCode.InternalServerError,
+                    )
                 }
             }
             get<Paths.getHeaders> {
@@ -317,18 +321,23 @@ class KeyValueLikeModelServer(val repositoriesManager: RepositoriesManager) {
                 HashUtil.isSha256(key) -> {
                     hashedObjects[key] = value ?: throw IllegalArgumentException("No value provided for $key")
                 }
+
                 BranchReference.tryParseBranch(key) != null -> {
                     branchChanges[BranchReference.tryParseBranch(key)!!] = value
                 }
+
                 key.startsWith(PROTECTED_PREFIX) -> {
                     throw NoPermissionException("Access to keys starting with '$PROTECTED_PREFIX' is only permitted to the model server itself.")
                 }
+
                 key.startsWith(RepositoriesManager.KEY_PREFIX) -> {
                     throw NoPermissionException("Access to keys starting with '${RepositoriesManager.KEY_PREFIX}' is only permitted to the model server itself.")
                 }
+
                 key == RepositoriesManager.LEGACY_SERVER_ID_KEY || key == RepositoriesManager.LEGACY_SERVER_ID_KEY2 -> {
                     throw NoPermissionException("'$key' is read-only.")
                 }
+
                 else -> {
                     userDefinedEntries[key] = value
                 }
@@ -342,9 +351,9 @@ class KeyValueLikeModelServer(val repositoriesManager: RepositoriesManager) {
             storeClient.putAll(userDefinedEntries)
             for ((branch, value) in branchChanges) {
                 if (value == null) {
-                    runBlocking { repositoriesManager.removeBranches(branch.repositoryId, setOf(branch.branchName)) }
+                    repositoriesManager.removeBranchesBlocking(branch.repositoryId, setOf(branch.branchName))
                 } else {
-                    runBlocking { repositoriesManager.mergeChanges(branch, value) }
+                    repositoriesManager.mergeChangesBlocking(branch, value)
                 }
             }
         }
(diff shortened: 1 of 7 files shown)


languitar (dismissed):

model-server/src/main/kotlin/org/modelix/model/server/handlers/KeyValueLikeModelServer.kt
```diff
@@ -164,7 +166,9 @@ class KeyValueLikeModelServer(val repositoriesManager: RepositoriesManager) {
                     val key = call.parameters["key"]!!
                     val value = call.receiveText()
                     try {
-                        putEntries(mapOf(key to value))
+                        withContext(Dispatchers.IO) {
```
odzhychko, line 169: This `withContext(Dispatchers.IO)` is redundant.

slisson:
The problematic cases are where a thread is blocked and not doing anything. If there are a lot of request that are blocked at the same location we need a lot of threads.

Using `withContext(Dispatchers.IO)` is not the best solution, because it just uses the unconstrained thread pool, but this still creates a lot of unnecessary threads. A better solution is to replace the blocking calls by coroutine features. 

From the changes in this PR I identified the following problematic calls:

---

The `transactions.txStart()` call in `IgniteStoreClient.runTransaction`.

We can wrap that code with a [Mutex](https://kotlinlang.org/docs/shared-mutable-state-and-concurrency.html#mutual-exclusion) to avoid trying to start a transaction while there is already an active one. This suspends the coroutine instead of blocking the thread.

---

The usage of `lazy` in `RepositoriesManager.computeDelta`.

This can be replaced by a `Deferred` data type and an async call.

---

Both changes will force you to make the methods, and all the transitive usages, suspendable. This automatically fixes all the locations where you now added `withContext(Dispatchers.IO)`, which is then not required anymore.

I guess this also answers the question

> On which level do we want to decide about execution (e.g. which thread pool to execute computation on)

We avoid blocking calls by using coroutine features. If not possible, we use `withContext(Dispatchers.IO)` around that call. The propagated `suspend` keyword automatically ensures that all HTTP endpoints are fixed.

slisson (changes requested):
The problematic cases are where a thread is blocked and not doing anything. If there are a lot of request that are blocked at the same location we need a lot of threads.

Using `withContext(Dispatchers.IO)` is not the best solution, because it just uses the unconstrained thread pool, but this still creates a lot of unnecessary threads. A better solution is to replace the blocking calls by coroutine features. 

From the changes in this PR I identified the following problematic calls:

---

The `transactions.txStart()` call in `IgniteStoreClient.runTransaction`.

We can wrap that code with a [Mutex](https://kotlinlang.org/docs/shared-mutable-state-and-concurrency.html#mutual-exclusion) to avoid trying to start a transaction while there is already an active one. This suspends the coroutine instead of blocking the thread.

---

The usage of `lazy` in `RepositoriesManager.computeDelta`.

This can be replaced by a `Deferred` data type and an `async` call.

---

Both changes will force you to make the methods, and all the transitive usages, suspendable. This automatically fixes all the locations where you now added `withContext(Dispatchers.IO)`, which is then not required anymore.

I guess this also answers the question

> On which level do we want to decide about execution (e.g. which thread pool to execute computation on)

We avoid blocking calls by using coroutine features. If not possible, we use `withContext(Dispatchers.IO)` around that call. The propagated `suspend` keyword automatically ensures that all HTTP endpoints are fixed.

model-server/src/main/kotlin/org/modelix/model/server/handlers/RepositoriesManager.kt
```diff
@@ -337,3 +371,56 @@ class RepositoriesManager(val client: LocalModelClient) {
...
+private fun Flow<Pair<String, String>>.checkObjectHashes(): Flow<Pair<String, String>> {
+    return onEach { HashUtil.checkObjectHash(it.first, it.second) }
+}
+
+class VersionDeltaCache(val store: IDeserializingKeyValueStore) {
+    private val cacheMap = LRUMap<Pair<String, String?>, SoftReference<Deferred<ObjectDataMap>>>(10)
+    private val cacheMapMutex = Mutex()
+
+    suspend fun getOrComputeDelta(versionHash: String, baseVersionHash: String): ObjectDataMap {
+        val deferredDelta = cacheMapMutex.withLock {
```
slisson, line 407: The mutex is unnecessary. `synchronized (cacheMap)` can be used, because the actual computation happens afterwards inside the `async` coroutine.

odzhychko: Done.

model-server/src/main/kotlin/org/modelix/model/server/handlers/RepositoriesManager.kt
```diff
@@ -178,53 +199,76 @@ class RepositoriesManager(val client: LocalModelClient) {
...
-                val headVersion = CLVersion(headHash, client.storeCache)
-                val newVersion = CLVersion(newVersionHash, client.storeCache)
-                require(headVersion.tree.getId() == newVersion.tree.getId()) {
-                    "Attempt to merge a model with ID '${newVersion.tree.getId()}'" +
-                        " into one with ID '${headVersion.tree.getId()}'"
-                }
-                val mergedVersion = VersionMerger(client.storeCache, client.idGenerator)
-                    .mergeChange(headVersion, newVersion)
-                mergedVersion.hash
+    suspend fun mergeChanges(branch: BranchReference, newVersionHash: String): String {
```
slisson, line 222: Having a suspendable and non-suspendable version of each function everywhere is not a nice solution. Transactions and coroutines don't seem mix very well.

I don't have a better solution yet, so let's merge it for now and change it in case we find a better alternative.

model-server/src/main/kotlin/org/modelix/model/server/handlers/RepositoriesManager.kt
```diff
@@ -178,53 +199,76 @@ class RepositoriesManager(val client: LocalModelClient) {
...
+    suspend fun removeBranches(repository: RepositoryId, branchNames: Set<String>) {
+        return store.runTransactionSuspendable {
+            removeBranchesInsideTransaction(repository, branchNames)
+        }
+    }
+
+    /**
+     * Same as [removeBranchesInsideTransaction] but assumes an already running transaction.
+     */
+    fun removeBranchesInsideTransaction(repository: RepositoryId, branchNames: Set<String>) {
```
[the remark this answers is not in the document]
odzhychko: Done. This is definitely better.

slisson (approved):

model-server/src/main/kotlin/org/modelix/model/server/handlers/RepositoriesManager.kt
odzhychko: If we could create a transaction that is not bound to a thread in Ingite, we could store it in the
coroutine context.
I got hung up on another issue when trying to look into Ignite code.
But in general, this should be doable.

PR #567, opened 2024-03-07
MODELIX-775 model-server cannot be launched if warmup is requested for a to be created repository

slisson:
The /health endpoint was unhealthy if the specified repository for the eager loading of the ModelQL in-memory model doesn't exist. But if it never gets healthy then there is no way to create that repository. Therefore, it must be legal to specify a non-existing repository.


languitar (commented):
Can we add a test for this bug fix now that this already had impact?

PR #568, opened 2024-03-08
perf(model-datastructure): use a memory efficient map for MapBasedStore

odzhychko:
This improves the memory footprint of the ModelClientV2. This in turn improves the memory footprint of bulk-model-sync.

### DIFF
--- a/bulk-model-sync-lib/build.gradle.kts
+++ b/bulk-model-sync-lib/build.gradle.kts
@@ -21,12 +21,6 @@ kotlin {
             }
         }
 
-        val jvmMain by getting {
-            dependencies {
-                implementation(libs.trove4j)
-            }
-        }
-
         val commonTest by getting {
             dependencies {
                 implementation(project(":model-api"))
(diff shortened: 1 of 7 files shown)


slisson (approved):
The difference for object types is not that big. The main advantage of the gnu trove collections is that they have implementations for primitive types. There is nothing wrong with using it in this case, but trying to reduce the number of elements needed in these maps should be the actual goal.

odzhychko:
Yeah, compared to the string entries, the `java.utils.HashMap$Node` wrappers are peanuts.
In one use on a single, `CLVersion` has already four million nodes and around six million entries in the store.

Thinking out loud further:

Reducing them in that use case (bulk-sync) is hardly possible because we access all nodes to create `NodeData`/`ModelData`. Interestingly, the JSON serialization of `NodeData`/`ModelData` is around 1 GB and GZip compressed it is only 40 MB.  (`NodeData`/`ModelData` does not have the hashes and should be significantly smaller.)

Maybe for those use cases, instead of downloading all objects into a map, serializing and deserializing just to traverse them once, we could create them on the server side and cache the compressed files. (Just downloading data might be a common enough uses case to optimize for.)
 
As for the cases, where we want to commit a new version once (also bulk-sync), the client could calculate the needed operations based on the `ModelData` and transfer serialized operations.

To reduce fetched and cached objects in continuous sync (e.g. frontend apps, IDE plugin), explicit loading and unloading like in #396 is usefull

PR #586, opened 2024-03-08
Fix value changed notification

odzhychko:
I tried out ThreadLocal and IgniteEvents.
ThreadLocals work.

With IgniteEvents we lose
* the ability to control if changes should be sent at all (`silent` parameter).
  We could only control to ignore events for SHAs
  Open: Check when we put things other than SHAs into store?

With IgniteEvents we gain:
* Simplifies code.

Open questions IgniteEvents:
* That hidden assumptions does it break?
* Are local events more or less much overhead than custom solution?

### DIFF
--- a/model-server/src/main/kotlin/org/modelix/model/server/store/IgniteStoreClient.kt
+++ b/model-server/src/main/kotlin/org/modelix/model/server/store/IgniteStoreClient.kt
@@ -18,6 +18,7 @@ import mu.KotlinLogging
 import org.apache.ignite.Ignite
 import org.apache.ignite.IgniteCache
 import org.apache.ignite.Ignition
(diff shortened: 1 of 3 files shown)


odzhychko:
@slisson Have you by chance considered Ignite events when implementing the IgniteStoreClient? If yes, do we have reasons not to use them a stick to custom tracking?

slisson:
I didn't use events, because that would create a lot of unnecessary overhead for all the immutable entries that we don't need an event for. For other performance reasons (different type of transaction) it probably makes sense to separate the immutable and non-immutable entries into different ignite caches.

slisson (dismissed):
I refactored it a bit by moving the ThreadLocal into the PendingChangeMessages class.

slisson (approved):

PR #588, opened 2024-03-10
build: check binary compatibility of Kotlin JVM APIs

slisson:
It's not always obvious if a change in the Kotlin code is a breaking change of the JVM API. Sometimes annotations such as `@JvmOverloads` are required to maintain binary compatibility.

Some changes are not breaking changes in the sense that they don't require changes in Kotlin code that uses the API, but if the JVM API is incompatible it can cause LinkageErrors at runtime, if the Kotlin code wasn't re-compiled against the new version.

See https://github.com/Kotlin/binary-compatibility-validator for details.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -46,6 +46,7 @@ plugins {
     alias(libs.plugins.dokka)
     alias(libs.plugins.node) apply false
     alias(libs.plugins.detekt) apply false
+    alias(libs.plugins.kotlin.binary.compatibility)
 }
 
 group = "org.modelix"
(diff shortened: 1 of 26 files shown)


odzhychko (approved):

languitar:
As discussed in a meeting today: Failing the build contradicts [ADR 0007](https://issues.modelix.org/articles/MODELIX-A-56/0007-API-compatibility-only-for-Kotlin). If we want to stick to that decision, we should only treat those JVM API incompatibilities as hints, for instance, by putting them in a PR comment.

build.gradle.kts
```diff
@@ -409,3 +410,39 @@ publishing {
...
+
+    /**
+     * Flag to programmatically disable compatibility validator
+     */
+//    validationDisabled = false
+
+    /**
+     * A path to a subdirectory inside the project root directory where dumps should be stored.
+     */
+//    apiDumpDirectory = "aux/validation"
```
languitar, line 447: Can we reduce this to the required settings?

languitar (commented):
Is it necessary to put the API definitions into the project? We could also store them in some CI pipeline artifacts to avoid polluting the repo with these more or less derived files.

slisson:
It's probably annoying to run `gradle dumpApi` all the time and it solves a problem that we currently don't really have and will ignore for now..

PR #596, opened 2024-03-12
MODELIX-814 Newly created child nodes aren't properly persisted in MPS

mhuster23:


bulk-model-sync-lib/src/commonMain/kotlin/org/modelix/model/sync/bulk/ModelImporter.kt
```diff
@@ -224,7 +224,9 @@ class ModelImporter(
                     val existingNodeReference = originalIdToExisting[expectedId]
                     if (existingNodeReference == null) {
                         val newChild = existingParent.addNewChild(role, newIndex, expectedConcept)
-                        newChild.setPropertyValue(NodeData.idPropertyKey, expectedId)
+                        if (newChild.originalId() == null) {
```
odzhychko, line 227: nit: Can this be sensibly teste? Maybe in `bulk-model-sync-lib-mps-test`?

mps-model-adapters/src/main/kotlin/org/modelix/model/mpsadapters/MPSLanguageRepository.kt
```diff
@@ -39,8 +38,6 @@ data class MPSLanguageRepository(private val repository: SRepository) : ILanguag
 
         val conceptDescriptor = ConceptRegistry.getInstance().getConceptDescriptor(conceptId)
 
-        if (conceptDescriptor is IllegalConceptDescriptor) return null
```
odzhychko, line 42: Is this fixing the test added in 19cccb9f? (If yes, could we combine both commits to make it more clear?)

bulk-model-sync-lib/src/commonMain/kotlin/org/modelix/model/sync/bulk/ModelImporter.kt
mhuster23: This is covered by the new test case. That's how I discovered the issue.

mps-model-adapters/src/main/kotlin/org/modelix/model/mpsadapters/MPSLanguageRepository.kt
mhuster23: No, this is WIP but might be a necessary step towards the solution.

languitar (approved):
In terms of functionality this works in the client project.

PR #597, opened 2024-03-12
Make the transformation work for multiple models and modules

benedekh:
With other words: automatically sync modules and models that are imported in the model we use, but have not been synched yet.

When downloading modules (models) are dependent modules (and models) will be downloaded into MPS.

This PR fixes:
- https://issues.modelix.org/issue/MODELIX-625
- https://issues.modelix.org/issue/MODELIX-809


KarstenDiekhoff:
question: why do you save references to singletons e.g.
`private val nodeMap = MpsToModelixMap`

occurences in ModelBinding, ModuleBinding, ModelixTreeChangeVisitor, ModelTransformer, ModuleTrensformer, NodeTransformer, ModuleChangeListener, ModuleSynchronizer, NodeSynchronizer, SyncServiceImpl

KarstenDiekhoff:
question 2: can you use Javadoc coments within a method and what dies it do?

occurences in ModelBinding, ModuleBinding, SyncQueue, ModuleChangeListener

KarstenDiekhoff:
finding: why did you remove exception handeling for more specific Exceptions in ModelSyncService?

benedekh:
> question: why do you save references to singletons e.g. `private val nodeMap = MpsToModelixMap`

Previously these Singletons were passed to the classes via constructor parameters to ease the unit testing of those classes (in unit testing those parameters can be replaced by mocks). However, those tests have not happened yet, but our code started to be jammed by the passing of the most common Singletons (i.e. NodeMap, SyncQueue, BindingsRegistry) between constructors. Therefore, I moved them to fields inside the classes where they are used. This way, they can be still replaced by mocks if needed. 

The drawback of using them directly in-place in the code is, it's almost impossible to replace them there.

benedekh:
> question 2: can you use Javadoc coments within a method and what dies it do?

I use javadoc (more specificalyl multiline) comments inside methods when I had to add some context and explanation to why we do some actions that do not seem to be trivial and obvious in those contexts. Another use-case was when I added some future TODOs that are also documented in the corresponding YouTrack issues.

benedekh:
> why did you remove exception handeling for more specific Exceptions in ModelSyncService?

I joined them into the Exception block, because the other blocks were not adding any specific value to how we handle the exceptions. With other words: the error message prefixes were very similar and from the StackTrace we can see anyways what kind of exception flew.

However, I can revert that change if you wish to have the previous error handling.

benedekh:
https://github.com/modelix/modelix.core/pull/597#issuecomment-1993664529

As discussed on Slack, I've replaced javadoc-like comments with multiline comments inside methods. [57d1f81](https://github.com/modelix/modelix.core/pull/597/commits/57d1f81dcc005b161dbbfb5e889169704ab1c7a6) fixes this

PR #600, opened 2024-03-13
fix(model-client): fix usage of concept references `MetaModelBranch`

odzhychko:
Though `MetaModelBranch` is deprecated, some customers have old data they access through it. Without this fix, `INode.concept.getReference()` and `INode.getConceptReference()` returned different results. `INode.getConceptReference()` was not correctly resolved by `MetaModelBranch` which broke the generated typed API which was using it.

### DIFF
--- a/model-client/src/commonMain/kotlin/org/modelix/model/metameta/PersistedConcept.kt
+++ b/model-client/src/commonMain/kotlin/org/modelix/model/metameta/PersistedConcept.kt
@@ -13,6 +13,7 @@
  */
 package org.modelix.model.metameta
 
+import org.modelix.model.api.ConceptReference
 import org.modelix.model.api.IChildLink
 import org.modelix.model.api.IConcept
 import org.modelix.model.api.IConceptReference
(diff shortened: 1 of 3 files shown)


slisson:
We should add a migration to [ModelMigrations](https://github.com/modelix/modelix.core/blob/main/model-client/src/commonMain/kotlin/org/modelix/model/ModelMigrations.kt) and migrate these old models as soon as possible.

model-client/src/commonMain/kotlin/org/modelix/model/metameta/PersistedConcept.kt
```diff
@@ -42,7 +43,8 @@ data class PersistedConcept(val id: Long, val uid: String?) : IConcept, IConcept
     }
 
     override fun getReference(): IConceptReference {
-        throw UnsupportedOperationException()
+        // The reference uses the ID and not the UID because of how MetaModelSynchronizer works.
```
slisson, line 46: The `UID` would be correct. Where exactly is that be a problem in `MetaModelSynchronizer`?

odzhychko:
> We should add a migration to [ModelMigrations](https://github.com/modelix/modelix.core/blob/main/model-client/src/commonMain/kotlin/org/modelix/model/ModelMigrations.kt) and migrate these old models as soon as possible.

We also have this topic on the agenda with the customer.
https://issues.modelix.org/issue/MODELIX-824/Remove-MetaModelBranch

model-client/src/commonMain/kotlin/org/modelix/model/metameta/PersistedConcept.kt
odzhychko: I added a longer explanation, why we want the `PersistedConcept#id` be the `ConceptReference#uid` of the created reference.

odzhychko:
This fix will not implemented. We will directly implement the migration.

PR #624, opened 2024-03-20
fix(model-datastructure): make `MapBasedStore` thread safe

odzhychko:
Make it thread safe by using a synchronized map.

Missing thread safety is the suspected issue for such error messages

```
2024-03-20 08:58:01,282 [  48184]  ERROR - #com.intellij.ide.plugins.PluginManager - Equal objects must have equal hashcodes. During rehashing, Trove discovered that the following two objects claim to be equal (as in java.lang.Object.equals()) but their hashCodes (or those calculated by your TObjectHashingStrategy) are not equal.This violates the general contract of java.lang.Object.hashCode().  See bullet point two in that method's documentation. object #1 =class java.lang.String id= 996201210 hashCode= -209607169 toString= sHB6r*02gLuXVEvI31E7v9ijDVL5VSdSO8F9OO-2z2CA; object #2 =class java.lang.String id= 1530156665 hashCode= -209607169 toString= sHB6r*02gLuXVEvI31E7v9ijDVL5VSdSO8F9OO-2z2CA
[Warning] apparent concurrent modification of the key set. Size before and after rehash() do not match 81 vs 83
java.lang.IllegalArgumentException: Equal objects must have equal hashcodes. During rehashing, Trove discovered that the following two objects claim to be equal (as in java.lang.Object.equals()) but their hashCodes (or those calculated by your TObjectHashingStrategy) are not equal.This violates the general contract of java.lang.Object.hashCode().  See bullet point two in that method's documentation. object #1 =class java.lang.String id= 996201210 hashCode= -209607169 toString= sHB6r*02gLuXVEvI31E7v9ijDVL5VSdSO8F9OO-2z2CA; object #2 =class java.lang.String id= 1530156665 hashCode= -209607169 toString= sHB6r*02gLuXVEvI31E7v9ijDVL5VSdSO8F9OO-2z2CA
[Warning] apparent concurrent modification of the key set. Size before and after rehash() do not match 81 vs 83
	at gnu.trove.impl.hash.TObjectHash.buildObjectContractViolation(TObjectHash.java:464)
	at gnu.trove.impl.hash.TObjectHash.throwObjectContractViolation(TObjectHash.java:448)
	at gnu.trove.map.hash.THashMap.rehash(THashMap.java:403)
	at gnu.trove.impl.hash.THash.ensureCapacity(THash.java:175)
	at gnu.trove.map.hash.THashMap.putAll(THashMap.java:551)
	at org.modelix.model.persistent.MapBasedStore.putAll(MapBasedStore.kt:54)
```

### DIFF
--- a/kotlin-utils/src/commonMain/kotlin/org/modelix/kotlin/utils/DataStructures.kt
+++ b/kotlin-utils/src/commonMain/kotlin/org/modelix/kotlin/utils/DataStructures.kt
@@ -29,3 +29,5 @@ package org.modelix.kotlin.utils
  * Therefore, the memory efficient maps are used sparingly for only the very big maps.
  */
 expect fun <K, V> createMemoryEfficientMap(): MutableMap<K, V>
+
+expect fun <K, V> MutableMap<K, V>.toSynchronizedMap(): MutableMap<K, V>
--- a/kotlin-utils/src/jsMain/kotlin/org/modelix/kotlin/utils/DataStructures.js.kt
+++ b/kotlin-utils/src/jsMain/kotlin/org/modelix/kotlin/utils/DataStructures.js.kt
@@ -17,3 +17,8 @@
 package org.modelix.kotlin.utils
 
 actual fun <K, V> createMemoryEfficientMap(): MutableMap<K, V> = HashMap()
+
+actual fun <K, V> MutableMap<K, V>.toSynchronizedMap(): MutableMap<K, V> {
+    // Because JS is single-threaded, no extra measures are needed.
+    return this
+}
--- a/kotlin-utils/src/jvmMain/kotlin/org/modelix/kotlin/utils/DataStructures.jvm.kt
+++ b/kotlin-utils/src/jvmMain/kotlin/org/modelix/kotlin/utils/DataStructures.jvm.kt
@@ -17,5 +17,10 @@
 package org.modelix.kotlin.utils
 
 import gnu.trove.map.hash.THashMap
+import java.util.Collections
 
 actual fun <K, V> createMemoryEfficientMap(): MutableMap<K, V> = THashMap()
+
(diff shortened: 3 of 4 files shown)


odzhychko:
> Could this again become an issue with threading primitives being used in coroutine contexts later on?

I am not sure if I understand you correctly. Maybe my answer still helps.

The synchronized map ensures that two threads do not use the map at the same time.
When a coroutine is executed, it is executed on a thread.
Therefore, two coroutines running at the same time on different threads will not use the synchronized map at the same time.

languitar:
I was wondering if introducing more threading locks can lead to a similar situation fixed with https://github.com/modelix/modelix.core/pull/558. Maybe usage patterns should once be checked.

odzhychko:
`MapBasedStore` is only used in the `ModelClientV2`. So this change will not affect the server.

languitar:
image

Maybe we should configure cucumber to not include the chrome version in the test name?

odzhychko:
> Maybe we should configure cucumber to not include the chrome version in the test name?

I think this is caused by Kotlin-Multiplatform and not Cucumber.
I created a ticket for that issue https://issues.modelix.org/issue/MODELIX-836

kotlin-utils/src/commonMain/kotlin/org/modelix/kotlin/utils/DataStructures.kt
```diff
@@ -29,3 +29,5 @@ package org.modelix.kotlin.utils
  * Therefore, the memory efficient maps are used sparingly for only the very big maps.
  */
 expect fun <K, V> createMemoryEfficientMap(): MutableMap<K, V>
+
+expect fun <K, V> synchronizedMap(map: MutableMap<K, V>): MutableMap<K, V>
```
slisson, line 33: `fun <K, V> MutableMap<K, V>.toSynchronizedMap(): MutableMap<K, V>` might be more idiomatic Kotlin code.

slisson (approved):

PR #625, opened 2024-03-20
MODELIX-641 Replace GenerateAntScriptForMpsMetaModelExport

slisson:


metamodel-export/src/main/kotlin/org/modelix/metamodel/export/MPSMetaModelExporter.kt
```diff
@@ -0,0 +1,480 @@
...
+
+    companion object {
+        private fun <T> as_y4kbwa_a0a3a8(o: Any, type: Class<T>): T? {
+            return (if (type.isInstance(o)) o as T else null)
+        }
+
+        private fun <T> as_y4kbwa_a0a0a0a0a0f0a0a0a6a01(o: Any, type: Class<T>): T? {
+            return (if (type.isInstance(o)) o as T else null)
+        }
+    }
```
mhuster23, line 479: Seems like the generation has created some unreadable code. Can we refactor this? These extra methods are unneccessary overhead.

metamodel-export/src/main/kotlin/org/modelix/metamodel/export/MPSModelExporter.kt
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    private fun exportNode(node: SNode): NodeData {
+        val id: String = node.reference.toString()
+        val concept: String = "mps:" + MetaIdHelper.getConcept(SNodeOperations.getConcept(node)).serialize()
+        val role = node.containmentLink?.name ?: "rootNodes"
+        val children: List<NodeData> = SNodeOperations.getChildren(node).map { exportNode(it) }
+        val properties: MutableMap<String, String> = LinkedHashMap()
+        val references: MutableMap<String, String> = LinkedHashMap()
+        properties["#mpsNodeId#"] = node.nodeId.toString()
```
mhuster23, line 105: Maybe we should use `#originalRef#` here. But I don't know if it's worth it to change it here, since we wanted to replace this functionality with a part of the bulk-sync.

slisson: This is the existing behavior. Changing that is out of scope for this PR. We will eventually remove this legacy code.

slisson: https://issues.modelix.org/issue/MODELIX-842/Remove-model-export-feature-from-api-gen-gradle

metamodel-export/src/main/kotlin/org/modelix/metamodel/export/MPSMetaModelExporter.kt
slisson: I did some additional cleanup of the converted code.

mhuster23 (approved):

PR #643, opened 2024-03-26
MODELIX-832: MPSLanguageRepository returns "unknown" Concept for unknown concepts

benedekh:
I took the courage to revert the commit, because it makes the new sync plugin fail. A workaround for the new new sync plugin is in [this PR](https://github.com/modelix/modelix.mps-plugins/pull/70).

Since I'm not familiar with the bulk-sync at all, I would leave the following task for someone with more experience.

# TODO from the issue comments

- [ ] find a differnt solution for the bulk-sync

### DIFF
--- a/mps-model-adapters/src/main/kotlin/org/modelix/model/mpsadapters/MPSNode.kt
+++ b/mps-model-adapters/src/main/kotlin/org/modelix/model/mpsadapters/MPSNode.kt
@@ -145,8 +145,39 @@ data class MPSNode(val node: SNode) : IDefaultNodeAdapter {
 
     override fun addNewChild(role: IChildLink, index: Int, concept: IConceptReference?): INode {
         val repo = checkNotNull(node.model?.repository)
-        val targetConcept = concept?.let { MPSLanguageRepository(repo).resolveConcept(it.getUID()) }
+        val targetConcept = concept?.let {
+            MPSLanguageRepository(repo).resolveConcept(it.getUID())
+                ?: MPSConcept.tryParseUID(it.getUID())
+        }
+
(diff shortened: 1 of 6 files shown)


languitar (commented):
Any chance to add a test for this condition now that we have better test infrastructure for the MPS adapters?

mhuster23:
We can't merge it like this without a fix for the bulk-sync as it will break the customer project.

slisson:
I'm done fixing it. Can be reviewed and merged.

mps-model-adapters/src/main/kotlin/org/modelix/model/mpsadapters/MPSNode.kt
```diff
@@ -145,10 +145,41 @@ data class MPSNode(val node: SNode) : IDefaultNodeAdapter {
...
     }
 
+    override fun addNewChildren(role: String?, index: Int, concepts: List<IConceptReference?>): List<INode> {
+        requireNotNull(role) { "containment link required" }
+        val link = MPSChildLink(getMPSContainmentLink(role))
+        return addNewChildren(link, index, concepts)
+    }
+
+    override fun addNewChildren(link: IChildLink, index: Int, concepts: List<IConceptReference?>): List<INode> {
+        return concepts.asReversed().map { addNewChild(link, index, it) }.asReversed()
```
mhuster23, line 166: Why do we need `asReversed()` twice here?

slisson: Because they are inserted using the same index and would then end up in reversed order in the model, but that's actually not true if -1 is used for the index, which would append them in the correct order. I fixed that. Should also be easier to understand now.

mhuster23 (approved):

PR #646, opened 2024-03-26
chore(model-server): use non-blocking method to send version delta

odzhychko:
Use `respondBytesWriter` because it has a non-blocking `writeStringUtf8` method. `respondTextWriter` has only blocking methods to write data.

Using a `Writer` is just an indirection without purpose.

### DIFF
--- a/model-server/src/main/kotlin/org/modelix/model/server/handlers/IRepositoriesManager.kt
+++ b/model-server/src/main/kotlin/org/modelix/model/server/handlers/IRepositoriesManager.kt
@@ -1,0 +1,62 @@
+/*
+ * Copyright (c) 2024.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.modelix.model.server.handlers
+
+import org.modelix.model.lazy.BranchReference
+import org.modelix.model.lazy.CLVersion
+import org.modelix.model.lazy.RepositoryId
+
+interface IRepositoriesManager {
+    /**
+     * Used to retrieve the server ID. If needed, the server ID is created and stored.
+     *
+     * If a server ID was not created yet, it is generated and saved in the database.
+     * It gets stored under the current and all legacy database keys.
+     *
+     * If the server ID was created previously but is only stored under a legacy database key,
+     * it also gets stored under the current and all legacy database keys.
+     */
+    suspend fun maybeInitAndGetSeverId(): String
+    fun getRepositories(): Set<RepositoryId>
+    suspend fun createRepository(repositoryId: RepositoryId, userName: String?, useRoleIds: Boolean = true): CLVersion
+    suspend fun removeRepository(repository: RepositoryId): Boolean
+
+    fun getBranches(repositoryId: RepositoryId): Set<BranchReference>
+
+    suspend fun removeBranches(repository: RepositoryId, branchNames: Set<String>)
+
+    /**
+     * Same as [removeBranches] but blocking.
+     * Caller is expected to execute it outside the request thread.
+     */
+    fun removeBranchesBlocking(repository: RepositoryId, branchNames: Set<String>)
+    suspend fun getVersion(branch: BranchReference): CLVersion?
+    suspend fun getVersionHash(branch: BranchReference): String?
+    suspend fun pollVersionHash(branch: BranchReference, lastKnown: String?): String
+    suspend fun mergeChanges(branch: BranchReference, newVersionHash: String): String
+
+    /**
+     * Same as [mergeChanges] but blocking.
+     * Caller is expected to execute it outside the request thread.
+     */
+    fun mergeChangesBlocking(branch: BranchReference, newVersionHash: String): String
+    suspend fun computeDelta(versionHash: String, baseVersionHash: String?): ObjectData
+}
+
+fun IRepositoriesManager.getBranchNames(repositoryId: RepositoryId): Set<String> {
+    return getBranches(repositoryId).map { it.branchName }.toSet()
+}
--- a/model-server/src/main/kotlin/org/modelix/model/server/handlers/ModelReplicationServer.kt
+++ b/model-server/src/main/kotlin/org/modelix/model/server/handlers/ModelReplicationServer.kt
@@ -379,8 +379,11 @@ class ModelReplicationServer(
     }
 
     private suspend fun ApplicationCall.respondDeltaAsObjectStream(versionHash: String, baseVersionHash: String?, plainText: Boolean) {
+        // Call `computeDelta` before starting to respond.
+        // It could already throw an exception, and in that case we do not want a successful response status.
+        val objectData = repositoriesManager.computeDelta(versionHash, baseVersionHash)
         respondTextWriter(contentType = if (plainText) ContentType.Text.Plain else VersionDeltaStream.CONTENT_TYPE) {
-            repositoriesManager.computeDelta(versionHash, baseVersionHash).asFlow()
+            objectData.asFlow()
                 .flatten()
                 .withSeparator("\n")
                 .onEmpty { emit(versionHash) }
--- a/model-server/src/main/kotlin/org/modelix/model/server/handlers/ModelReplicationServer.kt
+++ b/model-server/src/main/kotlin/org/modelix/model/server/handlers/ModelReplicationServer.kt
@@ -27,13 +27,17 @@ import io.ktor.server.resources.get
 import io.ktor.server.resources.post
 import io.ktor.server.resources.put
 import io.ktor.server.response.respond
+import io.ktor.server.response.respondBytesWriter
 import io.ktor.server.response.respondText
-import io.ktor.server.response.respondTextWriter
 import io.ktor.server.routing.Route
 import io.ktor.server.routing.route
 import io.ktor.server.routing.routing
 import io.ktor.server.websocket.webSocket
+import io.ktor.util.cio.use
 import io.ktor.util.pipeline.PipelineContext
+import io.ktor.utils.io.ByteWriteChannel
+import io.ktor.utils.io.close
+import io.ktor.utils.io.writeStringUtf8
 import io.ktor.websocket.send
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.Job
@@ -84,6 +88,7 @@ class ModelReplicationServer(
     companion object {
         private val LOG = LoggerFactory.getLogger(ModelReplicationServer::class.java)
     }
+
     private val storeClient: IStoreClient get() = modelClient.store
 
     fun init(application: Application) {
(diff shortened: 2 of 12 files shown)


odzhychko:
This might have an unwanted consequences, when the fetching throws an error. I have to check this again.

languitar:
Is this somehow performance critical or does this fix a bug? I'm confused that this is only a `chore` commit.

odzhychko:
> Is this somehow performance critical or does this fix a bug? I'm confused that this is only a `chore` commit.

This removes unneeded usage of `java.io.Writer` and `runBlocking`.
See `ChannelWriterContent` vs `WriterContent`.

But the `runBlocking` was not executed on a request thread anyway.
I do not assume any great performance improvements.
So this is still more a chore than a performance improvement.


```

public class ChannelWriterContent(
    private val body: suspend ByteWriteChannel.() -> Unit,
    override val contentType: ContentType?,
    override val status: HttpStatusCode? = null,
    override val contentLength: Long? = null
) : OutgoingContent.WriteChannelContent() {
    override suspend fun writeTo(channel: ByteWriteChannel) {
        body(channel)
    }
}

```

```

public class WriterContent(
    private val body: suspend Writer.() -> Unit,
    override val contentType: ContentType,
    override val status: HttpStatusCode? = null,
    override val contentLength: Long? = null
) : OutgoingContent.WriteChannelContent() {

    override suspend fun writeTo(channel: ByteWriteChannel) {
        val charset = contentType.charset() ?: Charsets.UTF_8
        withBlocking {
            channel.writer(charset).use {
                it.body()
            }
        }
    }
}

```

odzhychko:
> This might have an unwanted consequence, when an error is thrown while producing the flow. I have to check this again.

Looking for the changed behavior, I was led to a bug that will only be fixed in Ktor 3.
See https://youtrack.jetbrains.com/issue/KTOR-4862/Ktor-hangs-if-exception-occurs-during-write-response-body

This led me to write a test with a Netty server to test our desired behavior.
To write the tests, I needed to introduce `IRepositoriesManager`.
Writing tests made me discover a different bug in our implementation that I also fixed.

model-server/src/main/kotlin/org/modelix/model/server/handlers/ModelReplicationServer.kt
```diff
@@ -372,17 +381,34 @@
...
+                objectData.asFlow()
+                    .flatten()
+                    .withSeparator("\n")
+                    .onEmpty { emit(versionHash) }
+                    .withIndex()
+                    .collect {
+                        if (it.index == 0) check(it.value == versionHash) { "First object should be the version" }
+                        writeStringUtf8(it.value)
+                    }
+            } catch (cause: Throwable) {
```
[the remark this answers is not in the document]
languitar: Any chance to be more specific here? Or at least resort to Exception and not Throwable?

model-server/src/test/kotlin/org/modelix/model/server/ModelServerTestUtil.kt
```diff
@@ -37,3 +41,23 @@ fun Application.installDefaultServerPlugins() {
     install(Resources)
     install(IgnoreTrailingSlash)
 }
+
+fun runWithNettyServer(
```
languitar, line 45: Maybe some documentation on when to use this over the default test server of ktor would be good.

model-server/src/test/kotlin/org/modelix/model/server/handlers/ModelReplicationServerTest.kt
```diff
@@ -133,4 +144,59 @@ class ModelReplicationServerTest {
             assertEquals(HttpStatusCode.InternalServerError, response.status)
         }
     }
+
+    @Test
+    fun `server closes connection when failing to compute delta after starting to respond`() =
+        kotlinx.coroutines.test.runTest {
```
languitar, line 150: Why no import?

model-server/src/test/kotlin/org/modelix/model/server/handlers/ModelReplicationServerTest.kt
```diff
@@ -133,4 +144,59 @@ class ModelReplicationServerTest {
...
+            val setupBlock = { application: Application -> modelReplicationServer.init(application) }
+            val testBlock: suspend (server: NettyApplicationEngine) -> Unit = { server: NettyApplicationEngine ->
+                withTimeout(10.seconds) {
+                    val port = server.resolvedConnectors().first().port
+                    val url = "http://localhost:$port/v2"
+                    val client = HttpClient(CIO) {
+                        install(HttpTimeout) {
+                            requestTimeoutMillis = 5_000
+                        }
+                    }
```
languitar, line 184: Couldn't the runWithNettyServer provide the client as the argument to the test block. That would make reading this test a bit easier.

languitar (commented):
Nice catches for annoying bugs that I would have hoped the framework prevented in the first place...

Some ideas for code improvements still.

model-server/src/test/kotlin/org/modelix/model/server/ModelServerTestUtil.kt
odzhychko: Done

model-server/src/main/kotlin/org/modelix/model/server/handlers/ModelReplicationServer.kt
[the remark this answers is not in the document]
odzhychko: I now implemented a solution that does not need a catch block and is reusabel.

---

It seems `Throwable`s are usually caught in libary code.

> It would be more typical in a framework type application (for example an application server or a testing framework) where it can be running unknown code and should not be affected by anything that goes wrong with that code, as much as possible

See https://stackoverflow.com/questions/2274102/difference-between-using-throwable-and-exception-in-a-try-catch

> ```kt
> public inline fun ByteWriteChannel.use(block: ByteWriteChannel.() -> Unit) {
>    contract {
>        callsInPlace(block, InvocationKind.EXACTLY_ONCE)
>    }
>
>    try {
>        block()
>    } catch (cause: Throwable) {
>        close(cause)
>        throw cause
>    } finally {
>        close()
>    }
>}
>```


See `ByteWriteChannel.use(block: ByteWriteChannel.() -> Unit)`

model-server/src/main/kotlin/org/modelix/model/server/handlers/RepositoriesManager.kt
```diff
@@ -51,7 +50,7 @@
 import java.lang.ref.SoftReference
 import java.util.UUID
 
-class RepositoriesManager(val client: LocalModelClient) {
+class RepositoriesManager(val client: LocalModelClient) : IRepositoriesManager {
```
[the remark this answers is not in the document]
odzhychko: I do not want to further refactor this part of the code.

model-server/src/main/kotlin/org/modelix/model/server/handlers/IRepositoriesManager.kt
```diff
@@ -0,0 +1,62 @@
...
+ * limitations under the License.
+ */
+
+package org.modelix.model.server.handlers
+
+import org.modelix.model.lazy.BranchReference
+import org.modelix.model.lazy.CLVersion
+import org.modelix.model.lazy.RepositoryId
+
+interface IRepositoriesManager {
```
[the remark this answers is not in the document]
odzhychko: I do not want to further refactor this part of the code.

languitar, line 23: Now that I look at this again: Let's also add some documentation here to clarify the purpose of the interface and potential design decisions behind it.

model-server/src/main/kotlin/org/modelix/model/server/handlers/IRepositoriesManager.kt
```diff
@@ -0,0 +1,62 @@
...
+     */
+    fun removeBranchesBlocking(repository: RepositoryId, branchNames: Set<String>)
+    suspend fun getVersion(branch: BranchReference): CLVersion?
+    suspend fun getVersionHash(branch: BranchReference): String?
+    suspend fun pollVersionHash(branch: BranchReference, lastKnown: String?): String
+    suspend fun mergeChanges(branch: BranchReference, newVersionHash: String): String
+
+    /**
+     * Same as [mergeChanges] but blocking.
+     * Caller is expected to execute it outside the request thread.
```
languitar, line 54: Not sure if this comment is appropriate here as the repositories manager shouldn't know anything about webservers.

model-server/src/main/kotlin/org/modelix/model/server/handlers/IRepositoriesManager.kt
```diff
@@ -0,0 +1,62 @@
...
+    /**
+     * Used to retrieve the server ID. If needed, the server ID is created and stored.
+     *
+     * If a server ID was not created yet, it is generated and saved in the database.
+     * It gets stored under the current and all legacy database keys.
+     *
+     * If the server ID was created previously but is only stored under a legacy database key,
+     * it also gets stored under the current and all legacy database keys.
+     */
+    suspend fun maybeInitAndGetSeverId(): String
```
languitar, line 33: Sounds weird to me that this method is part of the repository manager interface. That's a separate functionality that only accidentally or due to implementation reasons makes use of the repository. Maybe a separate interface?

model-server/src/main/kotlin/org/modelix/model/server/handlers/IRepositoriesManager.kt
```diff
@@ -0,0 +1,62 @@
...
+    fun getRepositories(): Set<RepositoryId>
+    suspend fun createRepository(repositoryId: RepositoryId, userName: String?, useRoleIds: Boolean = true): CLVersion
```
languitar, line 35: This lacks some documentation now that we touch the code ;-)

* Why is the username needed here?
* Why can it be nullable?
* What is `useRoleIds`?

model-server/src/main/kotlin/org/modelix/model/server/handlers/IRepositoriesManager.kt
```diff
@@ -0,0 +1,62 @@
...
+
+interface IRepositoriesManager {
+    /**
+     * Used to retrieve the server ID. If needed, the server ID is created and stored.
+     *
+     * If a server ID was not created yet, it is generated and saved in the database.
+     * It gets stored under the current and all legacy database keys.
+     *
+     * If the server ID was created previously but is only stored under a legacy database key,
+     * it also gets stored under the current and all legacy database keys.
```
languitar, line 31: Shouldn't this documentation be part of the implementation?

model-server/src/main/kotlin/org/modelix/model/server/handlers/IRepositoriesManager.kt
```diff
@@ -0,0 +1,62 @@
...
+     */
+    suspend fun maybeInitAndGetSeverId(): String
+    fun getRepositories(): Set<RepositoryId>
+    suspend fun createRepository(repositoryId: RepositoryId, userName: String?, useRoleIds: Boolean = true): CLVersion
+    suspend fun removeRepository(repository: RepositoryId): Boolean
+
+    fun getBranches(repositoryId: RepositoryId): Set<BranchReference>
```
languitar, line 38: Any reason for not making all operations suspend functions in this interface?

model-server/src/main/kotlin/org/modelix/model/server/handlers/IRepositoriesManager.kt
```diff
@@ -0,0 +1,62 @@
...
+
+    suspend fun removeBranches(repository: RepositoryId, branchNames: Set<String>)
+
+    /**
+     * Same as [removeBranches] but blocking.
+     * Caller is expected to execute it outside the request thread.
+     */
+    fun removeBranchesBlocking(repository: RepositoryId, branchNames: Set<String>)
+    suspend fun getVersion(branch: BranchReference): CLVersion?
+    suspend fun getVersionHash(branch: BranchReference): String?
```
languitar, line 48: We should add documentation when a null return value is expected.

model-server/src/main/kotlin/org/modelix/model/server/handlers/IRepositoriesManager.kt
languitar, line 23: Method documentations should ideally give hints on which exceptions to be expected by each method.

model-server/src/test/kotlin/org/modelix/model/server/handlers/ModelReplicationServerTest.kt
odzhychko: For now I do not know the futur usage patterns and would like to keep it flexibel.
For example I'm not sure, which timeout is reasonabel for the client.
Same goes for the timeout in the `testBlock` `withTimeout(10.seconds)`.

languitar (commented):
Sorry for another iteration. After looking at this again, I see some potential to improve our documentation state now that a new interface is extracted.

model-server/src/main/kotlin/org/modelix/model/server/handlers/RepositoriesManager.kt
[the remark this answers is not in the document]
languitar: The interface looks generally fine to me in terms of the number and cohesion of methods in included. So maybe we can put an explicit suppression for this alert in the code?

model-server/src/test/kotlin/org/modelix/model/server/handlers/ModelReplicationServerTest.kt
languitar: Alright, so then maybe put the client creation code that also derives the URL etc into a method somewhere close to the test? I was pretty distracted from all the client creation details when trying to understand what is actually being tested here.

model-server/src/test/kotlin/org/modelix/model/server/ModelServerTestUtil.kt
```diff
@@ -37,3 +41,33 @@ fun Application.installDefaultServerPlugins() {
     install(Resources)
     install(IgnoreTrailingSlash)
 }
+
+/**
+ * Allow running a model server for test with Netty.
+ *
+ * This allows testing properties that would not be testable with [io.ktor.server.testing.testApplication]
+ * because the request requests run on proper request threads.
```
languitar, line 49:
```suggestion
 * because the requests run on proper request threads.
```

model-server/src/test/kotlin/org/modelix/model/server/handlers/ModelReplicationServerTest.kt
odzhychko: Done. I created method for creating the client inside the test method.

model-server/src/main/kotlin/org/modelix/model/server/handlers/RepositoriesManager.kt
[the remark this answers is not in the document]
odzhychko: Done (Mal schauen, ob das von Detekt direkt erkannt wird)

odzhychko:
> Sorry for another iteration. After looking at this again, I see some potential to improve our documentation state now that a new interface is extracted.

I agree, but I would like to do the cleanup on an independant branch.

languitar:
> > Sorry for another iteration. After looking at this again, I see some potential to improve our documentation state now that a new interface is extracted.
> 
> I agree, but I would like to do the cleanup on an independant branch an PR.

Too bad we can't move discussions to other PRs.

languitar (approved):

PR #650, opened 2024-03-27
fix(model-client): detect incomplete version delta streams

odzhychko:
Introduce a new version delta stream format that enables the client to properly detect incomplete delta version streams.

Old clients can still use the old format, but they might not detect incomplete streams.

### DIFF
--- a/model-client/src/commonMain/kotlin/org/modelix/model/client2/ModelClientV2.kt
+++ b/model-client/src/commonMain/kotlin/org/modelix/model/client2/ModelClientV2.kt
@@ -65,6 +65,7 @@ import org.modelix.model.persistent.HashUtil
 import org.modelix.model.persistent.MapBasedStore
 import org.modelix.model.server.api.v2.VersionDelta
 import org.modelix.model.server.api.v2.VersionDeltaStream
+import org.modelix.model.server.api.v2.VersionDeltaStreamV2
 import org.modelix.model.server.api.v2.asStream
 import org.modelix.modelql.client.ModelQLClient
 import org.modelix.modelql.core.IMonoStep
@@ -491,32 +492,47 @@ private fun URLBuilder.appendPathSegmentsEncodingSlash(vararg components: String
 fun VersionDelta.getAllObjects(): Map<String, String> = objectsMap + objects.associateBy { HashUtil.sha256(it) }
 
 suspend fun HttpResponse.readVersionDelta(): VersionDeltaStream {
-    return if (contentType()?.match(VersionDeltaStream.CONTENT_TYPE) == true) {
-        val content = bodyAsChannel()
-        val versionHash = checkNotNull(content.readUTF8Line()) { "No objects received" }
-        val versionObject = content.readUTF8Line()
-        return if (versionObject == null) {
-            VersionDeltaStream(versionHash, emptyFlow())
-        } else {
-            VersionDeltaStream(
-                versionHash,
-                flow {
-                    emit(versionHash to versionObject)
-                    while (true) {
-                        val key = content.readUTF8Line() ?: break
-                        val value = checkNotNull(content.readUTF8Line()) { "Object missing for hash $key" }
-                        emit(key to value)
-                    }
-                },
-            )
-        }
+    val parsedContentType = contentType()
+    return if (parsedContentType?.match(VersionDeltaStreamV2.CONTENT_TYPE) == true) {
+        return readVersionDeltaStreamV2()
+    } else if (parsedContentType?.match(VersionDeltaStream.CONTENT_TYPE) == true) {
+        return readVersionDeltaStreamV1()
     } else {
         body<VersionDelta>().asStream()
     }
 }
 
+private suspend fun HttpResponse.readVersionDeltaStreamV1(): VersionDeltaStream {
+    val content = bodyAsChannel()
+    val versionHash = checkNotNull(content.readUTF8Line()) { "No objects received" }
+    val versionObject = content.readUTF8Line()
+    return if (versionObject == null) {
+        VersionDeltaStream(versionHash, emptyFlow())
+    } else {
+        VersionDeltaStream(
+            versionHash,
+            flow {
+                emit(versionHash to versionObject)
+                while (true) {
+                    val key = content.readUTF8Line() ?: break
+                    val value = checkNotNull(content.readUTF8Line()) { "Object missing for hash $key" }
+                    emit(key to value)
+                }
+            },
+        )
+    }
+}
+
+private suspend fun HttpResponse.readVersionDeltaStreamV2(): VersionDeltaStream {
+    val content = bodyAsChannel()
+    val decodeVersionDeltaStreamV2 = VersionDeltaStreamV2.decodeVersionDeltaStreamV2(content)
+    return VersionDeltaStream(decodeVersionDeltaStreamV2.versionHash, decodeVersionDeltaStreamV2.hashesWithDeltaObject)
+}
+
 fun HttpRequestBuilder.useVersionStreamFormat() {
-    headers.set(HttpHeaders.Accept, VersionDeltaStream.CONTENT_TYPE.toString())
+    headers[HttpHeaders.Accept] = VersionDeltaStreamV2.CONTENT_TYPE.toString()
+    // Add CONTENT_TYPE_VERSION_DELTA_V1 so that newer clients cant talk with older servers.
+    headers.append(HttpHeaders.Accept, VersionDeltaStream.CONTENT_TYPE.toString())
 }
 
 /**
--- a/model-server-api/src/commonMain/kotlin/org/modelix/model/server/api/v2/VersionDeltaStreamV2.kt
+++ b/model-server-api/src/commonMain/kotlin/org/modelix/model/server/api/v2/VersionDeltaStreamV2.kt
@@ -1,0 +1,110 @@
+/*
+ * Copyright (c) 2024.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.modelix.model.server.api.v2
+
+import io.ktor.http.ContentType
+import io.ktor.utils.io.ByteReadChannel
+import io.ktor.utils.io.ByteWriteChannel
+import io.ktor.utils.io.readUTF8Line
+import io.ktor.utils.io.writeStringUtf8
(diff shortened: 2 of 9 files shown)


odzhychko:
Use magic byte to detect end.

model-client/src/commonMain/kotlin/org/modelix/model/client2/ModelClientV2.kt
```diff
@@ -491,32 +492,55 @@ private fun URLBuilder.appendPathSegmentsEncodingSlash(vararg components: String
...
-                        val key = content.readUTF8Line() ?: break
-                        val value = checkNotNull(content.readUTF8Line()) { "Object missing for hash $key" }
-                        emit(key to value)
-                    }
-                },
-            )
-        }
+    return if (contentType()?.match(VersionDeltaStreamV2.CONTENT_TYPE) == true) {
+        return readVersionDeltaStreamV2()
+    } else if (contentType()?.match(VersionDeltaStream.CONTENT_TYPE) == true) {
```
languitar, line 497: Not sure how expensive `contentType()` is. Otherwise, the result could be put into a variable to avoid the repeated calls.

model-client/src/commonMain/kotlin/org/modelix/model/client2/ModelClientV2.kt
```diff
@@ -491,32 +492,55 @@ private fun URLBuilder.appendPathSegmentsEncodingSlash(vararg components: String
...
+                    emit(key to value)
+                }
+            },
+        )
+    }
+}
+
+private suspend fun HttpResponse.readVersionDeltaStreamV2(): VersionDeltaStream {
+    val content = bodyAsChannel()
+    return VersionDeltaStreamV2.decodeVersionDeltaStreamV2(content).toVersionDeltaStream()
```
languitar, line 538: Why do we have a different implementation pattern for the format here than for V1?

model-server-api/src/commonMain/kotlin/org/modelix/model/server/api/v2/VersionDeltaStreamV2.kt
```diff
@@ -0,0 +1,67 @@
...
+
+import io.ktor.http.ContentType
+import io.ktor.utils.io.ByteReadChannel
+import io.ktor.utils.io.ByteWriteChannel
+import io.ktor.utils.io.readUTF8Line
+import io.ktor.utils.io.writeStringUtf8
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.flow
+
+class VersionDeltaStreamV2(
```
languitar, line 27: Let's add some documentation to the format and a rationale why it is needed. Maybe, most of the implementation between v1 and v2 could also be shared apart from the final marker being sent.

model-server-api/src/commonTest/kotlin/org/modelix/model/server/api/v2/VersionDeltaStreamV2Test.kt
```diff
@@ -0,0 +1,170 @@
...
+import org.modelix.model.server.api.v2.VersionDeltaStreamV2.Companion.encodeVersionDeltaStreamV2
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertFailsWith
+import kotlin.test.assertTrue
+
+class VersionDeltaStreamV2Test {
+
+    @Test
+    fun parseStreamWithoutObjects() = runTest {
```
languitar, line 38:
```suggestion
    fun parsesStreamWithoutObjects() = runTest {
```

That's how we had the tests named mostly so far.

languitar (changes requested):
* This could be split into two feature commits, once introducing the new format in the model-server (with scope `model-server`) and one implementing support in the client.
* The OpenAPI spec needs to be updated.

model-server-api/src/commonMain/kotlin/org/modelix/model/server/api/v2/VersionDeltaStreamV2.kt
odzhychko: Sure. I will prepare an ADR for tomorrow's refinement.
It will be documentation and and also explain, why we use or do not reuse the old implementation.

languitar: I'd also be fine with a comment on that class ;-)

odzhychko: Ah, ok. I do not want to send the hashes like in the old version. So do not want to reuse the old implementation. I guess this is still something we should discuss in the refinement.

model-server-api/src/commonTest/kotlin/org/modelix/model/server/api/v2/VersionDeltaStreamV2Test.kt
```diff
@@ -0,0 +1,170 @@
...
+
+        val versionDeltaStream = decodeVersionDeltaStreamV2(channel)
+
+        assertEquals("CTVRw*a6KXJ4o7uzGlp-kUosxpyRf4fUpHnLokG9T86A", versionDeltaStream.versionHash)
+        val expectedObjects = listOf("L/100000017/xioDt*mnraICBf48DpWkvvtl2KuPixWn1p7yteYQ3XSg", "1/%00/0/%00///")
+        assertEquals(expectedObjects, versionDeltaStream.deltaObjects.toList())
+    }
+
+    @Test
+    fun failToParseAnySubstringOfAValidEncoding() = runTest {
```
odzhychko, line 71:
```suggestion
    fun failsToParseAnySubstringOfAValidEncoding() = runTest {
```

model-server-api/src/commonMain/kotlin/org/modelix/model/server/api/v2/VersionDeltaStreamV2.kt
languitar: I've also seen that regex once and was confused. So yes, if we can avoid the hashes, then that's fine for me. But I wouldn't have had to ask if this was documented somewhere ;-)

model-client/src/commonMain/kotlin/org/modelix/model/client2/ModelClientV2.kt
odzhychko: Because I wanted to implement encoding and decoding independant of the `ModelServer`/`ModelClient` to make it reusabel and testabel in isolation.
See `VersionDeltaStreamV2`

odzhychko:
This is still a draft. I will spend some time looking into `io.ktor.server.netty.NettyChannelInitializer.channelPipelineConfig`. Maybe it is possible to catch this exception and write invalid chunked data (e.g. a chunk with wrong lenght) to make clients recoginize an invalid response.

model-server-api/src/commonMain/kotlin/org/modelix/model/server/api/v2/VersionDeltaStreamV2.kt
slisson: Not sending the hashes is how it was implemented before. And then there was a bug in the JS library for computing SHA256 hashes that took me days to debug (see https://github.com/modelix/modelix.core/pull/213). Data integrity is more important than performance.

odzhychko: Good know. I will add it again.

odzhychko: Done. Added comment about rational for VersionDeltaStreamV2. Also mentioned why we still send hashes.

odzhychko:
> This is still a draft. I will spend some time looking into `io.ktor.server.netty.NettyChannelInitializer.channelPipelineConfig`. Maybe it is possible to catch this exception and write invalid chunked data (e.g. a chunk with wrong lenght) to make clients recoginize an invalid response.

The following channel handler could be used as a workaround for [KTOR-4862](https://youtrack.jetbrains.com/issue/KTOR-4862/Ktor-hangs-if-exception-occurs-during-write-response-body). But we would still be affected by [KTOR-6905](https://youtrack.jetbrains.com/issue/KTOR-6905/The-client-reads-incomplete-streamed-responses-without-failing).

I am not keen on maintaining this workaround. Especially because it might introduce other problems.

```
val configureNetty: NettyApplicationEngine.Configuration.() -> Unit = {
    this.channelPipelineConfig = {
        class OutboundHandlerClosingConnectionForIncompleteDeltaVersions : ChannelOutboundHandlerAdapter() {

            var isChunkedVersionDeltaStream = false
            var lastContent: Byte? = null
            override fun write(ctx: ChannelHandlerContext?, msg: Any?, promise: ChannelPromise?) {
                if (!isChunkedVersionDeltaStream && msg is DefaultHttpResponse) {
                    // TODO check for content-type of version delta stream
                    isChunkedVersionDeltaStream = msg.headers().containsValue("transfer-encoding", "chunked", false)
                }

                if (!isChunkedVersionDeltaStream) {
                    super.write(ctx, msg, promise)
                    return
                }
                if (msg !is HttpContent) {
                    super.write(ctx, msg, promise)
                    return
                }
                val capacity = msg.content().capacity()
                if (capacity > 0) {
                    lastContent = msg.content().getByte(capacity - 1)
                }
                if (msg !is LastHttpContent) {
                    super.write(ctx, msg, promise)
                    return
                }
                val magicByteToDetectProperEnding = '#'.toByte()
                if (lastContent != magicByteToDetectProperEnding) {
                    super.close(ctx, promise)
                } else {
                    try {
                        // TODO check if
                        val oldContent = msg.content()
                        // TODO check if `oldContent.capacity() - 1` does what is expected.
                        // Can `capacity()` be bigger than actual length?
                        val oldContentWithoutMagicByte = oldContent.copy(0, oldContent.capacity() - 1)
                        val newMsg = msg.replace(oldContentWithoutMagicByte)
                        super.write(ctx, newMsg, promise)
                    } finally {
                        ReferenceCountUtil.release(msg)
                    }
                }
            }
        }
        if (this.get("http1") != null) {
            this.addBefore(
                "http1",
                "closeConnectionForIncompleteDeltaVersions",
                OutboundHandlerClosingConnectionForIncompleteDeltaVersions()
            )
        }
    }
}
```

I think we should not add it, because it

model-server/src/main/kotlin/org/modelix/model/server/handlers/ModelReplicationServer.kt
```diff
@@ -365,12 +366,14 @@ class ModelReplicationServer(
 
     private suspend fun ApplicationCall.respondDelta(versionHash: String, baseVersionHash: String?) {
         val expectedTypes = request.acceptItems().map { ContentType.parse(it.value) }
-        return if (expectedTypes.any { it.match(VersionDeltaStream.CONTENT_TYPE) }) {
-            respondDeltaAsObjectStream(versionHash, baseVersionHash, false)
+        return if (expectedTypes.any { it.match(VersionDeltaStreamV2.CONTENT_TYPE) }) {
```
odzhychko, line 369: > The OpenAPI spec needs to be updated.

Update spec.

languitar (approved):

PR #651, opened 2024-03-27
Group ESLint updates in ts-model-api

odzhychko:


.pre-commit-config.yaml
```diff
@@ -39,13 +39,13 @@ repos:
...
-    rev: v8.50.0
+    rev: v8.56.0
     hooks:
       - id: eslint
         additional_dependencies:
           - "eslint@8.56.0"
           - "eslint-config-prettier@9.1.0"
-          - "@typescript-eslint/eslint-plugin@7.0.1"
-          - "@typescript-eslint/parser@7.0.1"
+          - "@typescript-eslint/eslint-plugin@7.4.0"
```
odzhychko, line 48: I am still unsure, if I want to run Prettier and ESLint with pre-commit.
(With Renovate we aim to have the diverging version under control automatically.)
Or if it is better just to run it through NPM and Gradle.
I currently tend to leave it in pre-commit because it fails faster and gives earlier feedback.

languitar:
This PR needs a manual rebase now.

languitar (approved):

PR #652, opened 2024-03-28
perf(model-server): do not block requests threads in `KeyValueLikeModelServer`

odzhychko:
This affects APIs like "/health", "/get/{key}", "/getRecursively/{key}", "/putAll" and "/put/{key}".

### DIFF
--- a/model-server/src/main/kotlin/org/modelix/model/server/handlers/KeyValueLikeModelServer.kt
+++ b/model-server/src/main/kotlin/org/modelix/model/server/handlers/KeyValueLikeModelServer.kt
@@ -27,6 +27,7 @@ import io.ktor.server.resources.put
 import io.ktor.server.response.respondText
 import io.ktor.server.routing.routing
 import io.ktor.util.pipeline.PipelineContext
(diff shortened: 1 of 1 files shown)


model-server/src/main/kotlin/org/modelix/model/server/handlers/KeyValueLikeModelServer.kt
```diff
@@ -199,20 +206,23 @@ class KeyValueLikeModelServer(val repositoriesManager: RepositoriesManager) {
...
-                        checkKeyPermission(key, EPermissionType.READ)
-                        keys.add(key)
-                    }
-                    val values = storeClient.getAll(keys)
-                    for (i in keys.indices) {
-                        val respEntry = JSONObject()
-                        respEntry.put("key", keys[i])
-                        respEntry.put("value", values[i])
-                        respJson.put(respEntry)
+                    val respJson = runBlocking(Dispatchers.IO) {
```
odzhychko, line 209: Seperate requesthandling, from non-blocking logic.

PR #660, opened 2024-04-02
MODELIX-846 Null-References lead to exception in data explorer

mhuster23:


languitar (changes requested):
Can we add a test for this?

mhuster23:
image
This is how the information is displayed now.

languitar (dismissed):

odzhychko (approved):

PR #665, opened 2024-04-03
build: remove action wagoid/commitlint-github-action

odzhychko:
Commitlint is executed by pre-commit and checked in the `linting.yml` workflow.

### DIFF
--- a/.github/workflows/dry-run-release.yml
+++ b/.github/workflows/dry-run-release.yml
@@ -3,15 +3,6 @@ on:
   pull_request:
 
 jobs:
-  lint-commits:
-    name: Lint PR commits
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v4
-        with:
-          fetch-depth: 0
-      - uses: wagoid/commitlint-github-action@v5
-
   test-release:
     name: Dry-run semantic-release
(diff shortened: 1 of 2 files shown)


languitar:
Hm, maybe we should leave that action: https://github.com/modelix/modelix.core/actions/runs/8540329213/job/23397105089?pr=665 the commitlint action in pre-commit is not run in CI, probably as it attaches to a different git commit hook phase or because pre-commit cannot know the range of commits to check.

odzhychko:
> Hm, maybe we should leave that action: https://github.com/modelix/modelix.core/actions/runs/8540329213/job/23397105089?pr=665 the commitlint action in pre-commit is not run in CI, probably as it attaches to a different git commit hook phase or because pre-commit cannot know the range of commits to check.

I am not sure what you mean.
Here are the logs of pre-commit for the run you linked to.
image

odzhychko:
> Hm, maybe we should leave that action: https://github.com/modelix/modelix.core/actions/runs/8540329213/job/23397105089?pr=665 the commitlint action in pre-commit is not run in CI, probably as it attaches to a different git commit hook phase or because pre-commit cannot know the range of commits to check.

Ah, ok.

PR #669, opened 2024-04-04
Cleanup `IRepositoriesManager`

odzhychko:


model-server/src/main/kotlin/org/modelix/model/server/handlers/IRepositoriesManager.kt
```diff
@@ -17,46 +17,77 @@
...
+import org.modelix.model.lazy.CLTree
 import org.modelix.model.lazy.CLVersion
 import org.modelix.model.lazy.RepositoryId
 
+/**
+ * Exposes high-level methods to interact with repositories, their branches and their versions.
+ *
+ * All methods are supposed to use non-blocking IO (or dispatch it appropriately).
+ */
 interface IRepositoriesManager {
```
odzhychko, line 29: > Now that I look at this again: Let's also add some documentation here to clarify the purpose of the interface and potential design decisions behind it.

odzhychko, line 29: > Method documentations should ideally give hints on which exceptions to be expected by each method.

model-server/src/main/kotlin/org/modelix/model/server/handlers/IRepositoriesManager.kt
```diff
@@ -17,46 +17,77 @@
...
+
     /**
-     * Used to retrieve the server ID. If needed, the server ID is created and stored.
-     *
-     * If a server ID was not created yet, it is generated and saved in the database.
-     * It gets stored under the current and all legacy database keys.
-     *
-     * If the server ID was created previously but is only stored under a legacy database key,
-     * it also gets stored under the current and all legacy database keys.
+     * Creates a repository or throws an [RepositoryAlreadyExistsException].
```
odzhychko, line 33: > This lacks some documentation now that we touch the code ;-)

> Why is the username needed here?
> Why can it be nullable?
> What is useRoleIds?

odzhychko (commented):
@languitar Here are your comments from #646

model-server/src/main/kotlin/org/modelix/model/server/ServerInitialization.kt
```diff
@@ -0,0 +1,40 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.modelix.model.server
+
+import org.modelix.model.server.store.IStoreClient
+import java.util.UUID
+
+class ServerInitialization(private val storeClient: IStoreClient) {
```
languitar, line 22: I think a few words on the intended concern of this class would be good. The name itself leaves a few question marks in my mind.

languitar, line 22: Any reason this is a class and not a free function?

model-server/src/main/kotlin/org/modelix/model/server/handlers/IRepositoriesManager.kt
languitar: Thanks, I think nowadays even the kotlin documentation has shifted towards using kdoc tags such as `@param` and `@throws` and I think we have mostly followed that approach in other places, too.

> [useRoleIds] is used when creating the [CLTree] for the initial version.

That still doesn't explain what that does actually.

model-server/src/main/kotlin/org/modelix/model/server/handlers/IRepositoriesManager.kt
```diff
@@ -17,46 +17,77 @@
...
     suspend fun getVersionHash(branch: BranchReference): String?
+
+    /**
+     * Returns the hash of the current version in this [branch] or null if the branch does not exist.
+     *
+     * The function executes immediately if the latest known version [lastKnown] is specified
+     * and the current version differs.
+     * Else the function delays the coroutine waiting for a new version on the [branch] and returns its hash.
+     * The delay is limited with a timeout after which the hash of the current version is returned
+     * even if no new version was created in the meantime.
```
languitar, line 68: We probably later have to think about this method again. Having a method that uses an unknown timeout is a bit weird as the interface.

model-server/src/main/kotlin/org/modelix/model/server/handlers/IRepositoriesManager.kt
```diff
@@ -17,46 +17,77 @@
...
+     *
+     * Throws [IllegalStateException] the branch does not exist or was removed while polling.
```
languitar, line 70:
```suggestion
     * Throws [IllegalStateException] if the branch does not exist or was removed while polling.
```

model-server/src/main/kotlin/org/modelix/model/server/handlers/RepositoryOverview.kt
```diff
@@ -27,21 +27,27 @@ import org.modelix.model.server.templates.PageWithMenuBar
 
 class RepositoryOverview(private val repoManager: IRepositoriesManager) {
 
+    class RepositoryIdWithBranchNames(val repositoryId: String, val branchNames: List<String>)
```
languitar, line 30: In IRepositoriesManager, branches are sets. Is order important or could we limit this to a set, too?

model-server/src/main/kotlin/org/modelix/model/server/store/IStoreClient.kt
```diff
@@ -38,8 +38,24 @@ interface IStoreClient : AutoCloseable {
     fun <T> runTransaction(body: () -> T): T
 }
 
-suspend fun <T> IStoreClient.runTransactionSuspendable(body: () -> T): T {
-    return withContext(Dispatchers.IO) { runTransaction(body) }
+val isInSuspendableTransactionMarker: ThreadLocal<Boolean> = ThreadLocal()
```
languitar, line 41: Can this be private?

model-server/src/test/kotlin/org/modelix/model/server/StoreClientTest.kt
```diff
@@ -143,4 +145,35 @@ abstract class StoreClientTest(val store: IStoreClient) {
 
         assertEquals(setOf<String?>(value3), valuesSeenByListener)
     }
+
+    @Test
+    fun nestedSuspendableTransactionAreOnlyDispatchedOnce() = runTest {
+        // Arrange
```
mhuster23, line 151: Small nitpick: While I think it's great to use the Arrange-Act-Assert structure, I feel like adding these comments each time just produces noise. I'd prefer just visibly separating the parts with new lines.

model-server/src/main/kotlin/org/modelix/model/server/handlers/IRepositoriesManager.kt
odzhychko: Done

model-server/src/main/kotlin/org/modelix/model/server/handlers/IRepositoriesManager.kt
odzhychko: Done (using Block Tags)
Done (explaining useRoleIds)

model-server/src/main/kotlin/org/modelix/model/server/ServerInitialization.kt
odzhychko: Obsolete. Class removed.

model-server/src/main/kotlin/org/modelix/model/server/handlers/IRepositoriesManager.kt
odzhychko: It is tricky. But `RepositoriesManager` is still just an internal and model server specific class, so this should be fine for now.

model-server/src/main/kotlin/org/modelix/model/server/handlers/IRepositoriesManager.kt
```diff
@@ -17,46 +17,77 @@
...
+    /**
+     * Merges the changes from [newVersionHash] into the current version in [branch].
+     * Returns the version hash of the merge version.
+     */
     suspend fun mergeChanges(branch: BranchReference, newVersionHash: String): String
 
     /**
-     * Same as [mergeChanges] but blocking.
-     * Caller is expected to execute it outside the request thread.
+     * Returns object data as [ObjectData] that does exist in the version specified by [versionHash]
```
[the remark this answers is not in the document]
odzhychko: Done

model-server/src/main/kotlin/org/modelix/model/server/handlers/RepositoryOverview.kt
odzhychko: Done

model-server/src/main/kotlin/org/modelix/model/server/store/IStoreClient.kt
odzhychko: Done

model-server/src/main/kotlin/org/modelix/model/server/ServerInitialization.kt
odzhychko: Done. Made it a free function.

I had originally intended the class to have more fields and to grow to include more initialisation and migration code. But for now a free function is more appropriate.

slisson (changes requested):
I think introducing the `runTransactionSuspendable` method was a mistake. This PR introduces a lot of `runBlocking` call, but I think we should do the opposite and try the avoid usages of it. Having any suspendable methods in `RepositoryManager` seems unnecessary. We have to find a different solution for the problem of the starving Ktor thread pool, all these unnecessary supendable functions is not a good one.

odzhychko:
This is WIP because it needs to redone after #691 is merged.
Also the methods on `IRepositoryManager` should not be suspendable.

slisson:
Outdated

PR #670, opened 2024-04-04
MODELIX-794: adapt the detekt configuration according to the discussions from the coding conventions workshop

languitar:
Updates detekt to:

* lint for missing docs on public members
* lint for usage of !! to prefer checkNotNull or requireNotNull

I've also fixed a bug in CI where multiplatform code was not linted due to the `detekt` target not including multiplatform targets.


detekt-rules/src/main/kotlin/org/modelix/detekt/AvoidNonNullAssertionOperator.kt
```diff
@@ -0,0 +1,50 @@
...
+import io.gitlab.arturbosch.detekt.api.CodeSmell
+import io.gitlab.arturbosch.detekt.api.Config
+import io.gitlab.arturbosch.detekt.api.Debt
+import io.gitlab.arturbosch.detekt.api.Entity
+import io.gitlab.arturbosch.detekt.api.Issue
+import io.gitlab.arturbosch.detekt.api.Rule
+import io.gitlab.arturbosch.detekt.api.Severity
+import org.jetbrains.kotlin.psi.KtReferenceExpression
+
+class AvoidNonNullAssertionOperator(config: Config) : Rule(config) {
```
odzhychko, line 28: Interesting how custom rules can be added.

But according to https://github.com/detekt/detekt/issues/4832 it can be replaced by `UnnecessaryNotNullOperator` and `UnsafeCallOnNullableType`

model-datastructure/src/commonTest/kotlin/LinearHistoryTest.kt
```diff
@@ -57,133 +57,6 @@ class LinearHistoryTest {
         assertEquals(expected, actual)
     }
 
-    @Test
-    fun knownPerformanceIssue() {
-        // This test was dumped from actual case discovered during a profiling session.
-
-        val v30000003a = version(12884901946, null)
```
odzhychko, line 64: > The `knownPerformanceIssue` test case was now constantly timing out when
being executed in JS. After clarification, all corner cases it included
are already covered by the **remaining tests in the test file**. Therefore,
I have dropped this test case to avoid the timeout and to reduce the
amount of hard to maintain test cases.

Actually not all tests are in this file.
Other relevant test include:
* MergeOrderTest
* ConflictResolutionTest
* UndoTest

See https://github.com/modelix/modelix.core/pull/354

odzhychko, line 64: If you mention the timeouts in the commit message, I would add that they were not caused by the production code, but by the expensive asserts in `assertHistoryIsCorrect`.

Someone might wonder why we did not look into why the test took so long.

odzhychko (changes requested):

detekt-rules/src/main/kotlin/org/modelix/detekt/AvoidNonNullAssertionOperator.kt
languitar: UnsafeCallOnNullableType is active by default but doesn't catch all of the cases the custom rule is reporting on. For whatever reason....

model-datastructure/src/commonTest/kotlin/LinearHistoryTest.kt
languitar: I've reworded the commit message to include the tests you mentioned.

languitar: I've reworded the commit message to include this hint.

odzhychko (approved):

detekt-rules/src/main/kotlin/org/modelix/detekt/AvoidNonNullAssertionOperator.kt
languitar: I just tried it again, even with actively enabling those two rules, despite the fact that they should be active by default, I can't get a single report from them. Not sure if that's a detekt bug.

odzhychko: Thist might be realted to type resolution.

See https://github.com/detekt/detekt/issues/2036#issuecomment-543492134
See https://detekt.github.io/detekt/type-resolution.html

PR #672, opened 2024-04-05
MODELIX-740 bulk-model-sync should not cache pull result from model-server

mhuster23:


bulk-model-sync-gradle/src/main/kotlin/org/modelix/model/sync/bulk/gradle/ModelSyncGradlePlugin.kt
```diff
@@ -95,25 +98,45 @@ class ModelSyncGradlePlugin : Plugin<Project> {
         jsonDir: File,
     ): TaskProvider<*> {
         val serverSource = syncDirection.source as ServerSource
-
         val name = "${syncDirection.name}ExportFromModelServer"
+
         val exportFromModelServer = project.tasks.register(name, ExportFromModelServer::class.java) {
```
odzhychko, line 103: My main concern is to understand, when the body of `register` is called. 
It is called lazily at least once when it is executed. But when is it called again?

I am still looking into this.

odzhychko (commented):
I have some small suggestions. And need to look further into how `project.tasks.register` works.

bulk-model-sync-gradle/src/main/kotlin/org/modelix/model/sync/bulk/gradle/ModelSyncGradlePlugin.kt
mhuster23: As far as I understand it happens during the [configuration phase](https://docs.gradle.org/current/userguide/build_lifecycle.html#sec:configuration).
So there is a tiny window between configuration and execution, where a new version could be created on the model-server that we would not pull but that's probably negligible.

odzhychko: Ah, so it would break if the [Configuration Cache](https://docs.gradle.org/current/userguide/configuration_cache.html) is used.

Do you know why the `ExportFromModelServer` was up-to-date anyway? I thought that it would not be cached by default because it extends `DefaultTask`.

odzhychko: I am still mixed up by up-to-date checks and chaching.
But the proper solution should be to say, that the task is never up-to-date when an branch is specified.
`it.outputs.upToDateWhen { serverSource.branchName != null }`

odzhychko: So incremental builds and caching are different things.
See https://docs.gradle.org/current/userguide/gradle_optimizations.html
See https://stackoverflow.com/questions/65101472/what-is-the-difference-between-from-cache-and-up-to-date-in-gradle

So we could have a fancy solution (maybe out-of-scope) like this:

```java
it.outputs.upToDateWhen { task ->
                if (it.branchName == null) {
                    val revisionChanged = TODO("check if revision changed since last run")
                    revisionChanged
                } else {
                    true
                }
            }
```

mhuster23: Yeah the issue title is misleading because this is about the up-to-date check and not the build cache.
I haven't thought about the interaction with the Configuration Cache though, that is problematic.

We shouldn't have to keep track of the revision change ourselves. I'll think of a different solution.

odzhychko: Pulling the revision in the configureation phase still seems wrong to me. Even when working around the configuration cache issue using `it.notCompatibleWithConfigurationCache`.

But I am no expert and it works. So it is fine.

odzhychko (approved):

bulk-model-sync-gradle/src/main/kotlin/org/modelix/model/sync/bulk/gradle/ModelSyncGradlePlugin.kt
mhuster23: If we want to avoid pulling during configuration, we could also have a new task that pulls and writes the revision to a file in the buildDir. 
The ExportFromModelServerTask would depend on that new task and then use that file as an input.

odzhychko (approved):

PR #677, opened 2024-04-08
fix(bulk-model-sync): update previously unresolvable references if they can be resolved

mhuster23:
For unresolvable references we store the serialized expected reference, as we assume the node is not part of the model but the reference might still be useful for dynamic resolution on access. 
If the target node became part of the model later on, we did not update it because the originalIds still matched. Now we check, if it was previously unresolvable and update the reference to the INode accordingly.

### DIFF
--- a/bulk-model-sync-lib/src/commonMain/kotlin/org/modelix/model/sync/bulk/ModelImporter.kt
+++ b/bulk-model-sync-lib/src/commonMain/kotlin/org/modelix/model/sync/bulk/ModelImporter.kt
@@ -88,12 +88,15 @@ class ModelImporter(
 
     private fun PostponedReference.setPostponedReference() {
         val expectedRefTarget = originalIdToExisting[expectedTargetId]
-        if (expectedRefTarget == null) {
+        if (expectedRefTarget != null) {
+            node.setReferenceTarget(role, expectedRefTarget)
+            return
+        }
+        // Ensure unresolvable reference wasn't stored before
+        if (node.getReferenceTargetRef(role)?.serialize() != expectedTargetId) {
             // The target node is not part of the model. Assuming it exists in some other model we can
             // store the reference and try to resolve it dynamically on access.
             node.setReferenceTarget(role, SerializedNodeReference(expectedTargetId))
-        } else {
-            node.setReferenceTarget(role, expectedRefTarget)
         }
     }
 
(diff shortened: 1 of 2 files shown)


bulk-model-sync-lib/src/commonMain/kotlin/org/modelix/model/sync/bulk/ModelImporter.kt
```diff
@@ -294,15 +294,14 @@ class ModelImporter(
 
     private fun syncReferences(node: INode, nodeData: NodeData) {
         nodeData.references.forEach {
```
odzhychko, line 296: tip: you can use destructuring like this: `nodeData.references.forEach { (key, value) ->`

bulk-model-sync-lib/src/commonMain/kotlin/org/modelix/model/sync/bulk/ModelImporter.kt
```diff
@@ -311,6 +310,14 @@ class ModelImporter(
             node.setReferenceTarget(it, null as INodeReference?)
         }
     }
+
+    private fun referenceNeedsUpdate(node: INode, referenceRole: String, expectedTargetRef: String): Boolean {
+        val actualTarget = node.getReferenceTarget(referenceRole)
+        val actualTargetRef = actualTarget?.originalId() ?: node.getReferenceTargetRef(referenceRole)?.serialize()
+
+        return actualTargetRef != expectedTargetRef ||
+            actualTarget == null // previously unresolvable reference can now be resolved
```
odzhychko, line 319: If `actualTarget == null` you can return early and can skip calling `actualTarget?.originalId() ?: node.getReferenceTargetRef(referenceRole)?.serialize()`.

Also, it would be more accurate to say `// Previously unresolvable reference might become resolvable. This is the case when a referenced node is added during the sync.`

odzhychko (changes requested):

bulk-model-sync-lib/src/commonMain/kotlin/org/modelix/model/sync/bulk/ModelImporter.kt
mhuster23: Your comment made me realize that we still have to check whether it's actually resolvable. With the current implementation we would set an unchanged unresolvable reference to the same unresolvable reference. 
Therefore we would use more operations than we should have.

mhuster23: added the additonal check and tested the number of used operations

odzhychko (approved):

PR #686, opened 2024-04-09
feat(model-server): add endpoint /about

odzhychko:
Show the server version in the endpoint.
Link to the endpoint in the title bar.

### DIFF
--- a/build-logic/src/main/kotlin/org/modelix/GenerateVersion.kt
+++ b/build-logic/src/main/kotlin/org/modelix/GenerateVersion.kt
@@ -1,0 +1,79 @@
+/*
+ * Copyright (c) 2024.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.modelix
+
+import org.gradle.api.Project
+import org.gradle.kotlin.dsl.configure
+import org.gradle.kotlin.dsl.get
+import org.gradle.kotlin.dsl.withType
+import org.jetbrains.kotlin.gradle.dsl.KotlinJvmProjectExtension
+import org.jetbrains.kotlin.gradle.dsl.KotlinMultiplatformExtension
+import org.jetbrains.kotlin.gradle.plugin.KotlinMultiplatformPluginWrapper
+import org.jetbrains.kotlin.gradle.plugin.KotlinPluginWrapper
+
+/**
+ * Causes code containing the Modelix version to be generated and used in a Kotlin project.
+ * The Kotlin project can be a JVM or Multiplatform project.
+ */
+fun Project.registerVersionGenerationTask(packageName: String) {
+    val packagePath = packageName.replace('.', '/')
+
+    val generateVersionVariable = tasks.register("generateVersionVariable") {
+        doLast {
+            val fileContent = buildString {
+                appendLine("package $packageName")
+                appendLine()
+                appendLine("""const val MODELIX_VERSION: String = "$version"""")
+                if (project.name == "modelql-core") {
+                    appendLine("""@Deprecated("Use the new MODELIX_VERSION", replaceWith = ReplaceWith("MODELIX_VERSION"))""")
+                    appendLine("const val modelqlVersion: String = MODELIX_VERSION")
+                }
+            }
+            val outputDir = layout.buildDirectory.dir("version_gen/$packagePath").get().asFile
+            outputDir.mkdirs()
+            outputDir.resolve("Version.kt").writeText(fileContent)
+        }
+    }
+
+    // Generate version constant for Kotlin JVM projects
+    plugins.withType<KotlinPluginWrapper> {
+        extensions.configure<KotlinJvmProjectExtension> {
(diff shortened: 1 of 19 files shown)


model-api/build.gradle.kts
```diff
@@ -58,6 +59,25 @@ kotlin {
     }
 }
 
+val generateVersionVariable by tasks.creating {
```
mhuster23, line 62:
```suggestion
val generateVersionVariable by tasks.registering {
```
Prefer using `registering`, so that the [configuration can be deferred](https://docs.gradle.org/current/userguide/task_configuration_avoidance.html).

model-api/build.gradle.kts
```diff
@@ -58,6 +59,25 @@ kotlin {
...
+        outputDir.resolve("Version.kt").writeText(
+            """
+            package org.modelix.model.api
+
+            const val MODELIX_VERSION: String = "$version"
+
+            """.trimIndent(),
+        )
+    }
+}
```
mhuster23, line 75: Putting this in the model-api project might be problematic, if we ever decide to start versioning model-api and model-server independently from each other. Should we move it to the model-server and call it sth like `MODELIX_SERVER_VERSION`?

model-server/src/main/kotlin/org/modelix/model/server/handlers/AboutInformationHandler.kt
```diff
@@ -0,0 +1,39 @@
...
+import io.ktor.server.application.Application
+import io.ktor.server.application.call
+import io.ktor.server.resources.get
+import io.ktor.server.response.respond
+import io.ktor.server.routing.routing
+import org.modelix.api.public.AboutInformation
+import org.modelix.api.public.Paths
+import org.modelix.model.api.MODELIX_VERSION
+
+object AboutInformationHandler {
```
mhuster23, line 28: What's the idea behind this being a singleton? The other handlers are regular classes.

mhuster23 (changes requested):

model-api/build.gradle.kts
odzhychko: I will add the generator programmatically to each sub-project.
Importing the version might be problematic if gradle resolves versions differently.

model-server/src/main/kotlin/org/modelix/model/server/handlers/AboutInformationHandler.kt
odzhychko: We do not need class properties, so we do not need a class. Therefore I keept it simple.

model-server-openapi/specifications/model-server.yaml
```diff
@@ -48,6 +48,12 @@ paths:
           $ref: '#/components/responses/200'
         "500":
           $ref: '#/components/responses/500'
+  /about:
```
odzhychko, line 51: Add it to model-server-operative.yaml

build.gradle.kts
```diff
@@ -170,6 +171,64 @@ subprojects {
     }
 }
 
+val byProjectNamePackageForVersionVariable: Map<String, String> = mapOf(
+    "model-server" to "org.modelix.model.server",
+    "modelql-core" to "org.modelix.modelql.core",
+)
```
languitar, line 177: Can't we localize the logic more in the (two) modules that currently need it? You could define a utility task to generate such a version file and the two modules then call that task, passing in their required package name.

model-server-openapi/specifications/model-server-operative.yaml
```diff
@@ -34,3 +43,11 @@ components:
         text/plain:
           schema:
             type: string
+  schemas:
+    AboutInformation:
+      type: object
+      properties:
+        version:
+          type: string
```
languitar, line 51: We could further specify the format with a regex.

odzhychko: We could do it. 
There is even a recommended Regex for [SemVer](https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string). 
But currently I see no benefit in specifying the pattern of the version.

I would rather not specify it yet.
This would entail doing additional checks in the Gradle build.
And maybe under some circumstances, we will want a version that does not match it.

languitar:
What's the state of this PR?

odzhychko:
> What's the state of this PR?

[This comment](https://github.com/modelix/modelix.core/pull/686#discussion_r1570265827) is valid and open.
I did not work on this PR because it has a low priority in comparison to other tasks I am working on.

build.gradle.kts
odzhychko: I currently do not know how to do this and will need to take some time to read up on Gradle.

odzhychko: Done

build-logic/src/main/kotlin/org/modelix/GenerateVersion.kt
```diff
@@ -0,0 +1,79 @@
...
+import org.gradle.kotlin.dsl.get
+import org.gradle.kotlin.dsl.withType
+import org.jetbrains.kotlin.gradle.dsl.KotlinJvmProjectExtension
+import org.jetbrains.kotlin.gradle.dsl.KotlinMultiplatformExtension
+import org.jetbrains.kotlin.gradle.plugin.KotlinMultiplatformPluginWrapper
+import org.jetbrains.kotlin.gradle.plugin.KotlinPluginWrapper
+
+/**
+ * Causes code containing the Modelix version to be generated and used in a Kolin project.
+ * The Kotlin project can be a JVM oder Multiplatform.
```
mhuster23, line 30:
```suggestion
 * Causes code containing the Modelix version to be generated and used in a Kotlin project.
 * The Kotlin project can be a JVM or Multiplatform project.
```

mhuster23 (approved):

model-server-openapi/specifications/model-server-operative.yaml
odzhychko: @languitar I will merge the commit as is with the type "string".
I will come back to you to talk about, whether it makes sens to make it stricter with an Regex.

PR #691, opened 2024-04-12
MODELIX-808 Change data storage to efficiently query whether a hash exists in a repository

slisson:
The versions of models where previously stored in the same storage
without knowing to which repository they belong.
To allow proper permission checks there is now an additional
`repository` field in the database.

Existing repositories remain in the global storage and don't benefit
from this new security feature. A migration is not yet implemented.

New repositories are created in the new storage, but are not accessible
via the deprecated `RestWebModelClient`.


languitar:
> Existing repositories remain in the global storage and don't benefit
from this new security feature. A migration is not yet implemented.

What's the rationale for this decision?

slisson:
> What's the rationale for this decision?

The deprecated API between `KeyValueLikeModelServer` and `RestWebModelClient` doesn't know anything about repositories and from individual entries we cannot derive to which repository they belong. That means, a model stored inside a repository specific storage cannot be accessed via that API. It's only possible to use that new feature for a repository if it doesn't need to be accessible by the old client.

model-server/src/main/kotlin/org/modelix/model/server/store/IgniteStoreClient.kt
```diff
@@ -14,29 +14,35 @@
...
+class IgniteStoreClient(jdbcConfFile: File? = null, inmemory: Boolean = false) : IGenericStoreClient<ObjectInRepository>, AutoCloseable {
 
     companion object {
         private const val ENTRY_CHANGED_TOPIC = "entryChanged"
     }
 
     private lateinit var ignite: Ignite
-    private val cache: IgniteCache<String, String?>
+    private val legacyCache: IgniteCache<String, String?>
+    private val mutableObjectsCache: IgniteCache<ObjectInRepository, String?>
```
odzhychko, line 44: idea: If we copy over all keys from `legacyCache` as global keys into `mutableObjectsCache` at startup (aka. migration phase), then we could get rid of `legacyCache`.

This is already done by only writing the newly created gloabal immutable keys only to the `mutableObjectsCache`.

See `igniteStoreClient.putAll`

I think simpler code and not having the keys scattared around multiple caches would be easier to understand and debug.

odzhychko, line 44: `mutableObjectsCache` is currently a missleaing name.
It actuall might contain:

- global and repository-scoped mutable keys
- global **immutable** keys

See `igniteStoreClient.putAll`

model-server/src/main/kotlin/org/modelix/model/server/store/IStoreClient.kt
```diff
@@ -20,29 +20,164 @@ import kotlinx.coroutines.coroutineScope
...
+
+class RepositoryScopedStoreClient(private val repositoryId: RepositoryId, client: IGenericStoreClient<ObjectInRepository>) : StoreClientAdapter(client) {
+    override fun getRepositoryId() = repositoryId
+}
+
+class GlobalStoreClient(client: IGenericStoreClient<ObjectInRepository>) : StoreClientAdapter(client) {
+    override fun getRepositoryId() = null
+}
+
+class ContextScopedStoreClient(client: IGenericStoreClient<ObjectInRepository>) : StoreClientAdapter(client) {
```
odzhychko, line 66: thought: I see the use of a context scope as accidental complexity.
I also see why it helps. It reduces the need for code rewrite to introduce this feature.

In the future, we could get rid of it by letting by breaking up the `IgniteStoreClient` to into two parts. One for model data and one for metadata (e.g., repositories, branches, cliendIds, serverIds etc.) and use them everywhere without mixing them together.

I am not too concerned about it because refactoring to get rid of `ContextScopedStoreClient` is possible without changing the data schema.

odzhychko (changes requested):

model-server-test/docker/initdb.sql
```diff
@@ -0,0 +1,251 @@
...
+    ('xQ_tk*pohMZohAA8ChoDUZotsSat0R5k5hSZPvwlfock', 'L/100000020/t2K0Z*2fRWwaHgE5ZJOGpbZZONBEmBpkU9yfROLSZD4M'),
+    ('xbcAG*o9JthBYkCcGFdSTvy6ygQXDTCsM4SEIHx0x09w', 'info/2/CIiES*UIqiJi7WiJCasu9yMw1nO3XzRxhvX9tWbLp01A'),
+    ('xioDt*mnraICBf48DpWkvvtl2KuPixWn1p7yteYQ3XSg', '100000017/mps%3A96533389-8d4c-46f2-b150-8d89155f7fca%2F2756110869689139961/100000016/equipment//%23mpsNodeId%23=8493110207576779271,equipment=2oZF3J5qvts%2FProjector/'),
+    ('xo8rS*7XCggPvvFZG2BhzrA3NJ-YtNdy3QucgQdNyjek', 'L/40000001c/MVmqr*dBb1Y6o0JLTKGrVOVdv6qKo82NtRi9qqEttmDs'),
+    ('yV6G2*hwGekzBG30vT5Cy43XRweqkgRM-uOCM_iKZZTI', '10000000d/mps%3A0a7577d1-d4e5-431d-98b1-fae38f9aee80%2F474657388638618895/1/modules/10000000e,10000000f,100000010,100000011,100000012,100000013/compileInMPS=true,id=a04444a1-c8a4-48e8-a940-32a70d0e8bfc,moduleVersion=0,name=University.Schedule.modelserver.backend.sandbox/'),
+    ('zK4Y2*xIEWlYlQGJL2Va4Z0ESgpWgnSQcOmnPeqt34PA', 'S/5/0/g4pFH*AQymlg23Mueo-AyIJU6n3fzGCwhSwAyOIncuFw'),
+    ('zxgZN*oLuudxsu42ppSEGnCib8LkrSvauQk2B6T7AW6o', 'courses/2/qJszC*kqdafBvlOAUkHASuNGyKiYijrmdPmA752gdjcY');
```
languitar, line 251: Why is this included here?

model-client/src/commonMain/kotlin/org/modelix/model/client2/ModelClientV2.kt
```diff
@@ -79,14 +82,20 @@ class ModelClientV2(
...
 
     suspend fun init() {
         updateClientId()
         updateUserId()
     }
 
+    private fun getStore(repository: RepositoryId) = runSynchronized(storeForRepository) { storeForRepository.getOrPut(repository) { ObjectStoreCache(MapBasedStore()) } }
+    private fun getRepository(store: IDeserializingKeyValueStore): RepositoryId {
+        return storeForRepository.asSequence().first { it.value == store }.key
+    }
```
languitar, line 97: How often will this be called? Due to the inverse operation, this cannot make use of an efficient lookup.

model-server-openapi/specifications/model-server-html.yaml
```diff
@@ -37,7 +37,7 @@ paths:
               examples:
                 Example#1:
                   value: "../repos/"
-  /content/{repository}/{branch}/latest:
+  /content/repositories/{repository}/branches/{branch}/latest:
```
languitar, line 40: These changes should be marked as breaking changes.

model-server-test/README.md
languitar, line 1: Can we move the test into a separate commit?

model-server-test/build.gradle.kts
```diff
@@ -0,0 +1,40 @@
...
+            File(it).exists()
+        }?.let { docker ->
+            executable.set(docker)
+        }
+        listOf("/usr/bin/docker", "/usr/local/bin/docker").firstOrNull {
+            File(it).exists()
+        }?.let { docker ->
+            dockerExecutable.set(docker)
+        }
+    }
```
languitar, line 39: What about a gradle property that users can use to give their non-standard location of `docker-compose` if `PATH` is not an option?

model-server-test/docker-compose.yaml
```diff
@@ -0,0 +1,11 @@
+services:
+  db:
+    build:
+      dockerfile: Dockerfile
+      context: docker
```
languitar, line 5: I think we can get rid of the build step, which only takes time. The `initdb.sql` could simply be mounted to `/docker-entrypoint-initdb.d/initdb.sql` using a volume mount here in the compose file with the native postgres image.

model-server-test/docker/Dockerfile
```diff
@@ -0,0 +1,3 @@
+FROM postgres:9.6
```
languitar, line 1: Any reason for using such an old postgres version? Looks like we should be at least on version 12: https://endoflife.date/postgresql

model-server-test/docker/initdb.sql
languitar, line 1: Do we really need a copy of the init file here? Maybe at least a symlink?

model-server/src/main/kotlin/org/modelix/model/server/SqlUtils.kt
```diff
@@ -40,12 +40,15 @@
...
             stmt.execute(creationSql)
         }
-        val stmt = connection.createStatement()
-        stmt.execute(
-            "GRANT ALL ON TABLE $schemaName.$tableName TO $username;",
-        )
+        try {
+            val stmt = connection.createStatement()
+            stmt.execute("GRANT ALL ON TABLE $schemaName.$tableName TO $username;")
+        } catch (ex: Exception) {
```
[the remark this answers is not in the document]
languitar: Isn't there a more specific exception to handle?

model-server/src/main/kotlin/org/modelix/model/server/SqlUtils.kt
```diff
@@ -78,13 +85,41 @@ internal class SqlUtils(private val connection: Connection) {
...
                         CONSTRAINT kv_pkey PRIMARY KEY (key)
                     );
                 """,
             )
+            ensureTableIsPresent(
+                schemaName,
+                userName,
+                "immutable_objects",
+                """
+                    CREATE TABLE $schemaName.immutable_objects (
```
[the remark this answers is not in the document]
languitar: Cf. https://issues.modelix.org/issue/MODELIX-857/Add-migrations-for-data-in-database

model-server/src/main/kotlin/org/modelix/model/server/handlers/IRepositoriesManager.kt
```diff
@@ -55,6 +57,17 @@ interface IRepositoriesManager {
...
     suspend fun computeDelta(versionHash: String, baseVersionHash: String?): ObjectData
+
+    suspend fun <R> runWithRepository(repository: RepositoryId, body: suspend () -> R): R
+    fun findObject(hash: String): List<ObjectInRepository>
+
+    /**
+     * The data of a repository is stored separately from other repositories, but that wasn't always the case.
+     * For backwards compatibility, existing repositories remain in the global storage.
+     *
+     * @return null if the repository doesn't exist
```
languitar, line 68: I'm not sure if I like this use of nullability. Other opinions on that?

model-server/src/main/kotlin/org/modelix/model/server/store/IStoreClient.kt
```diff
@@ -20,29 +20,164 @@ import kotlinx.coroutines.coroutineScope
...
+fun <R> IStoreClient.withGlobalRepository(body: () -> R): R = withRepository(null, body)
+fun <R> IStoreClient.withRepository(repository: RepositoryId?, body: () -> R): R {
+    assert(this is ContextScopedStoreClient || this is StoreClientAdapter && this.getRepositoryId() == repository) {
+        "Store is not context scoped: $this"
+    }
+    return ContextScopedStoreClient.CONTEXT_REPOSITORY.computeWith(repository, body)
+}
+suspend fun <R> IStoreClient.withGlobalRepositoryS(body: suspend () -> R): R = withRepositoryS(null, body)
+
+suspend fun <R> IStoreClient.withRepositoryS(repository: RepositoryId?, body: suspend () -> R): R {
```
languitar, line 157: What is that `S` suffix? Do we have that convention used somewhere else already?

model-server/src/main/kotlin/org/modelix/model/server/store/IStoreClient.kt
```diff
@@ -104,11 +245,15 @@ fun IStoreClient.loadDump(file: File): Int {
...
 @Throws(IOException::class)
-fun IStoreClient.writeDump(file: File) {
+fun IGenericStoreClient<ObjectInRepository>.writeDump(file: File) {
     file.writer().use { writer ->
         for ((key, value) in getAll()) {
             if (value == null) continue
-            writer.append(key)
+            writer.append(key.key)
+            if (!key.isGlobal()) {
+                writer.append("@@")
```
languitar, line 254: Constant for `@@` would be good.

model-server/src/main/kotlin/org/modelix/model/server/store/ObjectInRepository.kt
```diff
@@ -0,0 +1,35 @@
...
+    fun isGlobal() = repositoryId == ""
+    fun isImmutable() = HashUtil.isSha256(key)
+    fun getRepositoryId() = repositoryId.takeIf { it != "" }
+
+    companion object {
+        /**
+         * Empty string is used for entries that don't belong to a repository, because PostgreSQL doesn't allow null
+         * values in primary keys.
+         */
+        fun global(key: String) = ObjectInRepository("", key)
```
languitar, line 31: Can't we hide this implementation detail on some level below?

model-server-test/build.gradle.kts
```diff
@@ -0,0 +1,40 @@
...
+
+dependencies {
+    testImplementation(kotlin("test"))
+    testImplementation(project(":model-server"))
+}
+
+tasks.test {
+    doFirst {
+        val db = dockerCompose.servicesInfos.getValue("db")
+        systemProperty("jdbc.url", "jdbc:postgresql://${db.host}:${db.port}/")
```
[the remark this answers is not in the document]
slisson: We would still have to use system properties to pass the correct port number from Gradle to the test code.

model-server/src/main/kotlin/org/modelix/model/server/handlers/DeprecatedLightModelServer.kt
```diff
@@ -144,95 +142,97 @@ class DeprecatedLightModelServer(val client: LocalModelClient) {
...
-                userId,
-                newTree,
-                null,
-                emptyArray(),
-            )
-            client.asyncStore!!.put(repositoryId.getBranchKey(), newVersion.hash)
-            respondVersion(newVersion)
+            repositoriesManager.runWithRepository(repositoryId) {
+                val userId = call.getUserName()
+                val newVersion = repositoriesManager.createRepository(repositoryId, userId)
```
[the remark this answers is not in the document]
slisson: I don't really care. This whole API is deprecated and unused. Just left it there because the unit tests indirectly test all the other components. This is still useful at least for this migration.

model-server-openapi/specifications/model-server-html.yaml
slisson: These endpoints are only used by the content explorer which is part of the model-server. I don't know why there is an Open API specification of it. I don't consider this a breaking change.

languitar: oh yes, missed that this is the content explorer file, which should be deleted in any case.

model-server-test/README.md
slisson: Done.

model-server/src/main/kotlin/org/modelix/model/server/store/IStoreClient.kt
slisson: I think splitting the StoreClient is a more complex solution. That moves responsibilities about transaction management out of the StoreClient to users of it.

model-server/src/main/kotlin/org/modelix/model/server/store/IgniteStoreClient.kt
slisson: I reverted back to a single cache and decided that this is out of scope for this PR and MODELIX-866 should be fixed in a separate PR.

slisson: I reverted back to a single cache. All data remains in the single existing table. No copying necessary.

model-server/src/main/kotlin/org/modelix/model/server/handlers/IRepositoriesManager.kt
```diff
@@ -55,6 +57,17 @@ interface IRepositoriesManager {
      */
     fun mergeChangesBlocking(branch: BranchReference, newVersionHash: String): String
     suspend fun computeDelta(versionHash: String, baseVersionHash: String?): ObjectData
+
+    suspend fun <R> runWithRepository(repository: RepositoryId, body: suspend () -> R): R
+    fun findObject(hash: String): List<ObjectInRepository>
```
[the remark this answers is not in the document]
slisson: Removed the dead code.

model-server/src/main/kotlin/org/modelix/model/server/handlers/DeprecatedLightModelServer.kt
```diff
@@ -144,95 +142,97 @@ class DeprecatedLightModelServer(val client: LocalModelClient) {
...
             val updateData = JSONArray(call.receiveText())
             val repositoryId = RepositoryId(call.parameters["repositoryId"]!!)
-            val baseVersionHash = call.parameters["versionHash"]!!
-            val baseVersionData = getStore().get(baseVersionHash, { CPVersion.deserialize(it) })
-            if (baseVersionData == null) {
-                call.respond(HttpStatusCode.NotFound, "version not found: $baseVersionHash")
-                return@post
+            repositoriesManager.runWithRepository(repositoryId) {
+                val baseVersionHash = call.parameters["versionHash"]!!
+                val baseVersion = repositoriesManager.getVersion(repositoryId, baseVersionHash)
```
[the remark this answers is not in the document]
slisson: Changed the implementation to return null, because it was required to fix a new test.

model-server/src/main/kotlin/org/modelix/model/server/SqlUtils.kt
```diff
@@ -40,12 +40,15 @@ internal class SqlUtils(private val connection: Connection) {
     ) {
         if (!isTableExisting(schemaName, tableName)) {
             val stmt = connection.createStatement()
+            println(creationSql)
```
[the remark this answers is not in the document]
slisson: Removed.

model-server/src/main/kotlin/org/modelix/model/server/store/IStoreClient.kt
languitar: Btw, looking at the DDD community, transaction management is usually understood as something that has to happen on the highest application level as it impacts the domain-specific consistency guarantees you want or need to give. That's also the reason why in Spring or Quarkus, usually either the REST controllers or the called application service method (assuming each request is served by a signle application service method) is marked as `@Transactional`.

slisson: We already documented to do it that way in https://issues.modelix.org/articles/MODELIX-A-91/0010-Dispatch-blocking-code-directly-from-request-handlers.

model-server/src/main/kotlin/org/modelix/model/server/store/IStoreClient.kt
slisson: Renamed it to `withRepositoryInCoroutine`.

model-server/src/main/kotlin/org/modelix/model/server/store/IStoreClient.kt
slisson: Done.

model-server/src/main/kotlin/org/modelix/model/server/store/ObjectInRepository.kt
slisson: Removed the comment. Now the reason is hidden in the implementation.

model-server/src/main/kotlin/org/modelix/model/server/handlers/IRepositoriesManager.kt
slisson: Without an alternative there is nothing to discuss.

model-server/src/main/kotlin/org/modelix/model/server/SqlUtils.kt
```diff
@@ -78,13 +85,41 @@ internal class SqlUtils(private val connection: Connection) {
...
+            ensureTableIsPresent(
+                schemaName,
+                userName,
+                "immutable_objects",
+                """
+                    CREATE TABLE $schemaName.immutable_objects (
+                        key VARCHAR(65535) NOT NULL,
+                        repository VARCHAR(65535) NULL,
+                        value VARCHAR(65535),
+                        reachable boolean,
```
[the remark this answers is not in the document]
slisson: Reverted to a single cache and the existing table. The new table doesn't exist anymore.

model-server/src/main/kotlin/org/modelix/model/server/store/ObjectInRepository.kt
```diff
@@ -0,0 +1,35 @@
...
+ * limitations under the License.
+ */
+
+package org.modelix.model.server.store
+
+import org.modelix.model.persistent.HashUtil
+
+data class ObjectInRepository(private val repositoryId: String, val key: String) {
+    fun isGlobal() = repositoryId == ""
+    fun isImmutable() = HashUtil.isSha256(key)
```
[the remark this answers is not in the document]
slisson: Removed that method.
I also optimized the performance of `isSha256`, which is now ~6 times faster.

model-server/src/main/kotlin/org/modelix/model/server/handlers/RepositoriesManager.kt
```diff
@@ -419,28 +493,28 @@ class VersionDeltaCache(val store: IDeserializingKeyValueStore) {
...
+                val key = versionHash to baseVersionHash
+                val existingDeferredDelta = cacheMap[key]?.get()
+                if (existingDeferredDelta != null) {
+                    LOG.debug("Version delta found in cache for {}.", key)
+                    existingDeferredDelta
+                } else {
+                    LOG.debug("Version delta not found in cache for {}.", key)
+                    val version = CLVersion(versionHash, store)
+                    val baseVersion = CLVersion(baseVersionHash, store)
+                    val newDeferredDelta = async {
```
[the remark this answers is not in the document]
slisson: Changed

model-server/src/main/kotlin/org/modelix/model/server/handlers/IRepositoriesManager.kt
languitar: An exception would be the alternative

model-client/src/commonMain/kotlin/org/modelix/model/client2/ModelClientV2.kt
```diff
@@ -124,14 +133,19 @@ class ModelClientV2(
...
     override suspend fun initRepository(repository: RepositoryId): IVersion {
+        return initRepository(repository, false)
+    }
+
+    suspend fun initRepository(repository: RepositoryId, legacyGlobalStorage: Boolean = false): IVersion {
         return httpClient.preparePost {
             url {
                 takeFrom(baseUrl)
                 appendPathSegmentsEncodingSlash("repositories", repository.id, "init")
+                if (legacyGlobalStorage) parameters["legacyGlobalStorage"] = legacyGlobalStorage.toString()
```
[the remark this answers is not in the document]
languitar: You could put that information in a `BREAKING CHANGE: ` footer on the commit. That way, it becomes part of the Changelog.

model-server/src/main/kotlin/org/modelix/model/server/SqlUtils.kt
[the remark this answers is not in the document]
languitar: But as a customer, how do I update my currently existing database now if the model-server doesn't do that on its own?

model-server-test/docker/initdb.sql
languitar: What about putting the insertion of test data into a separate file and reuse the "official" initidb.sql file from production code for setting up the database in the test. That way, we also verify that the official statements work as expected.

model-server-test/docker/Dockerfile
languitar: This comment is still valid for the compose file used in the tests.

model-server/src/main/kotlin/org/modelix/model/server/store/ObjectInRepository.kt
languitar: Not really? The reason is hidden but the caller still needs to know the special meaning of an empty string in contrast to maybe `null`, which would be the more obvious choice without knowing the implications for the database.

slisson: The empty string is not visible in the interface of this class. `getRepository()` returns `null` instead of an empty string and `ObjectInRepository.create` converts a null value to an empty string. I don't understand the problem.

model-server/src/main/kotlin/org/modelix/model/server/handlers/IRepositoriesManager.kt
slisson: That would require an additional check, if a repository exists before calling this method. That's not a better alternative. I don't understand why a null value is an issue at all.

model-server-test/docker/Dockerfile
slisson: That's why: https://github.com/modelix/modelix.kubernetes/blob/main/db/Dockerfile#L1

slisson: Updated it to 16.2

model-server-test/docker/initdb.sql
slisson: This file is used for backwards compatibility tests and should put the database into an "old" state. It's not supposed to be in sync with the latest version of the "official" database.

model-server/src/main/kotlin/org/modelix/model/server/handlers/IRepositoriesManager.kt
languitar: Not an issue. Just a question of what would be more intuitive for users of this interface.

model-server-test/docker/initdb.sql
languitar: Due to similar naming of the files I though it had the same content as the initdb.sql from the production code.

model-server-test/docker/initdb.sql
languitar: Then let's clarify this in the filename.

model-server/src/main/kotlin/org/modelix/model/server/store/ObjectInRepository.kt
languitar: aaaah, the whole removed comment targeted the first argument of ObjectInRepository. I understood this to talk about `key`. Good that the confusing comment is gone then.

model-server/src/main/kotlin/org/modelix/model/server/SqlUtils.kt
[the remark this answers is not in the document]
slisson: The model-server does update an existing database automatically.

model-server-test/docker/initdb.sql
slisson: Renamed it to `legacy-database.sql` and added a comment.

model-client/src/commonMain/kotlin/org/modelix/model/client2/ModelClientV2.kt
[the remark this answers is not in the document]
languitar: The trailer needs to be indented on continuation lines:

> The value may be split over multiple lines with each subsequent line starting with whitespace, like the "folding" in RFC 822.

https://git-scm.com/docs/git-interpret-trailers/2.22.0

model-server-test/build.gradle.kts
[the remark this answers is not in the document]
odzhychko: If `IgniteStoreClient` did nor rely on our properties names, we could use the properties or environment variables that the plugin sets directly in the test when creating an `IgniteStoreClient`.

(Here it does not make a difference. It might be more interesting in tests, where we do not want to have system properties at all.)
 
> * The name of environment variable is ${serviceName}_HOST and ${serviceName}_TCP_${exposedPort} (e.g. WEB_HOST and WEB_TCP_80).
> * The name of Java system property is ${serviceName}.host and ${serviceName}.tcp.${exposedPort} (e.g. web.host and web.tcp.80).
> * If the service is scaled then the serviceName has _1, _2... suffix (e.g. WEB_1_HOST and WEB_1_TCP_80, web_1.host and web_1.tcp.80)

https://github.com/avast/gradle-docker-compose-plugin?tab=readme-ov-file

model-server/src/main/kotlin/org/modelix/model/server/store/IStoreClient.kt
odzhychko: > I think splitting the StoreClient is a more complex solution. That moves responsibilities about transaction management out of the StoreClient to users of it.

Good point.

model-server/src/main/resources/org/modelix/model/server/store/ignite.xml
```diff
@@ -38,43 +38,19 @@
...
-                    </property>
-                </bean>
-            </list>
-        </property>
-    </bean>
-
     <bean id="ignite.cfg" class="org.apache.ignite.configuration.IgniteConfiguration">
-        <property name="peerClassLoadingEnabled" value="true"/>
-
+        <property name="consistentId" value="modelix-postgres"/>
```
odzhychko, line 42: > Sets consistent globally unique node ID which survives node restarts.

This might break starting multiple model-server in the same cluster.
That would break rolling deployments and disable https://issues.modelix.org/issue/MODELIX-344.

I think it will break because I guess that each node in the cluster needs a "globally unique node ID".
I am going to test it.

odzhychko (changes requested):

model-server/src/main/resources/org/modelix/model/server/store/ignite.xml
odzhychko:
```
SEVERE: Failed to start manager: GridManagerAdapter [enabled=true, name=o.a.i.i.managers.discovery.GridDiscoveryManager]
class org.apache.ignite.IgniteCheckedException: Failed to start SPI: TcpDiscoverySpi [addrRslvr=null, addressFilter=null, sockTimeout=5000, ackTimeout=5000, marsh=JdkMarshaller [clsFilter=org.apache.ignite.marshaller.MarshallerUtils$1@c82d925], reconCnt=10, reconDelay=2000, maxAckTimeout=600000, soLinger=0, forceSrvMode=false, clientReconnectDisabled=false, internalLsnr=null, skipAddrsRandomization=false]
	at org.apache.ignite.internal.managers.GridManagerAdapter.startSpi(GridManagerAdapter.java:280)
	at org.apache.ignite.internal.managers.discovery.GridDiscoveryManager.start(GridDiscoveryManager.java:1076)
	at org.apache.ignite.internal.IgniteKernal.startManager(IgniteKernal.java:1774)
	at org.apache.ignite.internal.IgniteKernal.start(IgniteKernal.java:1161)
	at org.apache.ignite.internal.IgnitionEx$IgniteNamedInstance.start0(IgnitionEx.java:1725)
	at org.apache.ignite.internal.IgnitionEx$IgniteNamedInstance.start(IgnitionEx.java:1647)
	at org.apache.ignite.internal.IgnitionEx.start0(IgnitionEx.java:1089)
	at org.apache.ignite.internal.IgnitionEx.startConfigurations(IgnitionEx.java:983)
	at org.apache.ignite.internal.IgnitionEx.start(IgnitionEx.java:889)
	at org.apache.ignite.internal.IgnitionEx.start(IgnitionEx.java:828)
	at org.apache.ignite.Ignition.start(Ignition.java:350)
	at org.modelix.model.server.store.IgniteStoreClient.<init>(IgniteStoreClient.kt:79)
	at org.modelix.model.server.store.IgniteStoreClient.<init>(IgniteStoreClient.kt:33)
	at org.modelix.model.server.Main.main(Main.kt:152)
Caused by: class org.apache.ignite.spi.IgniteSpiException: Failed to add node to topology because it has the same hash code for partitioned affinity as one of existing nodes [cacheName=ignite-sys-cache, existingNodeId=03232df7-8c18-4d79-9951-73e04c76e7b6]
	at org.apache.ignite.spi.discovery.tcp.TcpDiscoverySpi.checkFailedError(TcpDiscoverySpi.java:2110)
Caused by: class org.apache.ignite.spi.IgniteSpiException: Failed to add node to topology because it has the same hash code for partitioned affinity as one of existing nodes [cacheName=ignite-sys-cache, existingNodeId=03232df7-8c18-4d79-9951-73e04c76e7b6]

	at org.apache.ignite.spi.discovery.tcp.ServerImpl.joinTopology(ServerImpl.java:1204)
	at org.apache.ignite.spi.discovery.tcp.ServerImpl.spiStart(ServerImpl.java:472)
	at org.apache.ignite.spi.discovery.tcp.TcpDiscoverySpi.spiStart(TcpDiscoverySpi.java:2206)
	at org.apache.ignite.internal.managers.GridManagerAdapter.startSpi(GridManagerAdapter.java:277)
	... 13 more
```

It seems that setting a `consistentId` is indeed problematic.

model-server/src/main/kotlin/org/modelix/model/server/handlers/RepositoriesManager.kt
```diff
@@ -274,23 +316,31 @@ class RepositoriesManager(val client: LocalModelClient) : IRepositoriesManager {
...
     private fun getVersionHashBlocking(branch: BranchReference): String? {
         return store.runTransaction {
-            store[branchKey(branch)] ?: store[legacyBranchKey(branch)]?.also {
-                store.put(
-                    branchKey(branch),
-                    it,
-                    true,
-                )
-            }
+            store.getGenericStore()[branchKey(branch)]
```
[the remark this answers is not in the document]
slisson: The case `isIsolated() == null` is now handled more explicitly.

model-server/src/main/resources/org/modelix/model/server/store/ignite.xml
slisson: Removed it. Was just needed on my machine, because by default an ID is generated from the addresses of all network interface. I just have too many network interfaces, which resulted in a ID that is too long to be a valid directory name.

model-server/src/main/kotlin/org/modelix/model/server/handlers/RepositoriesManager.kt
```diff
@@ -68,21 +80,24 @@ class RepositoriesManager(val client: LocalModelClient) : IRepositoriesManager {
...
+                    migrateLegacyRepositoriesList(infoBranch)
+                }
             }
         }
 
         doMigrations()
     }
 
     private val store: IStoreClient get() = client.store
+    private val genericStore: IsolatingStore get() = (client.store as RepositoryScopedStoreClient).client
```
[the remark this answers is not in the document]
slisson: Removed.

model-server/src/main/kotlin/org/modelix/model/server/store/StoreClientAdapter.kt
```diff
@@ -0,0 +1,151 @@
...
+import org.modelix.model.persistent.HashUtil
+
+abstract class StoreClientAdapter(val client: IsolatingStore) : IStoreClient {
+
+    abstract fun getRepositoryId(): RepositoryId?
+
+    private fun String.withRepoScope(): ObjectInRepository {
+        return if (HashUtil.isSha256(this)) {
+            val id = getRepositoryId()
+            // checkNotNull(id) { "Repository required for key $this" }
```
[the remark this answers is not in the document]
slisson: Removed.

model-server/src/main/kotlin/org/modelix/model/server/SqlUtils.kt
```diff
@@ -78,13 +112,28 @@ internal class SqlUtils(private val connection: Connection) {
                 "model",
                 """
                     CREATE TABLE $schemaName.model (
-                        key character varying NOT NULL,
-                        value character varying,
-                        reachable boolean,
+                        key VARCHAR(65535) NOT NULL,
```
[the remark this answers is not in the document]
slisson: I tried writing tests using an in-memory database and that required this limit. Now I'm using docker-compose to run PostgreSQL. Removed it.

model-server/src/main/kotlin/org/modelix/model/server/SqlUtils.kt
```diff
@@ -42,10 +57,25 @@ internal class SqlUtils(private val connection: Connection) {
...
+        try {
+            val stmt = connection.createStatement()
+            stmt.execute("GRANT ALL ON TABLE $schemaName.$tableName TO $username;")
+        } catch (ex: SQLException) {
+            LOG.error("Failed to change permissions on $schemaName.$tableName", ex)
+        }
+    }
+
+    @Throws(SQLException::class)
+    fun ensureColumnIsPresent(
```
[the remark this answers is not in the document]
slisson: Removed the nullability.

model-server/src/main/kotlin/org/modelix/model/server/SqlUtils.kt
```diff
@@ -31,6 +31,21 @@ internal class SqlUtils(private val connection: Connection) {
         return false
     }
 
+    @Throws(SQLException::class)
+    fun isColumnExisting(schemaName: String?, tableName: String, columnName: String): Boolean {
```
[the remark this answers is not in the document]
slisson: Removed the nullability.

model-client/src/commonMain/kotlin/org/modelix/model/client2/ModelClientV2.kt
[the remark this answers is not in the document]
slisson: Fixed the commit message.

odzhychko (approved):

PR #692, opened 2024-04-15
fix(model-server): useRoleIds is a boolean parameter

benedekh:


model-server/src/main/kotlin/org/modelix/model/server/handlers/ModelReplicationServer.kt
```diff
@@ -170,7 +170,7 @@ class ModelReplicationServer(
             fun ApplicationCall.repositoryId() = RepositoryId(parameters["repository"]!!)
             fun PipelineContext<Unit, ApplicationCall>.repositoryId() = call.repositoryId()
 
-            val useRoleIds = call.request.queryParameters["useRoleIds"] != "false"
+            val useRoleIds = call.request.queryParameters["useRoleIds"]?.toBooleanStrict() != false
```
odzhychko, line 173: I'd rather not change the behavior because some code may rely on it.
See https://www.hyrumslaw.com/

I'm definitely for such strict checks and would strongly prefer them in new endpoints.

The change in the definition is good.

odzhychko, line 173: `toBooleanStrict` throws an `IllegalArgumentException`. This results in HTTP status 500 (Internal Server Error) and not (400 Bad Request).

odzhychko: `?useRoleIds=False` is interpreted as `true`... I'm for making it strict now.

benedekh: Fixed the broken behavior in [f500745](https://github.com/modelix/modelix.core/pull/692/commits/f50074594b160985ca753a2f43f125623bce9ccc)

odzhychko: `call.request.queryParameters["useRoleIds"]?.toBooleanStrictOrNull() != false` has exatly the same result as `val useRoleIds = call.request.queryParameters["useRoleIds"] != "false"` just with more overhead...

```java
@SinceKotlin("1.5")
public fun String.toBooleanStrictOrNull(): Boolean? = when (this) {
    "true" -> true
    "false" -> false
    else -> null
}
```

---

If you don't want to make the parsing stricter then let us only update the `model-server.yaml`.

odzhychko (approved):

PR #696, opened 2024-04-15
fix(model-server): useRoleIds is a boolean parameter

odzhychko:


model-server/src/main/kotlin/org/modelix/model/server/handlers/ModelReplicationServer.kt
```diff
@@ -169,8 +169,7 @@ class ModelReplicationServer(
         post<Paths.initializeRepository> {
             fun ApplicationCall.repositoryId() = RepositoryId(parameters["repository"]!!)
             fun PipelineContext<Unit, ApplicationCall>.repositoryId() = call.repositoryId()
-
-            val useRoleIds = call.request.queryParameters["useRoleIds"] != "false"
+            val useRoleIds = call.parseQueryParameterAsBoolean("useRoleIds") ?: return@post
```
benedekh, line 172: Alternative could be:

`val useRoleIds = call.request.queryParameters["useRoleIds"]?.toBooleanStrictOrNull() != false`

PR #606, opened 2024-03-14
perf: lazy loading for bulk sync

slisson:

PR #701, opened 2024-04-17
feat(model-client): lazy loading support for IModelClientV2

slisson:
Extracted the lazy loading feature without the bulk sync changes from https://github.com/modelix/modelix.core/pull/606

This still has the same weakness as the lazy loading in the V1 client (to many small requests), but it introduces the necessary REST endpoint and all future optimizations can happen entirely on the client side.

### DIFF
--- a/model-client/src/commonMain/kotlin/org/modelix/model/client2/IModelClientV2.kt
+++ b/model-client/src/commonMain/kotlin/org/modelix/model/client2/IModelClientV2.kt
@@ -85,4 +85,6 @@ interface IModelClientV2 {
     suspend fun <R> query(branch: BranchReference, body: (IMonoStep<INode>) -> IMonoStep<R>): R
 
     suspend fun <R> query(repositoryId: RepositoryId, versionHash: String, body: (IMonoStep<INode>) -> IMonoStep<R>): R
+
+    suspend fun getObjects(repository: RepositoryId, keys: Sequence<String>): Map<String, String>
 }
--- a/model-client/src/jvmMain/kotlin/org/modelix/model/client2/LazyLoading.kt
+++ b/model-client/src/jvmMain/kotlin/org/modelix/model/client2/LazyLoading.kt
@@ -1,0 +1,71 @@
+/*
+ * Copyright (c) 2024.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.modelix.model.client2
+
+import kotlinx.coroutines.runBlocking
+import org.modelix.model.IKeyListener
+import org.modelix.model.IKeyValueStore
+import org.modelix.model.IVersion
+import org.modelix.model.lazy.BranchReference
+import org.modelix.model.lazy.CLVersion
+import org.modelix.model.lazy.ObjectStoreCache
+import org.modelix.model.lazy.RepositoryId
+
+fun IModelClientV2.lazyLoadVersion(repositoryId: RepositoryId, versionHash: String, cacheSize: Int = 100_000): IVersion {
+    val store = ObjectStoreCache(ModelClientAsStore(this, repositoryId), cacheSize)
+    return CLVersion.loadFromHash(versionHash, store)
+}
+
+suspend fun IModelClientV2.lazyLoadVersion(branchRef: BranchReference, cacheSize: Int = 100_000): IVersion {
+    return lazyLoadVersion(branchRef.repositoryId, pullHash(branchRef), cacheSize)
+}
+
+class ModelClientAsStore(val client: IModelClientV2, val repositoryId: RepositoryId) : IKeyValueStore {
+    override fun get(key: String): String? {
+        return getAll(listOf(key))[key]
+    }
+
+    override fun put(key: String, value: String?) {
+        TODO("Not yet implemented")
+    }
+
+    override fun getAll(keys: Iterable<String>): Map<String, String?> {
+        return runBlocking {
+            client.getObjects(repositoryId, keys.asSequence())
+        }
+    }
+
+    override fun putAll(entries: Map<String, String?>) {
+        TODO("Not yet implemented")
+    }
+
+    override fun prefetch(key: String) {
+        TODO("Not yet implemented")
+    }
+
+    override fun listen(key: String, listener: IKeyListener) {
+        TODO("Not yet implemented")
+    }
+
+    override fun removeListener(key: String, listener: IKeyListener) {
+        TODO("Not yet implemented")
+    }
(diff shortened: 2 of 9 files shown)


model-server/src/main/kotlin/org/modelix/model/server/handlers/ModelReplicationServer.kt
```diff
@@ -213,6 +214,26 @@ class ModelReplicationServer(
             val newVersionHash = repositoriesManager.pollVersionHash(branchRef(), lastKnownVersionHash)
             call.respondDelta(newVersionHash, lastKnownVersionHash)
         }
+        post<Paths.postRepositoryObjectsGetAll> {
+            val repository = RepositoryId(call.parameters["repository"]!!)
+            val keys = call.receiveStream().bufferedReader().use { reader ->
+                reader.lineSequence().toHashSet()
+            }
+            val objects = withContext(Dispatchers.IO) { modelClient.store.getAll(keys) }
+            call.respondTextWriter(contentType = VersionDeltaStream.CONTENT_TYPE) {
```
odzhychko, line 223: This will suffer from the same bugs in Ktor we worked around in https://github.com/modelix/modelix.core/pull/650
Consider using a similar mechanism to `VersionDeltaStreamV2`.

model-server/src/main/kotlin/org/modelix/model/server/handlers/ModelReplicationServer.kt
```diff
@@ -213,6 +214,26 @@ class ModelReplicationServer(
             val newVersionHash = repositoriesManager.pollVersionHash(branchRef(), lastKnownVersionHash)
             call.respondDelta(newVersionHash, lastKnownVersionHash)
         }
+        post<Paths.postRepositoryObjectsGetAll> {
+            val repository = RepositoryId(call.parameters["repository"]!!)
```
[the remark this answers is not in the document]
slisson: It will be used soon in https://github.com/modelix/modelix.core/pull/691

model-server/src/main/kotlin/org/modelix/model/server/handlers/ModelReplicationServer.kt
slisson: I added an additional empty line to indicate the end of the stream. All these special characters as used in `VersionDeltaStreamV2` seems to be an unnecessary complex solution.

model-client/src/jvmMain/kotlin/org/modelix/model/client2/LazyLoading.kt
```diff
@@ -0,0 +1,71 @@
...
+import kotlinx.coroutines.runBlocking
+import org.modelix.model.IKeyListener
+import org.modelix.model.IKeyValueStore
+import org.modelix.model.IVersion
+import org.modelix.model.lazy.BranchReference
+import org.modelix.model.lazy.CLVersion
+import org.modelix.model.lazy.ObjectStoreCache
+import org.modelix.model.lazy.RepositoryId
+
+fun IModelClientV2.lazyLoadVersion(repositoryId: RepositoryId, versionHash: String, cacheSize: Int = 100_000): IVersion {
```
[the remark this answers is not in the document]
languitar: Can you add the missing kdoc for the new methods. For instance, I'd be lost on what `cacheSize` does without documentation here.

model-client/src/jvmMain/kotlin/org/modelix/model/client2/LazyLoading.kt
```diff
@@ -0,0 +1,71 @@
...
+        TODO("Not yet implemented")
+    }
+
+    override fun removeListener(key: String, listener: IKeyListener) {
+        TODO("Not yet implemented")
+    }
+
+    override fun getPendingSize(): Int {
+        TODO("Not yet implemented")
+    }
```
languitar, line 70: What about all these TODOs?

model-server/src/main/kotlin/org/modelix/model/server/handlers/ModelReplicationServer.kt
```diff
@@ -246,6 +247,23 @@ class ModelReplicationServer(
             val newVersionHash = repositoriesManager.pollVersionHash(branchRef(), lastKnownVersionHash)
             call.respondDelta(newVersionHash, lastKnownVersionHash)
         }
+        post<Paths.postRepositoryObjectsGetAll> {
+            val repository = RepositoryId(call.parameters["repository"]!!) // will be used in the future
```
languitar, line 251: Can we add that to the PR that will make use of it instead?

model-server/src/main/kotlin/org/modelix/model/server/handlers/ModelReplicationServer.kt
languitar: Whitespaces are pretty easy to mess up and a trailing newline, while formally sufficient to resolve the problem, is not a very explicit marker and can easily be overlooked. I'd be more in favor of an explicit end marker as it is then less likely that the next one touching the code forgets about the reason for the trailing newline.

model-server/src/test/kotlin/org/modelix/model/server/LazyLoadingTest.kt
```diff
@@ -0,0 +1,109 @@
...
+                    parentNode.addNewChild(NullChildLink, 0)
+                    return
+                }
+                val subtreeSize1 = numberOfNodes / 2
+                val subtreeSize2 = numberOfNodes - subtreeSize1
+                createNodes(parentNode.addNewChild(NullChildLink, 0), subtreeSize1 - 1)
+                createNodes(parentNode.addNewChild(NullChildLink, 1), subtreeSize2 - 1)
+            }
+
+            createNodes(it, 5_000)
```
languitar, line 85: Is it necessary to create that many nodes for being able to verify the behavior? This adds test execution time, of which we already have enough.

slisson:
> Whitespaces are pretty easy to mess up and a trailing newline, while formally sufficient to resolve the problem, is not a very explicit marker and can easily be overlooked. I'd be more in favor of an explicit end marker as it is then less likely that the next one touching the code forgets about the reason for the trailing newline.

I like that it's doesn't pollute the data with special characters and the response, when inspected, remains as easy to understand as before. The unit tests will fail if someone changes the code to not include the additional new line. Also, there are comments and error messages that document the reason.

model-server/src/main/kotlin/org/modelix/model/server/handlers/ModelReplicationServer.kt
slisson: I like that it's doesn't pollute the data with special characters and the response, when inspected, remains as easy to understand as before. The unit tests will fail if someone changes the code to not include the additional new line. Also, there are comments and error messages that document the reason.

model-client/src/commonMain/kotlin/org/modelix/model/client2/IModelClientV2.kt
```diff
@@ -85,4 +85,6 @@ interface IModelClientV2 {
     suspend fun <R> query(branch: BranchReference, body: (IMonoStep<INode>) -> IMonoStep<R>): R
 
     suspend fun <R> query(repositoryId: RepositoryId, versionHash: String, body: (IMonoStep<INode>) -> IMonoStep<R>): R
+
+    suspend fun getObjects(repository: RepositoryId, keys: Sequence<String>): Map<String, String>
```
languitar, line 89: From #723:

Any reason for using a sequence here as the input type when the output type is not lazy?

PR #714, opened 2024-04-25
feat(model-server): implement deleting branches via REST API and ModelClient

languitar:
Adds a new DELETE endpoint to the REST API for deleting a branch from a repository. Moreover, the ModelClient is extended with the matching `deleteBranch` method.

Garbage collection of data potentially not referenced anymore is not handled here and will be part of a later feature implementation.

### DIFF
--- a/model-server/src/main/kotlin/org/modelix/model/server/handlers/ModelReplicationServer.kt
+++ b/model-server/src/main/kotlin/org/modelix/model/server/handlers/ModelReplicationServer.kt
@@ -23,6 +23,7 @@ import io.ktor.server.plugins.origin
 import io.ktor.server.request.acceptItems
 import io.ktor.server.request.receive
 import io.ktor.server.request.receiveStream
+import io.ktor.server.resources.delete
 import io.ktor.server.resources.get
 import io.ktor.server.resources.post
 import io.ktor.server.resources.put
@@ -146,6 +147,37 @@ class ModelReplicationServer(
                 return@get
             }
             call.respondDelta(versionHash, baseVersionHash)
+        }
+
+        delete<Paths.deleteRepositoryBranch> {
+            val repositoryName = call.parameters["repository"]
+            if (repositoryName == null) {
+                call.respondText("Request lacks repository name", status = HttpStatusCode.BadRequest)
+                return@delete
+            }
+            val repositoryId = try {
+                RepositoryId(repositoryName)
+            } catch (e: IllegalArgumentException) {
+                call.respondText("Invalid repository name '$repositoryName'", status = HttpStatusCode.BadRequest)
+                return@delete
+            }
+            val branch = call.parameters["branch"]
+            if (branch == null) {
+                call.respondText("Request lacks branch name", status = HttpStatusCode.BadRequest)
+                return@delete
+            }
+
+            if (!repositoriesManager.getBranchNames(repositoryId).contains(branch)) {
+                call.respondText(
+                    "Repository does not exist or branch '$branch' does not exist in repository '$repositoryId'",
+                    status = HttpStatusCode.NotFound,
+                )
+                return@delete
+            }
+
+            repositoriesManager.removeBranches(repositoryId, setOf(branch))
+
+            call.respond(HttpStatusCode.NoContent)
         }
 
         get<Paths.getRepositoryBranchHash> {
--- a/model-server/src/test/kotlin/org/modelix/model/server/handlers/ModelReplicationServerTest.kt
+++ b/model-server/src/test/kotlin/org/modelix/model/server/handlers/ModelReplicationServerTest.kt
@@ -19,7 +19,9 @@ import io.ktor.client.HttpClient
 import io.ktor.client.engine.cio.CIO
 import io.ktor.client.plugins.HttpTimeout
 import io.ktor.client.plugins.defaultRequest
+import io.ktor.client.request.delete
 import io.ktor.client.request.get
+import io.ktor.client.statement.bodyAsText
 import io.ktor.http.HttpHeaders
 import io.ktor.http.HttpStatusCode
 import io.ktor.http.appendPathSegments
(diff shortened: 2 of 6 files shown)


model-server/src/main/kotlin/org/modelix/model/server/handlers/ModelReplicationServer.kt
```diff
@@ -147,6 +148,37 @@ class ModelReplicationServer(
             call.respondDelta(versionHash, baseVersionHash)
         }
 
+        delete<Paths.deleteRepositoryBranch> {
+            val repositoryName = call.parameters["repository"]
+            if (repositoryName == null) {
+                call.respondText("Request lacks repository name", status = HttpStatusCode.BadRequest)
+                return@delete
```
odzhychko, line 155: thought: It is unfortunate that we can not just throw the `io.ktor.server.plugins.BadRequestException`.

We would need to refactor the usage of the StatusPage plugin
See https://github.com/modelix/modelix.core/blob/main/authorization/src/main/kotlin/org/modelix/authorization/KtorAuthUtils.kt#L88

model-server/src/main/kotlin/org/modelix/model/server/handlers/ModelReplicationServer.kt
```diff
@@ -147,6 +148,37 @@ class ModelReplicationServer(
...
+            }
+            val repositoryId = try {
+                RepositoryId(repositoryName)
```
odzhychko, line 158: Maybe add a method to `RepositoryId` to parse and optionally return an ID.
Not relying on the exception would also get rid of the Detekt warning.

odzhychko: Also this case is not tested.

model-server/src/test/kotlin/org/modelix/model/server/handlers/ModelReplicationServerTest.kt
```diff
@@ -113,6 +117,81 @@ class ModelReplicationServerTest {
...
+    @Test
+    fun `responds with 404 when deleting a branch from a non-existent repository`() {
+        val url = "http://localhost/v2"
+        val storeClient = InMemoryStoreClient()
+        val modelClient = LocalModelClient(storeClient)
+        val repositoriesManager = RepositoriesManager(modelClient)
+        val modelReplicationServer = ModelReplicationServer(repositoriesManager, modelClient, InMemoryModels())
+
+        runWithTestModelServer(modelReplicationServer) {
+            val response = client.delete {
```
odzhychko, line 129: Should we directly add a `deleteBranch` method to `ModelClientV2`?

model-server/src/test/kotlin/org/modelix/model/server/handlers/ModelReplicationServerTest.kt
```diff
@@ -113,6 +117,81 @@ class ModelReplicationServerTest {
...
 
+    @Test
+    fun `responds with 404 when deleting a branch from a non-existent repository`() {
+        val url = "http://localhost/v2"
+        val storeClient = InMemoryStoreClient()
+        val modelClient = LocalModelClient(storeClient)
+        val repositoriesManager = RepositoriesManager(modelClient)
+        val modelReplicationServer = ModelReplicationServer(repositoriesManager, modelClient, InMemoryModels())
+
+        runWithTestModelServer(modelReplicationServer) {
```
odzhychko, line 128: In this case you do not need to cunstruct a `modelReplicationServer`.
`runWithTestModelServer` uses `getDefaultModelReplicationServer()` to construct it.

model-server/src/main/kotlin/org/modelix/model/server/handlers/ModelReplicationServer.kt
```diff
@@ -147,6 +148,37 @@ class ModelReplicationServer(
...
+
+            if (!repositoriesManager.getBranchNames(repositoryId).contains(branch)) {
+                call.respondText(
+                    "Repository does not exist or branch '$branch' does not exist in repository '$repositoryId'",
+                    status = HttpStatusCode.NotFound,
+                )
+                return@delete
+            }
+
+            repositoriesManager.removeBranches(repositoryId, setOf(branch))
```
odzhychko, line 177: I suspect this only deletes a branch in the V2 storage and not the V1 storage. I will take look at this and come back to you.

odzhychko (changes requested):

model-server/src/main/kotlin/org/modelix/model/server/handlers/ModelReplicationServer.kt
odzhychko: Never mind. This is already correctly handled.

model-server/src/main/kotlin/org/modelix/model/server/handlers/ModelReplicationServer.kt
languitar: Yes, the lack of a default HTTP exception hierarchy in ktor is a pain. However, we should probably split that refactoring (for the whole implementation) into a separate PR?

odzhychko: Definitly. We will propably also implement https://issues.modelix.org/articles/MODELIX-A-83/0009-Use-application-problemjson-for-error-responses-in-REST-APIs when cleaning it up.

model-server/src/test/kotlin/org/modelix/model/server/handlers/ModelReplicationServerTest.kt
languitar: I had to do that to get access to the repositories manager, but I refactor the tests to provide that object to the block passed to `runWithTestModelServer`.

model-server/src/main/kotlin/org/modelix/model/server/handlers/ModelReplicationServer.kt
languitar: I'm fine with the detect warning. That's a problem for all top-level exceptions that are handled differently from rethrowing them somehow. And the ktor approach requires this.

We have to generally refactor this for the whole file, similar to the exception handling. I'd like to postpone that to a separate PR.

model-server/src/test/kotlin/org/modelix/model/server/handlers/ModelReplicationServerTest.kt
languitar: it's added as a second commit now

model-server/src/test/kotlin/org/modelix/model/server/handlers/ModelReplicationServerTest.kt
languitar: better this way?

odzhychko: Definitely. This is also a nice pattern for other tests.

odzhychko (approved):

odzhychko (approved):

PR #715, opened 2024-04-25
fix(model-server): make initial branch created through V2 API visible in V1 API

odzhychko:
Writing to a branch through V2 API already makes a branch visible through the V1 API. Making the branch visible after initial creation makes the backwards compatibility more consistent.

I found this issue while reviewing and testing #714

### DIFF
--- a/model-server/src/test/kotlin/org/modelix/model/server/handlers/ModelReplicationServerBackwardsCompatibilityTest.kt
+++ b/model-server/src/test/kotlin/org/modelix/model/server/handlers/ModelReplicationServerBackwardsCompatibilityTest.kt
@@ -1,0 +1,77 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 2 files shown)


languitar:
This should probably be a draft until a rebase after merging #714?

model-server/src/test/kotlin/org/modelix/model/server/handlers/ModelReplicationServerBackwardsCompatibilityTest.kt
```diff
@@ -0,0 +1,77 @@
...
+            keyValueLikeModelServer.init(this)
+        }
+
+        coroutineScope {
+            block(this)
+        }
+    }
+
+    @Test
+    fun `master branch of repository initialized by V2 API is visible through V2 API`() {
```
languitar, line 61: One of them should be V1?

languitar (changes requested):

model-server/src/test/kotlin/org/modelix/model/server/handlers/ModelReplicationServerBackwardsCompatibilityTest.kt
odzhychko: Yes. Done.

languitar (approved):

PR #725, opened 2024-05-02
Changes in modelix to make MODELIX-630 work

benedekh:
Changes needed in modelix to make [MODELIX-630](https://github.com/modelix/modelix.mps-plugins/pull/97) work.

### DIFF
--- a/model-client/src/commonMain/kotlin/org/modelix/model/client2/ReplicatedModel.kt
+++ b/model-client/src/commonMain/kotlin/org/modelix/model/client2/ReplicatedModel.kt
@@ -23,29 +23,44 @@ import org.modelix.model.operations.OTBranch
 
 /**
  * Dispose should be called on this, as otherwise a regular polling will go on.
+ *
+ * @property client the model client to connect to the model server
+ * @property branchRef the model server branch to fetch the data from
+ * @property providedScope the CoroutineScope to use for the suspendable tasks
+ * @property initialRemoteVersion the last version on the server from which we want to start the synchronization
  */
 class ReplicatedModel(
     val client: IModelClientV2,
     val branchRef: BranchReference,
     private val providedScope: CoroutineScope? = null,
+    initialRemoteVersion: CLVersion? = null,
 ) {
     private val scope = providedScope ?: CoroutineScope(Dispatchers.Default)
     private var state = State.New
-    private lateinit var localModel: LocalModel
-    private val remoteVersion = RemoteVersion(client, branchRef)
+    private var localModel: LocalModel? = null
+    private val remoteVersion = RemoteVersion(client, branchRef, initialRemoteVersion)
     private var pollingJob: Job? = null
 
+    init {
+        if (initialRemoteVersion != null) {
+            localModel = LocalModel(initialRemoteVersion, client.getIdGenerator()) { client.getUserId() }
+        }
+    }
+
+    private fun getLocalModel(): LocalModel = checkNotNull(localModel) { "Model is not initialized yet" }
+
     fun getBranch(): IBranch {
-        if (state != State.Started) throw IllegalStateException("state is $state")
-        return localModel.otBranch
+        return getLocalModel().otBranch
     }
 
     suspend fun start(): IBranch {
         if (state != State.New) throw IllegalStateException("already started")
         state = State.Starting
 
-        val initialVersion = remoteVersion.pull()
-        localModel = LocalModel(initialVersion, client.getIdGenerator(), { client.getUserId() })
+        if (localModel == null) {
+            val initialVersion = remoteVersion.pull()
+            localModel = LocalModel(initialVersion, client.getIdGenerator()) { client.getUserId() }
+        }
 
         // receive changes from the server
         pollingJob = scope.launch {
@@ -66,7 +81,7 @@ class ReplicatedModel(
             }
         }
 
-        localModel.rawBranch.addListener(object : IBranchListener {
+        getLocalModel().rawBranch.addListener(object : IBranchListener {
             override fun treeChanged(oldTree: ITree?, newTree: ITree) {
                 if (isDisposed()) return
                 scope.launch {
@@ -80,7 +95,7 @@ class ReplicatedModel(
     }
 
     suspend fun resetToServerVersion() {
-        localModel.resetToVersion(client.pull(branchRef, lastKnownVersion = null).upcast())
+        getLocalModel().resetToVersion(client.pull(branchRef, lastKnownVersion = null).upcast())
     }
 
     fun isDisposed(): Boolean = state == State.Disposed
@@ -102,11 +117,11 @@ class ReplicatedModel(
         if (isDisposed()) return
 
         val mergedVersion = try {
-            localModel.mergeRemoteVersion(newRemoteVersion)
+            getLocalModel().mergeRemoteVersion(newRemoteVersion)
         } catch (ex: Exception) {
-            val currentLocalVersion = localModel.getCurrentVersion()
+            val currentLocalVersion = getLocalModel().getCurrentVersion()
             LOG.warn(ex) { "Failed to merge remote version $newRemoteVersion into local version $currentLocalVersion. Resetting to remote version." }
-            localModel.resetToVersion(newRemoteVersion)
+            getLocalModel().resetToVersion(newRemoteVersion)
             newRemoteVersion
         }
 
@@ -121,7 +136,7 @@ class ReplicatedModel(
     private suspend fun pushLocalChanges() {
         if (isDisposed()) return
 
-        val version = localModel.createNewLocalVersion() ?: localModel.getCurrentVersion()
+        val version = getLocalModel().createNewLocalVersion() ?: getLocalModel().getCurrentVersion()
         val received = remoteVersion.push(version)
         if (received.getContentHash() != version.getContentHash()) {
             remoteVersionReceived(received)
@@ -129,7 +144,7 @@ class ReplicatedModel(
     }
 
     suspend fun getCurrentVersion(): CLVersion {
-        return localModel.getCurrentVersion()
+        return getLocalModel().getCurrentVersion()
     }
 
     private enum class State {
@@ -250,8 +265,11 @@ private class LocalModel(initialVersion: CLVersion, val idGenerator: IIdGenerato
     }
 }
 
-private class RemoteVersion(val client: IModelClientV2, val branchRef: BranchReference) {
-    private var lastKnownRemoteVersion: CLVersion? = null
+private class RemoteVersion(
+    val client: IModelClientV2,
+    val branchRef: BranchReference,
+    private var lastKnownRemoteVersion: CLVersion? = null,
+) {
     private val unconfirmedVersions: MutableSet<String> = LinkedHashSet()
 
     fun getNumberOfUnconfirmed() = runSynchronized(unconfirmedVersions) { unconfirmedVersions.size }
--- a/model-server/src/test/kotlin/org/modelix/model/server/ReplicatedModelTest.kt
+++ b/model-server/src/test/kotlin/org/modelix/model/server/ReplicatedModelTest.kt
@@ -1,0 +1,183 @@
+/*
+ * Copyright (c) 2024.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.modelix.model.server
+
+import io.ktor.server.testing.ApplicationTestBuilder
+import io.ktor.server.testing.testApplication
+import io.ktor.util.reflect.instanceOf
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.launch
+import org.junit.Assert.assertFalse
+import org.modelix.authorization.installAuthentication
+import org.modelix.model.api.ChildLinkFromName
+import org.modelix.model.api.ConceptReference
+import org.modelix.model.api.IBranch
+import org.modelix.model.api.ITree
+import org.modelix.model.api.PBranch
+import org.modelix.model.api.getRootNode
+import org.modelix.model.client2.ModelClientV2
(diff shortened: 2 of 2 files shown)


model-client/src/commonMain/kotlin/org/modelix/model/client2/ReplicatedModel.kt
```diff
@@ -251,7 +258,7 @@ private class LocalModel(initialVersion: CLVersion, val idGenerator: IIdGenerato
 }
 
 private class RemoteVersion(val client: IModelClientV2, val branchRef: BranchReference) {
-    private var lastKnownRemoteVersion: CLVersion? = null
+    var lastKnownRemoteVersion: CLVersion? = null
```
languitar, line 261: Would probably be better to pass this in via the constructor?

benedekh: fixed in [6e976c7](https://github.com/modelix/modelix.core/pull/725/commits/6e976c797ed41c074ade6aaf90ef2b07a7a44c68)

model-client/src/commonMain/kotlin/org/modelix/model/client2/ReplicatedModel.kt
```diff
@@ -32,20 +32,23 @@ class ReplicatedModel(
...
+    private lateinit var remoteVersion: RemoteVersion
     private var pollingJob: Job? = null
 
     fun getBranch(): IBranch {
         if (state != State.Started) throw IllegalStateException("state is $state")
         return localModel.otBranch
     }
 
-    suspend fun start(): IBranch {
+    suspend fun start(providedInitialVersion: CLVersion? = null, onBeforeSync: (IBranch) -> Unit = {}): IBranch {
```
languitar, line 43: * Wouldn't it be more consistent to also provide `providedInitialVersion` via the constructor, similar to the branch reference?
* Why don't we use the same name as in the `RemoteVersion` object: `lastKnownRemoteVersion`?

languitar (changes requested):
Sorry, looking on this again after a few days I see some things we should still think about:

* What about tests for the new feature?
* (One of the) commits should probably be of type `feat` as the user of the model client is now able to do something new.
* What exactly happens when `providedInitialVersion` is given? Some kdoc for the new arguments and the consequences they cause if provided would be good.

model-client/src/commonMain/kotlin/org/modelix/model/client2/ReplicatedModel.kt
slisson: Moving `providedInitialVersion` to the constructor also allows initializing `localModel` earlier and then the `onBeforeSync` callback isn't necessary.

benedekh: We need the `onBeforeSync` call, because we would like to subscribe to changes to the IBranch as soon as it is fetched from the server (via the LocalModel). Is there any other way to accomplish this?

benedekh:
> (One of the) commits should probably be of type feat as the user of the model client is now able to do something new.
> What exactly happens when providedInitialVersion is given? Some kdoc for the new arguments and the consequences they cause if provided would be good.
> Wouldn't it be more consistent to also provide providedInitialVersion via the constructor, similar to the branch reference?
> Why don't we use the same name as in the RemoteVersion object: lastKnownRemoteVersion?

fixed in [86dd24c](https://github.com/modelix/modelix.core/pull/725/commits/86dd24c1183e2cf0213864f2d56d0d134077568f)

Missing open point that I will take care of later today (hopefully):

> What about tests for the new feature?

model-client/src/commonMain/kotlin/org/modelix/model/client2/ReplicatedModel.kt
benedekh: fixed in [86dd24c](https://github.com/modelix/modelix.core/pull/725/commits/86dd24c1183e2cf0213864f2d56d0d134077568f), except that the `onBeforeSync` has to stay, because we would like to subscribe to changes on the IBranch, before we fetch the latest version from the server.

benedekh:
> What about tests for the new feature?

fixed in [0829ca0](https://github.com/modelix/modelix.core/pull/725/commits/0829ca0afec72d8bccc079e931dcaebcfa44338f)

@languitar I've fixed all your requests. Please re-review this PR, when you have time.

model-client/src/commonMain/kotlin/org/modelix/model/client2/ReplicatedModel.kt
```diff
@@ -23,29 +23,47 @@ import org.modelix.model.operations.OTBranch
...
 /**
  * Dispose should be called on this, as otherwise a regular polling will go on.
+ *
+ * @property providedScope the CoroutineScope to use for the suspendable tasks
+ * @property lastKnownRemoteVersion the last version on the server from which we want to start the synchronization
+ * @property scope the CoroutineScope to use for the suspendable tasks. Uses providedScope or defaults to the default dispatchers
+ * @property state the current state of the ReplicatedModel
+ * @property localModel thread-safe maintenance of the local version
+ * @property remoteVersion to send and receive versions from the server
+ * @property pollingJob the coroutine job that periodically checks if there is a new remote version and fetches its content
```
languitar, line 33: > @ property name
> 
> Documents the property of a class which has the specified name. This tag can be used for documenting properties declared in the primary constructor, where putting a doc comment directly before the property definition would be awkward.

https://kotlinlang.org/docs/kotlin-doc.html#property-name

Looks like properties declared in the class body should be documented directly.

model-server/src/test/kotlin/org/modelix/model/server/ReplicatedModelTest.kt
```diff
@@ -0,0 +1,154 @@
...
+        // we do not provide an initial version, so we expect to fetch the latest one (twoHelloChildrenVersion)
+        val replicatedModel = ReplicatedModel(newClient, defaultBranchReference)
+        val branch = replicatedModel.start { branch ->
+            // Step 3: check, here we must have the twoHelloChildrenVersion (2 "hello" children)
+            val twoHelloChildren = getHelloChildrenOfRootNode(branch)
+            assertEquals(2, twoHelloChildren.size)
+        }
+
+        // Step 4: wait a bit so replicated model can fetch the new versions from the server (in case there would be any -- there is none by the way)
+        delay(500)
```
languitar, line 72: This looks like it has the potential to result in a flaky unit test if we need waiting for synchronizing. Moreover, from the `runTest` docs:

> On JVM and Native, this function behaves similarly to runBlocking, with the difference that the code that it runs will skip delays. This allows to use [delay](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/index.html) in tests without causing them to take more time than necessary. On JS, this function creates a Promise that executes the test body with the delay-skipping behavior.

So do we really need this here?

model-server/src/test/kotlin/org/modelix/model/server/ReplicatedModelTest.kt
```diff
@@ -0,0 +1,154 @@
...
+        // Step 1: prepare repository with two versions beside the initial version
+        // Step 1.1: create an empty repository
+        val initialVersion = client.initRepository(repositoryId)
+
+        // Step 1.2: add a new child node to get a new version
+        val defaultBranchReference = repositoryId.getBranchReference()
+        val oneHelloChildVersion = addHelloChild(initialVersion, client, defaultBranchReference)
+
+        // Step 1.3: add a new child node to get a newer version
+        val twoHelloChildrenVersion = addHelloChild(oneHelloChildVersion, client, defaultBranchReference)
```
languitar, line 59: Can't we eliminate this second child? One child (+ 1 version) should be enough to test that the latest version is fetched?

model-server/src/test/kotlin/org/modelix/model/server/ReplicatedModelTest.kt
```diff
@@ -0,0 +1,154 @@
...
+            lastKnownRemoteVersion = oneHelloChildVersionClone as CLVersion,
+        )
+        val branch = replicatedModel.start { branch ->
+            // Step 3: check, here we must have the oneHelloChildVersion, which has 1 "hello" child
+            val oneHelloChild = getHelloChildrenOfRootNode(branch)
+            assertEquals(1, oneHelloChild.size)
+        }
+
+        // Step 4: wait a bit so replicated model can fetch the new versions from the server
+        delay(500)
```
languitar, line 114: Can we get this deterministic without waiting?

benedekh: Good question. Because the point is, the ReplicatedModel must poll the latest version which is running in a Job started on the CoroutineScope.

Maybe with a blocking wait and a timeout. That way we do not wait 500 ms, but shorter if the response arrive quicker. Otherwise we wait untilt timeout.

model-server/src/test/kotlin/org/modelix/model/server/ReplicatedModelTest.kt
```diff
@@ -0,0 +1,154 @@
...
+        // Step 1: prepare repository with two versions beside the initial version
+        // Step 1.1: create an empty repository
+        val initialVersion = client.initRepository(repositoryId) as CLVersion
+
+        // Step 1.2: add a new child node to get a new version
+        val defaultBranchReference = repositoryId.getBranchReference()
+        val oneHelloChildVersion = addHelloChild(initialVersion, client, defaultBranchReference)
+
+        // Step 1.3: add a new child node to get a newer version
+        val twoHelloChildrenVersion = addHelloChild(oneHelloChildVersion, client, defaultBranchReference)
```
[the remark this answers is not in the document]
benedekh: outdated remark, line does not exist anymore.

model-server/src/test/kotlin/org/modelix/model/server/ReplicatedModelTest.kt
```diff
@@ -0,0 +1,154 @@
...
+
+        // Step 1: prepare repository with two versions beside the initial version
+        // Step 1.1: create an empty repository
+        val initialVersion = client.initRepository(repositoryId) as CLVersion
+
+        // Step 1.2: add a new child node to get a new version
+        val oneHelloChildVersion = addHelloChild(initialVersion, client, defaultBranchReference)
+
+        // Step 1.3: add a new child node to get a newer version
+        val twoHelloChildrenVersion = addHelloChild(oneHelloChildVersion, client, defaultBranchReference)
```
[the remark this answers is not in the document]
benedekh: outdated remark, line does not exist anymore.

model-client/src/commonMain/kotlin/org/modelix/model/client2/ReplicatedModel.kt
benedekh: in [481f83d](https://github.com/modelix/modelix.core/pull/725/commits/481f83da1fef69c3b946773d198703fcae70ce8e), I documented the public API in the end.

model-server/src/test/kotlin/org/modelix/model/server/ReplicatedModelTest.kt
benedekh: removed in [481f83d](https://github.com/modelix/modelix.core/pull/725/commits/481f83da1fef69c3b946773d198703fcae70ce8e)

model-server/src/test/kotlin/org/modelix/model/server/ReplicatedModelTest.kt
benedekh: removed in [481f83d](https://github.com/modelix/modelix.core/pull/725/commits/481f83da1fef69c3b946773d198703fcae70ce8e)

model-server/src/test/kotlin/org/modelix/model/server/ReplicatedModelTest.kt
```diff
@@ -0,0 +1,165 @@
...
+        // Step 1.2: add a new child node to get a new version
+        addHelloChild(initialVersion, client, defaultBranchReference)
+
+        // Step 2: in a new client, fetch the oneHelloChildVersion
+        val newClient = createModelClient()
+
+        // Step 2.1: to avoid version was not created by this client exception
+        val initialVersionClone = newClient.loadVersion(repositoryId, initialVersion.getContentHash(), null)
+
+        val scope = CoroutineScope(Dispatchers.Default)
```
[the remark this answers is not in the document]
languitar: Why is this used now?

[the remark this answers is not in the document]
benedekh: Because there is no direct way to check if ReplicatedModel has fetched the latest version. Indirectly, I have to wait until the "hello" child appears. 

In order to avoid a fixed wait of 500 ms in the coroutine that is running at line 97, I create a CountdownLatch and start a new coroutine that checks if the "hello" child has arrived. If it has, then it opens the latch. The coroutine running at line 97 will be stopped by the Latch and wait until the Latch is opened or 500 ms has passed whichever occurs first.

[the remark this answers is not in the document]
languitar: ah, and the single-threaded test dispatcher would be blocked already by waiting on the latch?

model-client/src/commonMain/kotlin/org/modelix/model/client2/ReplicatedModel.kt
languitar: @slisson fine with that now?

slisson: I you provide a `lastKnownRemoteVersion` to the constructor then you can initialize `localModel` already in the constructor instead of in `start`. You can then call `getBranch()` and register your listener before calling `start()`, which means the `onBeforeSync` callback isn't necessary anymore.

benedekh: I'll make this change then.

benedekh: fixed in [9b0ea3f](https://github.com/modelix/modelix.core/pull/725/commits/9b0ea3fa558315b5a31846a9057adb2e484365ba)

benedekh: @slisson is the fix in [9b0ea3f](https://github.com/modelix/modelix.core/pull/725/commits/9b0ea3fa558315b5a31846a9057adb2e484365ba) fine for you?

slisson: I changed the commit a bit. Now it can be merged.

slisson (approved):

languitar (approved):

PR #728, opened 2024-05-03
fix(mps-model-adapters): allow removing references on node

odzhychko:


mps-model-adapters/src/main/kotlin/org/modelix/model/mpsadapters/MPSNode.kt
```diff
@@ -220,8 +220,8 @@ data class MPSNode(val node: SNode) : IDefaultNodeAdapter, IReplaceableNode {
                 ?: SReferenceLinkAdapterById(SReferenceLinkId.deserialize(link.getUID()), "")
         }
 
-        val targetNode = target?.let { getArea().resolveNode(it.reference) } as MPSNode
-        node.setReferenceTarget(refLink, targetNode.node)
+        val targetNode = target?.let { getArea().resolveNode(it.reference) as MPSNode? }
```
slisson, line 223: This line didn't make sense before and still doesn't. Why do we try to resolve a node that is already an `INode`? Either the caller passes a valid `MPSNode` or this method has to fail. This is the wrong place for re-resolving nodes.

odzhychko: Done

mps-model-adapters-plugin/src/test/kotlin/org/modelix/model/mpsadapters/ChangeReferenceTest.kt
```diff
@@ -0,0 +1,63 @@
...
+        runCommandOnEDT {
+            val module = repositoryNode.getChildren(BuiltinLanguages.MPSRepositoryConcepts.Repository.modules)
+                .single { it.getPropertyValue(BuiltinLanguages.jetbrains_mps_lang_core.INamedConcept.name) == "Solution1" }
+            val model = module.getChildren(BuiltinLanguages.MPSRepositoryConcepts.Module.models)
+                .single { it.getPropertyValue(BuiltinLanguages.jetbrains_mps_lang_core.INamedConcept.name) == "Solution1.model1" }
+            val rootNode = model.getChildren(BuiltinLanguages.MPSRepositoryConcepts.Model.rootNodes).single()
+            rootNode.setReferenceTarget(referenceLink, rootNode)
+            assertEquals(rootNode, rootNode.getReferenceTarget(referenceLink))
+
+            rootNode.setReferenceTarget(referenceLink, null as INodeReference?)
```
slisson, line 39:
```suggestion
            rootNode.setReferenceTarget(referenceLink, null as INode?)
```

This is calling the wrong method. The correct one is called indirectly, but that might change.

odzhychko: Done.
Testing directly is better.

mhuster23 (approved):

PR #729, opened 2024-05-03
fix(mps-model-adapters): return null when resolving project module in an unresolvable project

odzhychko:
Previously, the method failed while trying to cast null to a non-null type.


mps-model-adapters/src/main/kotlin/org/modelix/model/mpsadapters/MPSArea.kt
```diff
@@ -310,9 +310,12 @@ data class MPSArea(val repository: SRepository) : IArea, IAreaReference {
             NodeReference(projectRef)
         }
 
+        val resolvedNodeForProject = resolveNode(projectRef) ?: return null
+        val resolvedProject = (resolvedNodeForProject as MPSProjectAsNode).project
```
languitar, line 314: Should we use a safe cast here and throw a more explicit error in case the node cannot be cast?

PR #732, opened 2024-05-07
feat(model-server): report errors as application/problem+json

languitar:
This commit modifies the v2 REST API to always report errors as application/problem+json instances following RFC 9475. We do not declare this as a breaking change as we never guaranteed a certain error format.

The OpenAPI specification has been extended with a default error response for all routes to avoid having to repeat the definition many times.

The installation of the StatusPages extension had to be moved to the model-server Main file as this extension is now used to map instances of the new HttpException class to suitable responses.

HttpException has been introduced due to the lack of a standard HTTP exception hierarchy in ktor. Throwing exceptions instead of call.respondXXX usually makes route implementations shorter and easier to read.

### DIFF
--- a/model-server/src/main/kotlin/org/modelix/model/server/handlers/ModelReplicationServer.kt
+++ b/model-server/src/main/kotlin/org/modelix/model/server/handlers/ModelReplicationServer.kt
@@ -19,6 +19,7 @@ import io.ktor.http.HttpStatusCode
 import io.ktor.server.application.Application
 import io.ktor.server.application.ApplicationCall
 import io.ktor.server.application.call
+import io.ktor.server.plugins.BadRequestException
 import io.ktor.server.plugins.origin
 import io.ktor.server.request.acceptItems
 import io.ktor.server.request.receive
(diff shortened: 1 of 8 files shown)


model-server/src/main/kotlin/org/modelix/model/server/handlers/ModelReplicationServer.kt
```diff
@@ -72,6 +73,31 @@ import org.modelix.model.server.store.LocalModelClient
...
+    )
+
+private class BadRequestException(details: String, typeSuffix: String, cause: Throwable? = null) : HttpException(
+    HttpStatusCode.BadRequest,
+    title = "Bad request",
+    details = details,
+    type = "/problems/$typeSuffix",
+    cause = cause,
+)
+
```
mhuster23, line 100: Can we move these Exceptions into a separate file? They are not specific to `ModelReplicationServer` and could be reused in other places like the `ContentExplorer`.

languitar: These exceptions are closely related to the REST API. So they would need to stay in the `handlers` package.

mhuster23 (approved):

PR #751, opened 2024-05-14
chore: reduce warnings

languitar:
This PR tries to get rid of some of the easy to solve compiler warnings. With all the warnings, it becomes more and more tedious to actually spot real issues in the logs.


model-server/src/test/kotlin/org/modelix/model/server/ReplicatedRepositoryTest.kt
```diff
@@ -437,9 +439,9 @@ class ReplicatedRepositoryTest {
             nonMergedVersions.remove(versionToMerge)
         }
 
-        for (i in 1..100) {
+        (1..100).forEach {
```
mhuster23, line 442: We should probably keep the regular for loop to conform with the style guide.

model-server/src/main/kotlin/org/modelix/model/server/handlers/HistoryHandler.kt
```diff
@@ -169,7 +178,14 @@ class HistoryHandler(val client: IModelClient, private val repositoriesManager:
         skip: Int,
         limit: Int,
     ) {
-        val headVersion = if (headHash == null || headHash.length == 0) latestVersion else CLVersion(headHash, client.storeCache!!)
+        val headVersion = if (headHash == null || headHash.length == 0) {
```
mhuster23, line 181: `headHash.isNullOrEmpty()`

model-server/src/test/kotlin/org/modelix/model/server/ReplicatedRepositoryTest.kt
languitar: Right, the change was older than our discussion about the style guide. I am still confused about that rule ;-)

languitar: Replaced with `repeat`, which is hopefully even more readable now.

model-server/src/test/kotlin/org/modelix/model/server/ReplicatedRepositoryTest.kt
```diff
@@ -108,7 +108,9 @@ class ReplicatedRepositoryTest {
             }
             branchToChange.runWrite {
                 val changeGenerator = RandomModelChangeGenerator(branchToChange.getRootNode(), rand)
-                for (i in 1..1000) changeGenerator.applyRandomChange()
+                (1..1000).forEach { _ ->
```
mhuster23, line 111:
```suggestion
                repeat(1000) { _ ->
```

mhuster23 (approved):

PR #754, opened 2024-05-17
feat(model-client): let the user set useRoleIds in the initRepository call

benedekh:
The `initRepository` endpoint of the model server has an optional useRoleIds parameter. Its default value is true on the server.

ModelClientV2 has an initRepository method that does not accept the useRoleIds parameter, so it sends the request without it. It may cause problem for the user, if they create a repository via the modelClient V2, and then load/read the model assuming it has role names instead of role IDs.

This PR adds the (optional) useRoleIds parameter to that method. The default value is true.


slisson (approved):
It doesn't hurt to have this option, but I consider this a workaround for a bug that should be fixed instead.

PR #755, opened 2024-05-17
fix(model-client): sort the uploadable objects to avoid freezing the model server

benedekh:
Fixes [MODELIX-910](https://issues.modelix.org/issue/MODELIX-910).


slisson (approved):
Since the reason for the deadlock are the Ignite transactions the proper place to fix it is the `IgniteStoreClient`. I added this server-side fix.

I also changed the client-side sorting to only sort the chunk and not the whole sequence, which should save some memory and CPU time.

PR #786, opened 2024-06-04
MODELIX-708: find a configuration for open api generator where the compiler helps finding missing implementations

languitar:
Picks up #526 by adapting the OpenAPI generator templates such that abstract base classes for the different APIs are generated. This ensures that the implementation can't diverge much from the spec.

Notable mentions:

* I've dropped useless OpenAPI specs for things that cannot be formalized via OpenAPI
* A v1 spec has been extracted to free the v2 spec from these concerns
* Where possible, handlers where implemented via the generated base classes
* The generation / retrieval of ID has been extracted into a separate API as an example of how we can continue with the remaining routes later.
* The duplication of the Problem data type and having to serve multiple Swagger UIs is not ideal. I'll try to resolve this in a follow-up PR.

There would be more to do to get to a clean state. However, I want to prevent this PR from growing with increased likelihood for conflicts. Everything is functional, so this should be fine to merge in the current state.

Review is probably easiest on a per-commit basis.

* [ ] Code is readable


model-server/src/main/resources/openapi/templates/api.mustache
```diff
@@ -0,0 +1,64 @@
...
+        {{#operations}}
+            {{#operation}}
+            install_{{operationId}}()
+            {{/operation}}
+        {{/operations}}
+    }
+
+    {{#operations}}
+    {{#operation}}
+    open protected fun Route.install_{{operationId}}() {
```
mhuster23, line 57:
```suggestion
    protected open fun Route.install_{{operationId}}() {
```
Non-canonical modifier order.

model-server/src/main/kotlin/org/modelix/model/server/handlers/IdsApiImpl.kt
```diff
@@ -0,0 +1,72 @@
...
+ * Implementation of the REST API that is responsible for handling client and server IDs.
+ */
+class IdsApiImpl(
+    private val repositoriesManager: IRepositoriesManager,
+    private val modelClient: LocalModelClient,
+) : IdsApi() {
+
+    constructor(modelClient: LocalModelClient) : this(RepositoriesManager(modelClient), modelClient)
+    constructor(storeClient: IStoreClient) : this(LocalModelClient(storeClient))
+    constructor(repositoriesManager: RepositoriesManager) : this(repositoriesManager, repositoriesManager.client)
```
mhuster23, line 42: Do we really need that many constructors here?

model-server/src/main/kotlin/org/modelix/model/server/handlers/IdsApiImpl.kt
```diff
@@ -0,0 +1,72 @@
...
+
+    private val storeClient: IStoreClient get() = modelClient.store
```
mhuster23, line 44: This seems to be the only usage of `modelClient`. Does it make sense to replace `modelClient` with `storeClient` in the constructor?

model-server/src/main/kotlin/org/modelix/model/server/handlers/explorer/ContentExplorer.kt
mhuster23, line 1: `RepositoryOverview` is also an HTML handler. If we want to make the separation, we should probably also separate it from the other request handlers like `ContentExplorer`.

mhuster23 (commented):
Looks good overall. I left a few thoughts.

model-server/src/main/kotlin/org/modelix/model/server/handlers/IdsApiImpl.kt
languitar: This pattern is currently copied from ModelReplicationServer, where this code comes from. In tests, the various forms of the constructors are used. One intermediate version is used to avoid instantiating the LocalModelClient twice and I am not aware of a different syntax to avoid this construct.

Generally, I'd like to get rid of most of this by replacing the hand-crafted test registration with a real ktor module. But I'd add that in a separate PR.

languitar: This also addresses the previous comment for the constructors.

model-server/src/main/kotlin/org/modelix/model/server/handlers/explorer/ContentExplorer.kt
languitar: Missed that. Moved aswell and renamed the package to "html".

model-server/src/main/resources/openapi/templates/api.mustache
languitar: Fixed

model-server/src/main/kotlin/org/modelix/model/server/handlers/explorer/ContentExplorer.kt
mhuster23: The package declarations don't match the directory structure now.

languitar: what, that was an automated IntelliJ refactoring...

languitar: fixed now

mhuster23 (approved):

PR #791, opened 2024-06-06
feat(model-server): log the cause of internal server errors

odzhychko:
Log the cause of internal server errors.
This should help understand internal server errors better.
It gives users the possibility to a stack trace to bug reports.

* [x] Code is readable

### DIFF
--- a/model-server/src/main/kotlin/org/modelix/model/server/Main.kt
+++ b/model-server/src/main/kotlin/org/modelix/model/server/Main.kt
@@ -310,6 +310,8 @@ object Main {
             }
 
             exception<Throwable> { call, cause ->
+                // Such errors were most likely not caught and logged anywhere else.
+                LOG.error("Encountered an internal server error.", cause)
                 call.respondProblem(
                     Problem(
                         title = "Internal server error",


model-server/src/main/kotlin/org/modelix/model/server/Main.kt
```diff
@@ -310,6 +310,8 @@ object Main {
             }
 
             exception<Throwable> { call, cause ->
+                // Such errors were most likely not caught and logged anywhere else.
+                LOG.error("Encountered internal server error.", cause)
```
languitar, line 314:
```suggestion
                LOG.error("Encountered an internal server error.", cause)
```

odzhychko: thx

languitar:
Do we want to test this somehow? From a user perspective, being able to - for instance - install log alerting with Loki, is a good addition and we should ensure that this doesn't break again accidentally.

odzhychko:
I see the benefit of treating it as a stable feature, but for me, this is out-of-scope for this PR.
Especially, I do not want to decide now the way we want to log such messages or any other details.

As the first step, I just want to log the exception to make debugging of "Internal Server" errors easier. This would have helped with finding #795 faster.

slisson (approved):

languitar (approved):
I would have been fine with just asserting that we log something in case of an internal server error. But fine for the moment.

PR #792, opened 2024-06-06
fix(model-server): fix `IgniteStoreClient.getAll` not returning values for existing entries

odzhychko:
With different field order in Ignite configuration and key type `ObjectInRepository`, `IgniteCache.getAll` returned null values because keys were serialized differently. 
This only happened when `IgniteCache.getAll` was called with more than one key and the keys were not loaded into the cache. 
The serialization of a key return from a read-through operation was different from the serialization of a key created from an `ObjectInRepository` object.

* [x] Code is readable

### DIFF
--- a/model-server-test/src/test/kotlin/IgnitePostgresTest.kt
+++ b/model-server-test/src/test/kotlin/IgnitePostgresTest.kt
@@ -21,6 +21,7 @@ import kotlin.test.AfterTest
 import kotlin.test.BeforeTest
 import kotlin.test.Test
 import kotlin.test.assertEquals
+import kotlin.test.assertTrue
 
 class IgnitePostgresTest {
     lateinit var store: IgniteStoreClient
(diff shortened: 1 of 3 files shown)


model-server-test/src/test/kotlin/IgnitePostgresTest.kt
```diff
@@ -35,6 +36,18 @@ class IgnitePostgresTest {
         store.dispose()
     }
 
+    @Test
+    fun `load multiple keys into cache`() {
+        // The actual keys for are irrelevant.
```
languitar, line 41: grammar

model-server-test/src/test/kotlin/IgnitePostgresTest.kt
```diff
@@ -35,6 +36,18 @@ class IgnitePostgresTest {
         store.dispose()
     }
 
+    @Test
+    fun `load multiple keys into cache`() {
```
languitar, line 40: Can this be a bit more explicit on which condition should be fulfilled? Currently, I find it hard to understand what is actually tested here.

model-server/src/main/kotlin/org/modelix/model/server/store/ObjectInRepository.kt
```diff
@@ -16,6 +16,7 @@
 
 package org.modelix.model.server.store
 
+// Order of fields must be consistent with the configuration of JdbcType in ignite.xml
```
languitar, line 19: Let's just use this change to create a proper kdoc for this class?

model-server-test/src/test/kotlin/IgnitePostgresTest.kt
odzhychko: Done

model-server-test/src/test/kotlin/IgnitePostgresTest.kt
odzhychko: Done.
Is it now better to understand?

model-server/src/main/kotlin/org/modelix/model/server/store/ObjectInRepository.kt
odzhychko: Done

PR #815, opened 2024-06-11
perf(model-server): inefficient SQL query was generated for getAll

slisson:
After adding the repository column to the primary key, a different SQL query was generated that is less efficient. Adjusted the PostgresDialect to generate a query with comparable performance to the one for single column primary keys.


languitar (approved):

languitar:
Btw, thanks for the detailed comments in the PR. This intricate change can be understood pretty well with the explanations.

PR #793, opened 2024-06-07
build(ts-model-api): remove dependencies on configuration files from The Really Project

odzhychko:
The Really Project is not suited to provide default configuration files because those configuration files are specific to that project.

I inlined and merged the configurations from The Really Project (v2.0.0). 
See https://github.com/reallyland/tsconfig/blob/v2.0.0/tsconfig.json

## To be verified by reviewers

* [x] Relevant public API members have been documented
* [x] Documentation related to this PR is complete
  * [x] Boundary conditions are documented
  * [x] Exceptions are documented
  * [x] Nullability is documented if used
* [x] Touched existing code has been extended with documentation if missing
* [x] Code is readable
* [x] New features and fixed bugs are covered by tests

PR #816, opened 2024-06-12
build(model-api-gen-gradle): remove dependencies on configuration files from The Really Project

odzhychko:
The Really Project is not suited to provide default configuration files because those configuration files are specific to that project.

I inlined and merged the configurations from The Really Project (v2.0.0). See https://github.com/reallyland/tsconfig/blob/v2.0.0/tsconfig.json

This will make #810 obsolete.

* [x] Code is readable


model-api-gen-gradle-test/typescript-generation/tsconfig.json
languitar, line 1: What about a symlink to the config from this PR? https://github.com/modelix/modelix.core/pull/793 I don't think we want different styles per module.

odzhychko: Aside from TS-configuration in `ts-model-api` (#793) and `model-api-gen-gradle/typescript-generation` (here) we have `vue-model-api` and `model-api-gen-gradle/vue-integration`. All have different configurations... consolidating it was at the back of my mind. Some considerations by the way are:

*  We want to replace `ts-model-api` by Kotlin
* `model-api-gen-gradle/typescript-generation` is a test and not somehting we publish
* `model-api-gen-gradle/vue-integration` is also a test

> I don't think we want different styles per module.

I agree.

> What about a symlink to the config from this PR?

I am not yet at the stage, where I want to start consolidating our NPM configurations.
Before we start to share configurations (e.g. by symlinking or importing) we would need to start updating version of packages together

languitar (approved):

model-api-gen-gradle-test/typescript-generation/tsconfig.json
odzhychko: I just created it for transparency.
https://issues.modelix.org/issue/MODELIX-939/Unify-dependencies-and-configurations-across-NPM-packages

PR #818, opened 2024-06-13
MODELIX-856 Delete its data when a repository is deleted

mhuster23:
Currently we only remove the repository id from the list of available repositories when deleting a repository.
The data still remains in the cache and the database.
With these changes we also delete the objects of a repository from the cache and the database of the model-server, when removing a repository.

* [ ] Code is readable


model-server-test/build.gradle.kts
```diff
@@ -6,9 +6,12 @@ plugins {
 dependencies {
     testImplementation(kotlin("test"))
     testImplementation(project(":model-server"))
+    testImplementation(project(":model-server").dependencyProject.sourceSets.test.get().runtimeClasspath)
```
languitar, line 9: What does this do? Could benefit from a comment.

model-server-test/src/test/kotlin/RepositoriesManagerWithDatabaseTest.kt
```diff
@@ -0,0 +1,98 @@
...
+        runBlocking {
+            repoManager.removeRepository(existingRepo)
+            repoManager.removeRepository(repoToBeDeleted)
+        }
+        getDbConnection().use { connection ->
+            connection.prepareStatement("DELETE FROM modelix.model WHERE repository IN (?,?)").apply {
+                setString(1, existingRepo.id)
+                setString(2, repoToBeDeleted.id)
+                execute()
+                close()
```
languitar, line 44: Why is this `close` needed. Doesn't the prepared statement also implement AutoCloseable? https://docs.oracle.com/javase/tutorial/jdbc/basics/prepared.html

model-server-test/src/test/kotlin/RepositoriesManagerWithDatabaseTest.kt
```diff
@@ -0,0 +1,98 @@
...
+            }
+        }
+    }
+
+    @Test
+    fun `database does not contain removed data that was not part of the cache`() = runTest {
```
languitar, line 50: What is "that" here?

model-server-test/src/test/kotlin/RepositoriesManagerWithDatabaseTest.kt
```diff
@@ -0,0 +1,98 @@
...
+            repoManager.removeRepository(repoToBeDeleted)
+
+            val statement = connection.prepareStatement("SELECT * FROM modelix.model WHERE repository = ?").apply {
+                setString(1, repoToBeDeleted.id)
+            }
+
+            statement.use {
+                val result = it.executeQuery()
+                assertFalse("Database contained leftover repository data.") { result.isBeforeFirst }
+            }
```
languitar, line 70:
```suggestion
            connection.prepareStatement("SELECT * FROM modelix.model WHERE repository = ?").use {
                setString(1, repoToBeDeleted.id)
                val result = executeQuery()
                assertFalse("Database contained leftover repository data.") { result.isBeforeFirst }
            }
```

Looks a bit weird that the assert block is split up artificially.

model-server-test/src/test/kotlin/RepositoriesManagerWithDatabaseTest.kt
```diff
@@ -0,0 +1,98 @@
...
+import org.junit.jupiter.api.Test
+import org.modelix.model.lazy.RepositoryId
+import org.modelix.model.server.handlers.RepositoriesManagerTest
+import org.modelix.model.server.store.IgniteStoreClient
+import java.sql.DriverManager
+import kotlin.test.AfterTest
+import kotlin.test.assertFalse
+import kotlin.test.assertTrue
+
+class RepositoriesManagerWithDatabaseTest : RepositoriesManagerTest(IgniteStoreClient()) {
```
languitar, line 27: This lacks some corner cases:

* Deleting a repo without any data
* Tests for error handling when a query fails

model-server/src/main/kotlin/org/modelix/model/server/handlers/RepositoriesManager.kt
```diff
@@ -234,12 +236,12 @@ class RepositoriesManager(val client: LocalModelClient) : IRepositoriesManager {
...
                 putVersionHash(repository.getBranchReference(branchName), null)
             }
-            store.getGenericStore().put(branchListKey(repository), null)
+            genericStore.put(branchListKey(repository), null)
             val isolated = checkNotNull(isIsolated(repository)) { "Repository not found: $repository" }
             val existingRepositories = getRepositories(isolated)
             val remainingRepositories = existingRepositories - repository
             store.put(repositoriesListKey(isolated), remainingRepositories.joinToString("\n") { it.id })
-
+            genericStore.removeRepositoryObjects(repository)
```
languitar, line 244: How expensive is it to do all of this inside the transaction?

model-server/src/main/kotlin/org/modelix/model/server/store/IGenericStoreClient.kt
```diff
@@ -17,8 +17,21 @@
 package org.modelix.model.server.store
 
 import org.modelix.model.IGenericKeyListener
+import org.modelix.model.lazy.RepositoryId
 
-typealias IsolatingStore = IGenericStoreClient<ObjectInRepository>
+interface IsolatingStore : IGenericStoreClient<ObjectInRepository> {
```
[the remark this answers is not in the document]
languitar: Could be addressed in the PR now

model-server/src/main/kotlin/org/modelix/model/server/store/IGenericStoreClient.kt
```diff
@@ -17,8 +17,21 @@
...
+    /**
+     * Default implementation for removing repository objects.
+     * May be overridden by more efficient, store-specific implementations.
```
languitar, line 25: Should include valid exceptions to be raised when removing fails.

model-server/src/main/kotlin/org/modelix/model/server/store/IgniteStoreClient.kt
```diff
@@ -103,6 +112,43 @@ class IgniteStoreClient(jdbcConfFile: File? = null, inmemory: Boolean = false) :
...
+
+        val filter = IgniteBiPredicate<ObjectInRepository, String?> { key, _ ->
+            key.getRepositoryId() == repositoryId.id
+        }
+        val transformer = IgniteClosure<Cache.Entry<ObjectInRepository, String?>, ObjectInRepository?> { entry ->
+            entry.key
+        }
+        val query = ScanQuery(filter)
+        val toDelete = cache.query(query, transformer).all.asSequence().filterNotNull()
+            .toSortedSet() // sorting is necessary to avoid deadlocks
```
languitar, line 128: Why can deadlocks occur in case it's not sorted? Some reference to documentation explaining this would be good.

model-server/src/main/kotlin/org/modelix/model/server/store/IgniteStoreClient.kt
```diff
@@ -103,6 +112,43 @@ class IgniteStoreClient(jdbcConfFile: File? = null, inmemory: Boolean = false) :
...
+        LOG.info { "Deleting cache entries asynchronously. count=${toDelete.size}" }
+        cache.removeAllAsync(toDelete).listen { LOG.info { "Cache entries deleted." } }
+    }
+
+    private fun removeRepositoryObjectsFromDatabase(repositoryId: RepositoryId) {
+        require(!inmemory) { "Cannot remove from database in inmemory mode." }
```
languitar, line 134:
```suggestion
        require(!inmemory) { "Cannot remove from database in in-memory mode." }
```

model-server/src/main/kotlin/org/modelix/model/server/store/IgniteStoreClient.kt
```diff
@@ -103,6 +112,43 @@ class IgniteStoreClient(jdbcConfFile: File? = null, inmemory: Boolean = false) :
         return cache.associate { it.key to it.value }
     }
 
+    override fun removeRepositoryObjects(repositoryId: RepositoryId) {
+        if (!inmemory) {
+            removeRepositoryObjectsFromDatabase(repositoryId)
+        }
```
languitar, line 118: It would probably be good to document here why deletion happens twice (when not in in-memory mode) and also why this order of operations is safe (why doesn't ignite try to access the keys that are already deleted now?)

model-server/src/main/kotlin/org/modelix/model/server/store/IgniteStoreClient.kt
```diff
@@ -103,6 +112,43 @@ class IgniteStoreClient(jdbcConfFile: File? = null, inmemory: Boolean = false) :
...
+        require(!inmemory) { "Cannot remove from database in inmemory mode." }
+        LOG.info { "Removing repository objects from database." }
+
+        dataSource.connection.use {
+            val statement = it.prepareStatement("DELETE from model WHERE repository = ?").apply {
+                setString(1, repositoryId.id)
+            }
+            try {
+                val deletedRows = statement.executeUpdate()
+                LOG.info { "Deleted rows from database. count=$deletedRows" }
```
languitar, line 143: The meeting notes for the coding conventions workshop proposed this format: "Transaction failed [transaction=42]"

model-server/src/main/kotlin/org/modelix/model/server/store/IgniteStoreClient.kt
```diff
@@ -103,6 +112,43 @@ class IgniteStoreClient(jdbcConfFile: File? = null, inmemory: Boolean = false) :
...
+            } catch (e: SQLException) {
+                LOG.error { e }
+            } finally {
+                statement.close()
```
languitar, line 147: Should probably be converted to a `use` call instead of explicit `close`.

model-server/src/test/kotlin/org/modelix/model/server/handlers/RepositoriesManagerTest.kt
```diff
@@ -0,0 +1,62 @@
...
+import org.modelix.model.server.store.LocalModelClient
+import kotlin.test.AfterTest
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertTrue
+
+class RepositoriesManagerWithInMemoryStoreClientTest : RepositoriesManagerTest(InMemoryStoreClient())
+class RepositoriesManagerWithIgniteStoreClientTest : RepositoriesManagerTest(IgniteStoreClient(inmemory = true))
+
+abstract class RepositoriesManagerTest(val store: IsolatingStore) {
```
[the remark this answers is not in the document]
languitar: Should we put an explicit suppression explaining why this pattern is chosen (probably to prevent junit from executing the test?)?

model-server/src/test/kotlin/org/modelix/model/server/handlers/RepositoriesManagerTest.kt
```diff
@@ -0,0 +1,62 @@
...
+    protected val repoManager = RepositoriesManager(LocalModelClient(store))
+    protected fun initRepository(repoId: RepositoryId) {
+        runBlocking {
+            repoManager.createRepository(repoId, "testUser", useRoleIds = true, legacyGlobalStorage = false)
+        }
+    }
```
languitar, line 24: This could simply be a suspend function as it is always called from a coroutine?

model-server/src/test/kotlin/org/modelix/model/server/handlers/RepositoriesManagerTest.kt
languitar, line 1: I think the test structure here can be improved, which would also get rid of the abstract test class. The repositories manager is only responsible for invoking the deletion of data. The actual execution of the deletion is the responsibility of each store implementation. Therefore I'd propose:

* Move tests for whether deleting works and prevents deleting other repo's data into tests for each store implementation. Seems we still lack a test for the in-memory store for this...
* Use a mock store in the RepositoriesManagerTest and simply assert that the implementation properly dispatches removal to the store implementation. If we do not have a mocking framework installed yet, I'd propose [mockk](https://mockk.io/).

languitar (changes requested):
Some nitpicks for corner-cases and documentation and a proposal for an improved test structure. Apart from that, this generally looks fine to me.

model-server/src/main/kotlin/org/modelix/model/server/store/IGenericStoreClient.kt
mhuster23: We currently don't explicitly throw custom exceptions in any of our StoreClients. We should probably add some to wrap store-specific ones (ignite and sql exceptions) but that would require [MODELIX-799](https://issues.modelix.org/issue/MODELIX-799/Create-a-proposal-on-how-to-organize-exceptions). After that we should apply it to this and the already existing operations.

model-server/src/main/kotlin/org/modelix/model/server/handlers/RepositoriesManager.kt
mhuster23: I don't know how to quantify that. In the long run repository objects should be stored in a separate cache without transactions, as the data is conceptually immutable (but removable).

languitar: How long does it take for our big customer? The question is if it's valid to do all that in the transaction and also in the scope of the REST request.

model-server/src/test/kotlin/org/modelix/model/server/handlers/RepositoriesManagerTest.kt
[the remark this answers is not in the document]
mhuster23: outdated

model-server-test/src/test/kotlin/RepositoriesManagerWithDatabaseTest.kt
mhuster23: fixed

model-server-test/src/test/kotlin/RepositoriesManagerWithDatabaseTest.kt
mhuster23: outdated

model-server-test/src/test/kotlin/RepositoriesManagerWithDatabaseTest.kt
mhuster23: fixed

model-server/src/main/kotlin/org/modelix/model/server/handlers/RepositoriesManager.kt
mhuster23: Haven't run into issues yet. We can follow up on it, if it becomes a problem.

model-server/src/main/kotlin/org/modelix/model/server/store/IgniteStoreClient.kt
mhuster23: added a hint

model-server/src/main/kotlin/org/modelix/model/server/store/IgniteStoreClient.kt
mhuster23: fixed

model-server/src/main/kotlin/org/modelix/model/server/store/IgniteStoreClient.kt
mhuster23: done

model-server/src/main/kotlin/org/modelix/model/server/store/IgniteStoreClient.kt
mhuster23: done

model-server/src/main/kotlin/org/modelix/model/server/store/IgniteStoreClient.kt
mhuster23: done

model-server/src/test/kotlin/org/modelix/model/server/handlers/RepositoriesManagerTest.kt
mhuster23: fixed

model-server/src/test/kotlin/org/modelix/model/server/handlers/RepositoriesManagerTest.kt
mhuster23: done

model-server-test/src/test/kotlin/IgnitePostgresRepositoryRemovalTest.kt
```diff
@@ -0,0 +1,127 @@
...
+
+    @Test
+    fun `removal does not affect other repository data in the cache`() {
+        val existingEntries = mapOf(
+            ObjectInRepository(existing.id, "key0") to "value0",
+            ObjectInRepository(existing.id, "key1") to "value1",
+        )
+        store.putAll(existingEntries)
+
+        val toDeleteId = RepositoryId("cache-other-removal")
```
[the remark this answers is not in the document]
languitar: This variable can probably be removed then

model-server/src/main/kotlin/org/modelix/model/server/store/IgniteStoreClient.kt
```diff
@@ -18,19 +18,25 @@
...
 import java.io.IOException
+import java.sql.SQLException
 import java.util.*
+import javax.cache.Cache
 import javax.sql.DataSource
 
 private val LOG = KotlinLogging.logger { }
 
-class IgniteStoreClient(jdbcConfFile: File? = null, inmemory: Boolean = false) : IsolatingStore, AutoCloseable {
+class IgniteStoreClient(jdbcConfFile: File? = null, private val inmemory: Boolean = false) : IsolatingStore, AutoCloseable {
```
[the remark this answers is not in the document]
languitar: > Touched existing code has been extended with documentation if missing

Can we address this here?

model-server/src/main/kotlin/org/modelix/model/server/store/InMemoryStoreClient.kt
```diff
@@ -103,7 +103,12 @@
                 try {
                     transactionValues = HashMap()
                     val result = body()
-                    values.putAll(transactionValues!!)
+                    val puts = transactionValues!!.filterValues { it != null }
```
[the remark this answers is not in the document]
languitar: How can `transactionValues` actually ever be `null` here?

model-server-test/src/test/kotlin/IgnitePostgresRepositoryRemovalTest.kt
```diff
@@ -0,0 +1,127 @@
...
+import org.modelix.model.server.store.IgniteStoreClient
+import org.modelix.model.server.store.ObjectInRepository
+import java.sql.DriverManager
+import kotlin.test.AfterTest
+import kotlin.test.assertEquals
+import kotlin.test.assertFalse
+import kotlin.test.assertTrue
+
+class IgnitePostgresRepositoryRemovalTest {
+    private fun getDbConnection() = DriverManager.getConnection(System.getProperty("jdbc.url"), "modelix", "modelix")
```
languitar, line 28: Why do we create a new connection on every access? Could be kept alive for all test cases.

model-server/src/main/kotlin/org/modelix/model/server/store/IGenericStoreClient.kt
languitar: Not totally happy to ignore this right now as the other task will grow with every change we make this way. But ok

model-server/src/main/kotlin/org/modelix/model/server/store/IgniteStoreClient.kt
```diff
@@ -103,6 +112,44 @@ class IgniteStoreClient(jdbcConfFile: File? = null, inmemory: Boolean = false) :
         return cache.associate { it.key to it.value }
     }
 
+    override fun removeRepositoryObjects(repositoryId: RepositoryId) {
+        if (!inmemory) {
+            // Not all entries are in the cache. We delete them directly instead of loading them into the cache first.
+            // This should be safe as the repository has already been removed from the list of available ones.
```
languitar, line 118: Let's explicitly document this contract in the interface kdoc for `removeRepositoryObjects` a la: "callers need to ensure that repositories are not usable anymore before calling this method". This increases the chance that this fact is not forgotten during the next refactoring.

languitar (dismissed):
A few nit-picks and doc improvements. But otherwise that looks fine to me and even more concise now.

model-server/src/main/kotlin/org/modelix/model/server/store/InMemoryStoreClient.kt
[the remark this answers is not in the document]
mhuster23: the `body` lambda could set it to `null` as it is a global variable

model-server-test/src/test/kotlin/IgnitePostgresRepositoryRemovalTest.kt
mhuster23: fixed

model-server/src/main/kotlin/org/modelix/model/server/store/IgniteStoreClient.kt
mhuster23: fixed

model-server/src/main/kotlin/org/modelix/model/server/store/IgniteStoreClient.kt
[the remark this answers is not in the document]
mhuster23: done

model-server/src/main/kotlin/org/modelix/model/server/store/InMemoryStoreClient.kt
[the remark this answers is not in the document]
languitar: so we should probably use `requireNotNull` then in case this must not happen.

languitar:
I can't find the one conversation that's supposed to still be open. The GitHub UI doesn't show it in the discussions and cannot navigate to it in the code review.

languitar (dismissed):

languitar (approved):

PR #819, opened 2024-06-17
feat(model-api): check if is sub-concept but only based on the IConceptReference

benedekh:
In MPS, we mostly have MPSConcepts, SimpleConcepts and GeneratedConcepts at runtime. 

If we use MPSConcept, then the IConcept.isSubConceptOf assumes that the superConcept is an MPSConcept. If not, then the method returns false. This makes it difficult to check if this and superConcept are the same superConcept if is a SimpleConcept. 

GeneratedConcept.isSubConceptOf checks the equality with the superConcept, which means if the superConcept is a different object that structurally may represent the same Concept then it returns false (e.g. a GeneratedConcept and a SimpleConcept with the same Concept UID and name would return false). 

In order to simplify all these cases we could just check the UID of the Concept and assume that if UID matches then the Concepts should also be the same. This implementation is similar to how SConceptAdapterById checks the isSubConcept in MPS. This adapter is often used in MPS, that is wrapped by MPSConcept at runtime.

* [x] Code is readable

### DIFF
--- a/model-api/src/commonMain/kotlin/org/modelix/model/api/IConcept.kt
+++ b/model-api/src/commonMain/kotlin/org/modelix/model/api/IConcept.kt
@@ -181,3 +181,22 @@ interface IConcept {
 fun IConcept?.isSubConceptOf(superConcept: IConcept?) = this?.isSubConceptOf(superConcept) == true
 
 fun IConcept.conceptAlias() = getConceptProperty("alias")
+
+/**
+ * Checks if this is a sub-concept of the [IConcept] that is identified by the [superConceptReference]'s UID.
+ *
+ * @param superConceptReference a reference to the potential super-concept
+ * @return true if this concept (or any of its ancestors) has the same UID as the [superConceptReference]
+ */
+fun IConcept.isSubConceptOf(superConceptReference: IConceptReference): Boolean {
+    if (this.getUID() == superConceptReference.getUID()) {
+        return true
+    } else {
+        for (parent in getDirectSuperConcepts()) {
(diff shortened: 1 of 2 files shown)


model-api/src/commonMain/kotlin/org/modelix/model/api/IConcept.kt
```diff
@@ -181,3 +181,22 @@ interface IConcept {
...
 
 fun IConcept.conceptAlias() = getConceptProperty("alias")
+
+/**
+ * Checks if this is a sub-concept of [superConcept], based on the Concepts' UID.
+ *
+ * @param superConcept the potential super-concept
+ * @return true if this concept (or any of its ancestors) has the same UID as the [superConcept]
+ */
+fun IConcept.isSubConceptOfBasedOnUid(superConcept: IConcept): Boolean {
```
benedekh, line 191:
```suggestion
fun IConcept.isSubConceptOfBasedOnUid(superConceptReference: IConceptReference): Boolean {
```
And then compare only by the UID.

benedekh: fixed in [f70040c](https://github.com/modelix/modelix.core/pull/819/commits/f70040c97d79bd6a0a9e77383c5bdc7844b3918f)

benedekh:
I've fixed the [suggestion](https://github.com/modelix/modelix.core/pull/819#discussion_r1645899267) we had in the refinement today.

@modelix/developers could you review this PR, please?

odzhychko (approved):

PR #825, opened 2024-06-18
feat(model-api-gen): allow passing nullish values to concept type guards

drik98:
- Also added tests to `model-api-gen-gradle-test/typescript-generation` to verify the changes.

## Rationale

With the SingleChildAccessor you often find yourself writing code similar to this.

```typescript
const child = foo.bar.get();
if(child && isOfConcept_Child(child) {
    // do your stuff
}
```

This change simplifies it so no additional null checking is needed anymore:

```typescript
const child = foo.bar.get();
if(isOfConcept_Child(child) {
    // do your stuff
}
```

TODO description of the PR

* [ ] Code is readable

### DIFF
--- a/model-api-gen-gradle-test/typescript-generation/build.gradle.kts
+++ b/model-api-gen-gradle-test/typescript-generation/build.gradle.kts
@@ -6,8 +6,23 @@ plugins {
     id("org.modelix.model-api-gen") apply false
 }
 
+val updateDependencies = tasks.create<NpmTask>("updateDependencies") {
+    args.set(
+        listOf(
+            "install",
+            "../../model-client/build/npmDevPackage/model-client.tgz",
+            "../../ts-model-api",
+        ),
+    )
+}
+
+tasks.npmInstall {
+    dependsOn(updateDependencies)
+}
+
 tasks.named("npm_run_build") {
     dependsOn(":metamodel-export:generateMetaModelSources")
+    dependsOn("npm_run_test")
     inputs.dir(layout.buildDirectory.dir("typescript_src"))
     inputs.file("package.json")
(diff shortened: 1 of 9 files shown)


model-api-gen-gradle-test/typescript-generation/src/typeguards.test.ts
```diff
@@ -0,0 +1,49 @@
+import { useModelsFromJson } from "@modelix/vue-model-api";
```
odzhychko, line 1: You do not need the `vue-model-api` for tests

You should be able to implement a helper `useModelFromJson` in this test only using `@modelix/model-client`.
Take a look at `vue-model-api/src/useModelsFromJson.ts`.

---

This could by the way solve the build issue you have.
I think I had similar issues when testing `model-api-gen-gradle-test/vue-integration`.

model-api-gen-gradle-test/typescript-generation/build.gradle.kts
```diff
@@ -8,6 +8,7 @@ plugins {
 
 tasks.named("npm_run_build") {
     dependsOn(":metamodel-export:generateMetaModelSources")
+    dependsOn("npm_run_test")
```
odzhychko, line 11: To make the NPM install succeed, you need to add something like in here: https://github.com/modelix/modelix.core/blob/da591c859f031895661e78ee8e42f1d8b61aa424/model-api-gen-gradle-test/vue-integration/build.gradle.kts#L8

drik98: Thanks for the help!

odzhychko (approved):

PR #831, opened 2024-06-19
feat(bulk-model-sync): add a configuration option to load plugins

odzhychko:
In simple cases, `MPSBulkSynchronizer` does not rely on concepts (and in turn of languages and plugins) of the synced nodes to be loaded. 
Plugins might become necessary when they provide custom persistence and in other, yet unknown cases.

Depends on https://github.com/modelix/modelix.mps-build-tools/pull/52. 
After it is merged and published `libs.versions.toml` needs to be updated. But the API in bulk-sync-plugin can be reviewed anyway.

I did not add tests.
The test would be complicated, long-running (starts MPS) and only test rather simple logic.
When someone deems test necessary I would add tests, that run one of the sync tasks and assert through the logs, that the specified plugins are loaded.
If someone has a better idea/different opinion on the tests here, please let me know.

* [ ] Code is readable

### DIFF
--- a/bulk-model-sync-gradle/src/main/kotlin/org/modelix/model/sync/bulk/gradle/config/ModelSyncGradleSettings.kt
+++ b/bulk-model-sync-gradle/src/main/kotlin/org/modelix/model/sync/bulk/gradle/config/ModelSyncGradleSettings.kt
@@ -95,6 +95,25 @@ sealed interface LocalEndpoint : SyncEndpoint {
     var repositoryDir: File?
     var mpsDebugPort: Int?
 
+    /**
+     * Add a plugin to be loaded when running MPS.
+     * `jetbrains.mps.core`, `jetbrains.mps.testing`, `jetbrains.mps.ide.make` are always loaded.
+     *
+     * All other plugins, even bundled ones, must be configured explicitly.
+     * In general cases, the sync does not rely on concepts (and in turn on languages and plugins) of the synced nodes.
+     *
+     * Loading other plugins might become necessary when they provide custom persistence
+     * and in other, yet unknown cases.
+     * First, try if the sync works for your project without adding plugins.
+     *
+     * Example usage:
+     * ```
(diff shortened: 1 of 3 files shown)


odzhychko:
This PR ist still a draft, because I want to talk about in our refinment.
See https://issues.modelix.org/issue/MODELIX-813/Avoid-changing-resolveInfo-tags-for-unrelated-children#focus=Comments-4-3611.0-0

bulk-model-sync-gradle/src/main/kotlin/org/modelix/model/sync/bulk/gradle/config/ModelSyncGradleSettings.kt
```diff
@@ -95,6 +95,25 @@ sealed interface LocalEndpoint : SyncEndpoint {
...
+     * All other plugins, even bundled ones, must be configured explicitly.
+     * In general cases, the sync does not rely on concepts (and in turn on languages and plugins) of the synced nodes.
+     *
+     * Loading other plugins might become necessary when they provide custom persistence
+     * and in other, yet unknown cases.
+     * First, try if the sync works for your project without adding plugins.
+     *
+     * Example usage:
+     * ```
+     * mpsPlugin(BundledPluginSpec("jetbrains.mps.vcs", File("plugins/mps-vcs")))
```
languitar, line 111: The path for bundled plugins cannot be resolved automatically?

odzhychko: It would be possible to only specify the ID and search for the folder the contains the `META-INF/plugin.xml`with that ID.

But I wanted to keep it simple and explicit.
Also I am not sure if Plugins always must have a META-INF/plugin.xml

languitar (approved):

PR #833, opened 2024-06-19
feat(ts-model-api): add method for removing node

drik98:
This PR introduces convenience methods to the `ITypedNode` and `INodeJS` interfaces to easily remove any node from its parent in both the typed and untyped APIs.

Example usage: 

```typescript
const namedElement: INamedConcept = foo();
namedElement.remove();

// instead of
namedElement.unwrap().getParent()?.removeChild(namedElement.unwrap());
```

* [ ] Code is readable

### DIFF
--- a/model-api-gen-gradle-test/typescript-generation/src/node.test.ts
+++ b/model-api-gen-gradle-test/typescript-generation/src/node.test.ts
@@ -1,0 +1,9 @@
+import { useFakeNode } from "./test-helpers";
+
+test("typed nodes can be removed", () => {
+  const { typedNode, rootNode } = useFakeNode();
+  expect(rootNode.getChildren("children1")).toHaveLength(1);
+
+  typedNode.remove();
+  expect(rootNode.getChildren("children1")).toHaveLength(0);
+});
--- a/vue-model-api/src/internal/ReactiveINodeJS.ts
+++ b/vue-model-api/src/internal/ReactiveINodeJS.ts
@@ -84,6 +84,10 @@ export class ReactiveINodeJS implements INodeJS {
       : unreacitveNode;
   }
 
+  remove(): void {
+    this.unreactiveNode.remove();
+  }
+
   getChildren(role: string | undefined): INodeJS[] {
     const { track } = this.getOrCreateTrackAndTriggerForRole(role);
     track();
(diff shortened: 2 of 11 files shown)


drik98:
Or do you rather prefer it on the untyped API (`INodeJS`) instead? Usage:

```typescript
const namedElement: INamedConcept = foo();
namedElement.unwrap().remove();
```

odzhychko:
> Or do you rather prefer it on the untyped API (`INodeJS`) instead? Usage:
> 
> ```ts
> const namedElement: INamedConcept = foo();
> namedElement.unwrap().remove();
> ```

Having a `remove` method on both is fine.
I'm fine with this PR only implementing it for `ITypedNode`.

odzhychko:
> I am also open to adding tests but could not find the best spot for it. If you have an idea go ahead.

You can test it like in `model-api-gen-gradle-test/typescript-generation/src/typeguards.test.ts`.

vue-model-api/src/internal/ReactiveINodeJS.ts
```diff
@@ -84,6 +84,10 @@ export class ReactiveINodeJS implements INodeJS {
       : unreacitveNode;
   }
 
+  remove(): void {
+    this.unreactiveNode.remove();
```
drik98, line 88: Question: Do we have to invoke `track` here?

odzhychko: No, because the parent already `tracked` when you accessed the child. The parrent will `trigger` when the child is removed.
I see that you already verifyed it in `ReactiveINodeJS.test.ts`.

drik98: Thanks for the explanation :)

model-api-gen-gradle-test/typescript-generation/src/node.test.ts
```diff
@@ -0,0 +1,17 @@
+import { useFakeNode } from "./test-helpers";
+
+test("untyped nodes can be removed", () => {
```
odzhychko, line 3: `model-api-gen-gradle-test/typescript-generation/src/node.test.ts` should test only features of typed nodes.

The features of the untyped nodes should be tested in closer to the implementation in `modelix.core/model-api/src/jsTest`.

Could you move the test "untyped nodes can be removed" to there?

odzhychko (changes requested):

odzhychko (approved):

PR #837, opened 2024-06-20
fix(ts-model-api): new single childs can have correct concept

drik98:
Prior to this fix when using the `setNew` method of the `SingleChildAccessor` class, the created node had no concept and could not be given one. This PR introduces the possibility (while maintaining backwards compatibility) to provide a concept when invoking the method.

```typescipt
// before
const typedNode: BasePlaceholder = foo();
const childNode: IPlaceholderContent = typedNode.content.setNew(); // Actually UnknownTypedNode 
expect(isOfConcept_IPlaceholderContent(childNode)).toBeTruthy(); // <---- fails

// after
const typedNode: BasePlaceholder = foo();
const childNode: IPlaceholderContent = typedNode.content.setNew(C_IPlaceholderContent); 
expect(isOfConcept_IPlaceholderContent(childNode)).toBeTruthy(); // <---- succeeds
```

* [ ] Code is readable

### DIFF
--- a/ts-model-api/src/ChildrenAccessor.ts
+++ b/ts-model-api/src/ChildrenAccessor.ts
@@ -44,11 +44,11 @@ export class SingleChildAccessor<ChildT extends ITypedNode> extends ChildrenAcce
     return children.length === 0 ? undefined : children[0]
   }
 
-  public setNew(): ChildT {
+  public setNew(subconcept?: IConceptJS | undefined): ChildT {
     const existing = this.get();
     if (existing !== undefined) {
       existing.remove();
     }
-    return this.wrapChild(this.parentNode.addNewChild(this.role, 0, undefined))
+    return this.wrapChild(this.parentNode.addNewChild(this.role, 0, subconcept))
   }
 }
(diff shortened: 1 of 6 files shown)


ts-model-api/src/ChildrenAccessor.ts
```diff
@@ -21,34 +21,55 @@ export abstract class ChildrenAccessor<ChildT extends ITypedNode> implements Ite
 }
 
 export class ChildListAccessor<ChildT extends ITypedNode> extends ChildrenAccessor<ChildT> {
-  constructor(parentNode: INodeJS, role: string | undefined) {
-    super(parentNode, role);
+  constructor(parentNode: INodeJS, role: string | undefined, concept?: IConceptJS | undefined) {
```
slisson, line 24: You can just use `val conceptRef = concept?.getUID()?.let { ConceptReference(it) } ?: role?.let { node.concept?.getChildLink(it) }?.targetConcept` in `NodeAdapterJS.addNewChild` instead.

drik98:
After discussion with  @odzhychko  we will split this into two separate PRs. This PR will then only take care of the fix.

ts-model-api/src/ChildrenAccessor.ts
drik98: Wow that really seems much simpler. I am gonna adress this in https://github.com/drik98/modelix.core/pull/2

odzhychko (approved):

PR #858, opened 2024-06-27
MODELIX-781 Add confirmation modal to "delete repository" button

mhuster23:
This PR also changes the position of the delete button and adds "delete branch"-buttons for non-master branches.

* [ ] Code is readable

### DIFF
--- a/model-server/src/main/kotlin/org/modelix/model/server/handlers/ui/RepositoryOverview.kt
+++ b/model-server/src/main/kotlin/org/modelix/model/server/handlers/ui/RepositoryOverview.kt
@@ -9,12 +9,14 @@ import io.ktor.server.routing.routing
 import kotlinx.html.FlowContent
 import kotlinx.html.FlowOrInteractiveOrPhrasingContent
 import kotlinx.html.a
+import kotlinx.html.button
 import kotlinx.html.form
 import kotlinx.html.h1
 import kotlinx.html.i
 import kotlinx.html.onClick
 import kotlinx.html.p
 import kotlinx.html.postButton
+import kotlinx.html.script
 import kotlinx.html.span
 import kotlinx.html.table
 import kotlinx.html.tbody
@@ -23,6 +25,8 @@ import kotlinx.html.th
 import kotlinx.html.thead
 import kotlinx.html.title
 import kotlinx.html.tr
+import kotlinx.html.unsafe
+import org.modelix.model.lazy.RepositoryId
 import org.modelix.model.server.handlers.IRepositoriesManager
 import org.modelix.model.server.templates.PageWithMenuBar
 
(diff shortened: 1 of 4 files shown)


model-server/src/test/kotlin/org/modelix/model/server/handlers/RepositoriesManagerTest.kt
```diff
@@ -26,6 +26,16 @@ class RepositoriesManagerTest {
         store.close()
     }
 
+    @Test
+    fun `deleted branches are not listed anymore`() = runTest {
```
languitar, line 30: Not sure if the test case name really reflects what is actually tested. Isn't this more in the direction of "deleting the last branch works"?

model-server/src/main/kotlin/org/modelix/model/server/handlers/ui/RepositoryOverview.kt
```diff
@@ -98,24 +109,37 @@ class RepositoryOverview(private val repoManager: IRepositoriesManager) {
...
         }
     }
 }
+
+internal fun FlowContent.buildDeleteBranchForm(repositoryId: String, branchName: String) {
+    if (branchName == RepositoryId.DEFAULT_BRANCH) return
+    form {
+        deleteButton {
+            name = "delete"
+            formAction = "../v2/repositories/${repositoryId.encodeURLPathPart()}/branches/${branchName.encodeURLPathPart()}"
```
languitar, line 140: How does it work that the browser doesn't end up in the response of the REST call after submission?

mhuster23: fixed it with some javascript

model-server/src/test/kotlin/org/modelix/model/server/handlers/RepositoriesManagerTest.kt
mhuster23: renamed

model-server/src/main/kotlin/org/modelix/model/server/handlers/ui/RepositoryOverview.kt
```diff
@@ -31,7 +36,21 @@ class RepositoryOverview(private val repoManager: IRepositoriesManager) {
...
-                    headContent { title("Repositories") }
+                    headContent {
+                        title("Repositories")
+                        script(type = "text/javascript") {
+                            unsafe {
+                                +"""
+                                    function removeBranch(repository, branch) {
+                                        if (confirm('Are you sure you want to delete the branch' + branch +' of repository' +repository + '?')) {
+                                            fetch('../v2/repositories/' + repository + '/branches/' + branch, { method: 'DELETE'})
+                                            .then( _ => location.reload())
```
languitar, line 47: I hope we never have an error response from the DELETE endpoint. That would probably be invisible to the user now without opening the browser console.

model-server/src/main/kotlin/org/modelix/model/server/handlers/ui/RepositoryOverview.kt
```diff
@@ -98,24 +125,33 @@ class RepositoryOverview(private val repoManager: IRepositoriesManager) {
...
         +"Explore Latest Version"
     }
 }
 
-fun FlowContent.buildDeleteForm(repositoryId: String) {
+internal fun FlowContent.buildDeleteRepositoryForm(repositoryId: String) {
     form {
         postButton {
             name = "delete"
             formAction = "../v2/repositories/${repositoryId.encodeURLPathPart()}/delete"
```
languitar, line 144: I think we have the same problem as with the other confirmation modal here: after deleting, the user might see a blank page? Or not?

languitar (commented):
Sorry, one more iteration of things I didn't see in the first round.

model-server/src/main/kotlin/org/modelix/model/server/handlers/ui/RepositoryOverview.kt
mhuster23: The browser stays on the same page, but the user needs a manual reload to see the removed repository. Should I also add the auto-refresh here?

languitar: Not seeing the effect of an action is probably a UX anti-pattern. So yes, a force reload would be good.

languitar (approved):

PR #862, opened 2024-06-27
MODELIX-949: media type versioning + branch meta data access

languitar:
This PR adds a new media type for requesting branch metadata as the first test case to prepare for media-type versioning. To achieve this versioning the scheme, the PR modifies the openapi-generator templates so that one can optionally declare separate handler methods per media type(s) to generate in the API interface.

This PR also adds kotest-assertions and therefore closes MODELIX-795

* [ ] Code is readable

### DIFF
--- a/model-server-openapi/specifications/model-server-v2.yaml
+++ b/model-server-openapi/specifications/model-server-v2.yaml
@@ -140,15 +140,36 @@ paths:
           required: true
           schema:
             type: string
+      x-modelix-media-type-handlers:
+        - v1:
+            - 'application/x.modelix.branch+json;version=1'
+        - delta:
+            - 'application/x-modelix-objects-v2'
+            - 'application/x-modelix-objects'
+            - 'application/json'
+            - 'text/plain'
+            - '*/*'
       responses:
         "404":
           $ref: '#/components/responses/404'
         "200":
-          $ref: '#/components/responses/versionDelta'
-#          content:
-#            '*/*':
-#              schema:
-#                $ref: "#/components/schemas/VersionDelta"
+          description: "Information about a branch for content type `application/x.modelix.branch+json;version=*'. Else all model data of the branch in version delta format."
+          content:
+            'application/x.modelix.branch+json;version=1':
+              schema:
+                $ref: "#/components/schemas/BranchV1"
+            'application/x-modelix-objects-v2':
+              schema:
+                type: string
+            'application/x-modelix-objects':
+              schema:
+                type: string
+            'application/json':
+              schema:
+                type: object
+            'text/plain':
+              schema:
+                type: string
         default:
           $ref: '#/components/responses/GeneralError'
     post:
@@ -538,3 +559,15 @@ components:
           type: string
         value2:
           type: string
+    BranchV1:
+      x-modelix-media-type: 'application/x.modelix.branch+json;version=1'
+      type: object
+      properties:
+        name:
+          type: string
+        current_hash:
+          type: string
+          example: 7fQeo*xrdfZuHZtaKhbp0OosarV5tVR8N3pW8JPkl7ZE
+      required:
+        - name
+        - current_hash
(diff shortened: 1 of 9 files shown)


model-server/src/main/kotlin/org/modelix/model/server/handlers/ModelReplicationServer.kt
```diff
@@ -119,6 +122,18 @@ class ModelReplicationServer(
...
 
+    override suspend fun PipelineContext<Unit, ApplicationCall>.getRepositoryBranchV1(
+        repository: String,
+        branch: String,
+        lastKnown: String?,
+    ) {
+        runWithRepository(repository) {
+            val branchRef = repositoryId(repository).getBranchReference(branch)
+            val versionHash = repositoriesManager.getVersionHash(branchRef) ?: throw BranchNotFoundException(branchRef)
+            call.respond(BranchV1(branch, versionHash))
```
odzhychko, line 133: // TODO Olekz What is the content-type of the repsonse?

model-server/src/test/kotlin/org/modelix/model/server/ModelServerTestUtil.kt
```diff
@@ -38,7 +41,13 @@ suspend fun ApplicationTestBuilder.createModelClient(): ModelClientV2 {
 
 fun Application.installDefaultServerPlugins() {
     install(WebSockets)
-    install(ContentNegotiation) { json() }
+    install(ContentNegotiation) {
+        json()
+        register(
```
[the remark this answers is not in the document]
languitar: Yes, I can try to figure out if I can generate a method for that automatically. That's a bit annoying because every data type that we use needs one of these lines now.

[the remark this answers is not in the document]
languitar: Does the pushed approach look right to you?

model-server-openapi/specifications/model-server-v2.yaml
```diff
@@ -52,6 +52,8 @@ paths:
   /repositories:
     get:
       operationId: getRepositories
+      x-ignore-this:
```
odzhychko, line 55: Is `x-ignore-this` still needed?

model-server/src/test/kotlin/org/modelix/model/server/handlers/ModelReplicationServerTest.kt
```diff
@@ -365,4 +383,53 @@ class ModelReplicationServerTest {
...
+
+            val response = client.get {
+                url {
+                    appendPathSegments("v2", "repositories", repositoryId.id, "branches", "master")
+                }
+                accept(VersionDeltaStreamV2.CONTENT_TYPE)
+            }
+
+            response shouldHaveStatus HttpStatusCodes.STATUS_CODE_OK
+            response.shouldHaveContentType(VersionDeltaStreamV2.CONTENT_TYPE)
```
odzhychko, line 402: Really consistent, that `shouldHaveContentType` is not marked as `infix` like `shouldHaveStatus` 😆

odzhychko (changes requested):

model-server/src/test/kotlin/org/modelix/model/server/ModelServerTestUtil.kt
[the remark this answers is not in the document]
odzhychko: Looks good in general.

It migth be an issue, that `registerJsonTypes` is an extension function.
How can I call extension functions from different packages (e. g. `org.modelix.api.v2.Paths.registerJsonType` and `org.modelix.api.v1.Paths.registerJsonType`)?

model-server-openapi/specifications/model-server-v2.yaml
languitar: oops no :D

model-server/src/test/kotlin/org/modelix/model/server/handlers/ModelReplicationServerTest.kt
languitar: I was already thinking about opening an upstream bug report for that missing syntax

model-server-openapi/specifications/model-server-v2.yaml
languitar: should be gone now.

model-server/src/test/kotlin/org/modelix/model/server/ModelServerTestUtil.kt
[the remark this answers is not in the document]
languitar: You need import aliases or calls without imports then. I couldn't find an easy version to include the file name in the function name and even without being an extension function, we would then have multiple imports to the same name if we need to register also for v1 later on (or v3).

PR #873, opened 2024-07-02
MODELIX-721 Migrate cucumber tests for model-server to ktor tests

mhuster23:
On my local machine the migrated test cases now only take 2s instead of 38s.
The saved 36s are roughly a 40% reduction of the previous total time to test the model-server.

* [ ] Code is readable


model-server/src/main/kotlin/org/modelix/model/server/handlers/ui/IndexPage.kt
```diff
@@ -0,0 +1,83 @@
...
+import io.ktor.server.routing.routing
+import kotlinx.html.a
+import kotlinx.html.h1
+import kotlinx.html.li
+import kotlinx.html.style
+import kotlinx.html.ul
+import kotlinx.html.unsafe
+import org.modelix.model.server.templates.PageWithMenuBar
+
+class IndexPage {
```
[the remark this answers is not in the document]
languitar: Let's just add a few word here.

model-server/src/test/kotlin/org/modelix/model/server/V1ApiTest.kt
```diff
@@ -0,0 +1,326 @@
...
+        assertEquals(HttpStatusCode.OK, response2.status)
+        assertNotEquals(response1.bodyAsText(), response2.bodyAsText())
+    }
+
+    @Test
+    fun `events are received after subscription`() = runTest {
+        val key = "dylandog"
+        val value = "a comic book"
+
+        coroutineScope {
```
languitar, line 75: Why do these tests need an explicit scope?

model-server/src/test/kotlin/org/modelix/model/server/handlers/HealthApiTest.kt
```diff
@@ -0,0 +1,57 @@
...
+        }
+
+        block()
+    }
+
+    @Test
+    fun `health endpoint returns healthy`() = runTest {
+        val response = client.get("/health")
+        assertEquals("healthy", response.bodyAsText())
+    }
```
languitar, line 56: For the sake of completeness, another case could be added non to see if unhealthy is returned when something is off. Also, the return code should be validated here.

model-server/src/test/kotlin/org/modelix/model/server/V1ApiTest.kt
languitar, line 1: Can we use this PR as an opportunity to clean up the test case names for matching the style we use elsewhere. For instance "retrieving multiple keys - all existing" could become "All existing keys can be pulled/retrieved".

languitar (commented):
Do we know the reasons for why some of the cucumber scenarios were commented out? Is it safe to drop them?

Apart from that, only minor things that I added.

After #862, kotest assertions would now also be available as an option.

model-server/src/main/kotlin/org/modelix/model/server/handlers/ui/IndexPage.kt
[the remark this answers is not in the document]
mhuster23: What kind of information are you missing here? Explaining what an index page is in general feels overkill here.

model-server/src/test/kotlin/org/modelix/model/server/V1ApiTest.kt
mhuster23: To be able to call `async`. The get request will only be answered once the put has been performed.

model-server/src/main/kotlin/org/modelix/model/server/handlers/ui/IndexPage.kt
```diff
@@ -0,0 +1,83 @@
...
+                            li {
+                                a("user") { +"View JWT token and permissions" }
+                            }
+                            li {
+                                a("swagger") { +"SwaggerUI" }
+                            }
+                        }
+                    }
+                }
+                call.respondText("Model Server")
```
languitar, line 79: Btw, what is this line actually doing?

model-server/src/test/kotlin/org/modelix/model/server/V1ApiTest.kt
languitar: That is because of the execution semantics of `runTest`?

If the intended effect of the whole code here is to achieve:

1. GET starts polling
2. PUT happens
3. GET returns

Then I am currently not sure if that really happens that way.

First, with any general type of dispatcher, the block in the `async` might start executing only after the PUT has happened. So we have no real guarantee that this test is actually testing what it assumes to test.

Second, with the test dispatcher in particular, only the `await` will actually trigger the execution of the `async` block. So we will never end up in GET being started before the PUT. And `coroutineScope` does not magically switch to a different dispatcher as far as I know.

model-server/src/main/kotlin/org/modelix/model/server/handlers/ui/IndexPage.kt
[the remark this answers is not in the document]
languitar: Not much ;-) But we'd eventually get to a state with less warnings, which makes it easier to spot the important ones.

model-server/src/test/kotlin/org/modelix/model/server/V1ApiTest.kt
mhuster23: I have verified that it happens that way. Note that this runs in the default dispatcher since it's not the `runTest` of `kotlinx.coroutines`.
By default async is scheduled for immediate execution. However there might be a race condition, where the PUT might happen before the GET. Though I'm not sure on that, since i haven't dived deep enough into kotlin coroutines yet.

model-server/src/main/kotlin/org/modelix/model/server/handlers/ui/IndexPage.kt
mhuster23: removed it

model-server/src/test/kotlin/org/modelix/model/server/V1ApiTest.kt
mhuster23: done

model-server/src/main/kotlin/org/modelix/model/server/handlers/ui/IndexPage.kt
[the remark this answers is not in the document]
mhuster23: done

model-server/src/test/kotlin/org/modelix/model/server/V1ApiTest.kt
languitar: Can we rename `runTest` to something that not conflicts with the kotlin test `runTest` method? That's part of the confusion.

To enforce the required ordering, you can use a [Semaphore](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.sync/-semaphore/)

mhuster23: done

languitar (approved):
Ah, good idea with the plugin.

The only thing potentially left is that we have the exact same logic now copied to two tests, which could be deduplicated.

languitar (approved):

PR #874, opened 2024-07-02
fix: IConcept.isSubConceptOf/.isExactly if concepts are of different …

slisson:
…implementations

All implementations returned false if the parameter was of a different type. This was causing exceptions when the typed model API was used on an MPS model.

The implementations now handle this case by also comparing UIDs.

TODO description of the PR

* [ ] Code is readable

### DIFF
--- a/model-api-gen-runtime/src/commonMain/kotlin/org/modelix/metamodel/GeneratedConcept.kt
+++ b/model-api-gen-runtime/src/commonMain/kotlin/org/modelix/metamodel/GeneratedConcept.kt
@@ -149,7 +149,10 @@ abstract class GeneratedConcept<NodeT : ITypedNode, ConceptT : ITypedConcept>(
     }
 
     override fun isExactly(concept: IConcept?): Boolean {
-        return concept == this
+        if (concept == null) return false
+        if (concept == this) return true
+        if (concept.getUID() == this.getUID()) return true
+        return false
     }
 
     override fun isSubConceptOf(superConcept: IConcept?): Boolean {
--- a/mps-model-adapters/src/main/kotlin/org/modelix/model/mpsadapters/MPSConcept.kt
+++ b/mps-model-adapters/src/main/kotlin/org/modelix/model/mpsadapters/MPSConcept.kt
@@ -63,8 +63,17 @@ data class MPSConcept(val concept: SAbstractConceptAdapter) : IConcept {
     }
 
     override fun isSubConceptOf(superConcept: IConcept?): Boolean {
-        val mpsSuperConcept = superConcept as? MPSConcept ?: return false
-        return concept.isSubConceptOf(mpsSuperConcept.concept)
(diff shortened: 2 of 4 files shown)


model-api-gen-runtime/src/commonMain/kotlin/org/modelix/metamodel/GeneratedConcept.kt
```diff
@@ -149,7 +149,10 @@ abstract class GeneratedConcept<NodeT : ITypedNode, ConceptT : ITypedConcept>(
     }
 
     override fun isExactly(concept: IConcept?): Boolean {
-        return concept == this
+        if (concept == null) return false
+        if (concept == this) return true
+        if (concept.getReference().getUID() == getReference().getUID()) return true
```
benedekh, line 154: **Question:** Why do we need the IConceptReference instead of just the IConcept to check the UID?

mps-model-adapters/src/main/kotlin/org/modelix/model/mpsadapters/MPSConcept.kt
```diff
@@ -72,8 +80,12 @@ data class MPSConcept(val concept: SAbstractConceptAdapter) : IConcept {
     }
 
     override fun isExactly(concept: IConcept?): Boolean {
-        val otherMpsConcept = concept as? MPSConcept ?: return false
-        return this.concept == otherMpsConcept.concept
+        if (concept == null) return false
+        if (concept is MPSConcept) {
```
benedekh, line 84: **Question:** Before checking for equality, why do we have to check first if it's an `MPSConcept`?

benedekh (commented):
**Question:** Shall we deprecate (or directly remove) `IConcept.isSubConceptOf(IConceptReference)` that was implemented as a workaround for the problem this PR implements?

P.S: sorry, if the review comments came too early. I realized late, that it's still a draft.

model-api-gen-runtime/src/commonMain/kotlin/org/modelix/metamodel/GeneratedConcept.kt
slisson: Replaced it by getting the UID directly from the IConcept.

mps-model-adapters/src/main/kotlin/org/modelix/model/mpsadapters/MPSConcept.kt
slisson: Removed the unnecessary instance of check.

slisson:
> **Question:** Shall we deprecate (or directly remove) `IConcept.isSubConceptOf(IConceptReference)` that was implemented as a workaround for the problem this PR implements?

It's still useful to have that method to avoid unnecessary resolving of a ConceptReference.

odzhychko (approved):

PR #895, opened 2024-07-11
MODELIX-971 allProperties missing in ModelQl on INode

mhuster23:

* [ ] Code is readable

### DIFF
--- a/modelql-untyped/src/commonMain/kotlin/org/modelix/modelql/untyped/AllPropertiesTraversalStep.kt
+++ b/modelql-untyped/src/commonMain/kotlin/org/modelix/modelql/untyped/AllPropertiesTraversalStep.kt
@@ -1,0 +1,70 @@
+/*
+ * Copyright (c) 2024.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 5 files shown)


modelql-untyped/src/commonMain/kotlin/org/modelix/modelql/untyped/AllPropertiesTraversalStep.kt
```diff
@@ -0,0 +1,70 @@
...
+    @OptIn(ExperimentalCoroutinesApi::class)
+    override fun createFlow(
+        input: StepFlow<INode>,
+        context: IFlowInstantiationContext,
+    ): StepFlow<Pair<IProperty, String?>> {
+        return input.flatMapConcat { it.value.getAllPropertiesAsFlow() }.asStepFlow(this)
+    }
+
+    override fun getOutputSerializer(serializationContext: SerializationContext): KSerializer<out IStepOutput<Pair<IProperty, String?>>> {
+        return serializationContext.serializer<Pair<IProperty, String?>>().stepOutputSerializer(this)
```
slisson, line 51: This only works, because the test case uses a `CLTree` with named based roles. Not all implementations of `IProperty` are serializable. I think we need https://issues.modelix.org/issue/MODELIX-976/Cleanup-relation-between-model-and-meta-model first and then return an `IPropertyReference` (or `IPropertyId`, not sure yet) instead, that can be resolved on the client side to an IPropertyDefinition if needed.

mhuster23: I tested it with a CLTree that uses `useRoleIds = true` and it works for that case as well.
But your proposed solution would definitely be cleaner.

slisson: Because the role is always wrapped by a `PropertyFromName`, even if it is an ID. This really needs to be cleaned up.

mhuster23:
This proposed fix is outdated with the modelql rewrite in core 10.0.0

PR #906, opened 2024-07-18
MODELIX-778 Chaining write operations in ModelQL leads to repeated actions

mhuster23:
**TODO:**
~~* [ ] in the non-linear test case the order of setting the name properties seems to be non-deterministic~~ --> [MODELIX-995
](https://issues.modelix.org/issue/MODELIX-995)

* [ ] Code is readable

### DIFF
--- a/model-api-gen-gradle-test/kotlin-generation/src/test/kotlin/org/modelix/modelql/typed/TypedModelQLTest.kt
+++ b/model-api-gen-gradle-test/kotlin-generation/src/test/kotlin/org/modelix/modelql/typed/TypedModelQLTest.kt
@@ -371,4 +371,37 @@ abstract class TypedModelQLTest {
         }
         assertEquals("", name)
     }
+
+    @Test
+    fun `chained write operations are executed once`() = runTypedModelQLTest { client ->
+        val oldNumChildren = client.query { root ->
+            root.children("classes").ofConcept(C_ClassConcept)
+                .first()
+                .member
+                .ofConcept(C_StaticMethodDeclaration)
+                .count()
+        }
+
+        val newChildName = "I am unique"
+
+        client.query { root ->
+            root.children("classes").ofConcept(C_ClassConcept)
+                .first()
+                .addToMember(C_StaticMethodDeclaration)
+                .setName(newChildName.asMono())
+        }
+
+        val actualChildren = client.query { root ->
+            root.children("classes").ofConcept(C_ClassConcept)
+                .first()
+                .member
+                .ofConcept(C_StaticMethodDeclaration)
+                .toList()
+        }
+
+        val numAddExecutions = actualChildren.size - oldNumChildren
+        assertEquals(1, numAddExecutions)
+        assertEquals(C_StaticMethodDeclaration.untyped().getUID(), actualChildren.last().untyped().concept?.getUID())
+        assertEquals(newChildName, actualChildren.last().name)
+    }
 }
--- a/model-api-gen-gradle-test/kotlin-generation/src/test/kotlin/org/modelix/modelql/typed/TypedModelQLTest.kt
+++ b/model-api-gen-gradle-test/kotlin-generation/src/test/kotlin/org/modelix/modelql/typed/TypedModelQLTest.kt
@@ -75,7 +75,7 @@ import kotlin.test.assertTrue
 
 abstract class TypedModelQLTest {
 
-    abstract fun runTest(block: suspend (ModelQLClient) -> Unit)
+    abstract fun runTypedModelQLTest(block: suspend (ModelQLClient) -> Unit)
 
     protected fun createTestData(rootNode: INode) {
         rootNode.allChildren.forEach { it.remove() }
@@ -116,12 +116,12 @@ abstract class TypedModelQLTest {
         // Example for single non-abstract child
         rootNode.addNewChild("xmlFile", -1, C_XmlFile.untyped())
 
-        // Example for mulitple non-abstract child
+        // Example for multiple non-abstract child
         rootNode.addNewChild("xmlComment", -1, C_XmlComment.untyped())
     }
 
     @Test
-    fun `simple query`() = runTest { client ->
+    fun `simple query`() = runTypedModelQLTest { client ->
         val result: Int = client.query { root ->
             root.children("classes").ofConcept(C_ClassConcept)
                 .member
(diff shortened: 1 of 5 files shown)


modelql-core/src/commonMain/kotlin/org/modelix/modelql/core/IStep.kt
```diff
@@ -212,7 +225,7 @@ abstract class AggregationStep<In, Out> : MonoTransformingStep<In, Out>() {
         val flow = flow {
             emit(aggregate(input))
         }
-        return if (outputIsConsumedMultipleTimes()) {
+        return if (needsCoroutineScope()) {
             val scope = context.coroutineScope ?: throw RuntimeException("Coroutine scope required for caching of $this")
             flow.shareIn(scope, SharingStarted.Lazily, 1)
```
odzhychko, line 230: Is the check `flow.assertNotEmpty() // empty flow would result in deadlock` not needed here?

modelql-core/src/commonMain/kotlin/org/modelix/modelql/core/MappingStep.kt
```diff
@@ -36,6 +36,10 @@ class MappingStep<In, Out>(val query: MonoUnboundQuery<In, Out>) : MonoTransform
         return query.requiresWriteAccess()
     }
 
+    override fun needsCoroutineScope(): Boolean {
```
odzhychko, line 39: Why is this override needed?
What if the `MappingStep.requiresWriteAccess` is `true`?

(It can also be simplified to `override fun needsCoroutineScope() = false`.)

modelql-core/src/commonMain/kotlin/org/modelix/modelql/core/IStep.kt
mhuster23: I don't think the flow can be empty here, as line 226 always emits an `IStepOutput`.

model-api-gen-gradle-test/kotlin-generation/src/test/kotlin/org/modelix/modelql/typed/TypedModelQLTest.kt
```diff
@@ -371,4 +371,37 @@ abstract class TypedModelQLTest {
...
         assertEquals("", name)
     }
+
+    @Test
+    fun `chained write operations are executed once`() = runTypedModelQLTest { client ->
+        val oldNumChildren = client.query { root ->
+            root.children("classes").ofConcept(C_ClassConcept)
+                .first()
+                .member
+                .ofConcept(C_StaticMethodDeclaration)
```
languitar, line 381: This repeated block could become a utility function

model-api-gen-gradle-test/kotlin-generation/src/test/kotlin/org/modelix/modelql/typed/TypedModelQLTest.kt
```diff
@@ -371,4 +371,37 @@ abstract class TypedModelQLTest {
         }
         assertEquals("", name)
     }
+
+    @Test
+    fun `chained write operations are executed once`() = runTypedModelQLTest { client ->
```
languitar, line 376: Nitpick: this test actually asserts more: it also validates that the correct write operations take place and result in the desired state.

modelql-core/src/commonMain/kotlin/org/modelix/modelql/core/IStep.kt
languitar: Shouldn't we assert these "i think" things? ;-)

modelql-core/src/commonMain/kotlin/org/modelix/modelql/core/IStep.kt
```diff
@@ -34,7 +34,7 @@ interface IStep {
 
     fun requiresWriteAccess(): Boolean = false
     fun hasSideEffect(): Boolean = requiresWriteAccess()
-    fun needsCoroutineScope() = false
+    fun needsCoroutineScope() = hasSideEffect()
```
languitar, line 37: It would probably be good to explain somewhere, why needing a couroutine scope depends on the side effect.

languitar (commented):
For the sake of completeness: wouldn't it be code to also test reference assignments? The tests right now only test child assignments and properties.

modelql-core/src/commonMain/kotlin/org/modelix/modelql/core/IStep.kt
odzhychko: > I don't think the flow can be empty here, as line 226 always emits an IStepOutput.

You right just two lines above we can see, that the flow has exactly one element.

modelql-core/src/commonMain/kotlin/org/modelix/modelql/core/MappingStep.kt
odzhychko: Hm, ok. I still do not fully understand it.

Anyway, now we have many other `MonoTransformingStep`s like `MapIfNotNullStep` that break with the current change because they are not adapted like `MappingStep` here.

odzhychko: Also `FluxTransformingStep`s or `TransformingStep`s might be affected. For now I only tested, that `MonoTransformingSteps` is broken.

languitar: Why don't we have tests that uncover those problems?

odzhychko: Also using `.map` in the provided test fails:

```
@Test
    fun `chained write operations are executed once`() = runTypedModelQLTest { client ->
       ...

        client.query { root ->
            root.children("classes").ofConcept(C_ClassConcept)
                .first()
                .map {
                    it.addToMember(C_StaticMethodDeclaration)
                }
                .setName(newChildName.asMono())
        }

        ...
        assertEquals(newChildName, actualChildren.last().name)
    }
  ```

odzhychko: > Why don't we have tests that uncover those problems?

I guess this is a fragile base class problem.

`ModelQLClientTest.testCaching()` tested correct implementation in `MonoTransformingSteps/TransformingStep` using  `MappingStep` as an example. But with changes to the base class things broke 🤷

odzhychko: @mhuster23 Do you think it makes still sense to fix it with the Flow-Implementation?
The new Bulk-Sync in [MODELIX-914](https://issues.modelix.org/issue/MODELIX-914) will possibly get rid of this bug anyway.

mhuster23:
The issue still persists in modelix 10.0
The proposed fix here is outdated.

PR #907, opened 2024-07-18
fix(model-api-gen): fix generated code not compiling if conceptProper…

mhuster23:
…tiesInterfaceName was not set

This was a regression introduced in commit 86c141bd.

* [x] Code is readable

### DIFF
--- a/model-api-gen/src/main/kotlin/org/modelix/metamodel/generator/internal/ConceptObjectGenerator.kt
+++ b/model-api-gen/src/main/kotlin/org/modelix/metamodel/generator/internal/ConceptObjectGenerator.kt
@@ -53,6 +53,7 @@ import kotlin.reflect.KClass
 internal class ConceptObjectGenerator(
     private val concept: ProcessedConcept,
     override val nameConfig: NameConfig,
+    private val conceptPropertiesInterfaceName: String?,
     private val alwaysUseNonNullableProperties: Boolean = true,
 ) : NameConfigBasedGenerator(nameConfig) {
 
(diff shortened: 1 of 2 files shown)


model-api-gen/src/main/kotlin/org/modelix/metamodel/generator/internal/ConceptObjectGenerator.kt
```diff
@@ -101,7 +102,7 @@ internal class ConceptObjectGenerator(
     }
 
     private fun TypeSpec.Builder.addConceptPropertiesGetter() {
-        if (concept.metaProperties.isEmpty()) return
+        if (conceptPropertiesInterfaceName == null || concept.metaProperties.isEmpty()) return
```
odzhychko, line 105: > This was a regression introduced in commit https://github.com/modelix/modelix.core/commit/86c141bdb73b5623c7954ead0de03d1156ed4d6c.

 nit: a test would be nice

odzhychko (approved):

model-api-gen/src/main/kotlin/org/modelix/metamodel/generator/internal/ConceptObjectGenerator.kt
mhuster23: I agree but it isn't really possible with our current testing setup, since it only allows us to test one single configuration of the plugin.
It's something we have to keep in mind for [MODELIX-798](https://issues.modelix.org/issue/MODELIX-798/Create-a-proposal-for-reorganizing-the-existing-tests)

odzhychko: Ok, I added it as an issue in the ticket.

PR #950, opened 2024-08-02
chore(model-server): remove implementation of `LightModelServer` from model-server code

odzhychko:
The implementation `org.modelix.model.server.handlers.LightModelServer` was only used as a fake to test the light-model-client.
Renaming and moving it into the test code of the light-model-client helps to understand that.

The light-model-server implementation running in MPS is `org.modelix.model.server.light.LightModelServer`.

The next step would be to remove `FakeLightModelServer` and use `org.modelix.model.server.light.LightModelServer` to test the light-model-client.
See MODELIX-994

* [ ] Code is readable

### DIFF
--- a/light-model-client/build.gradle.kts
+++ b/light-model-client/build.gradle.kts
@@ -40,12 +40,6 @@ kotlin {
                 implementation(libs.kotlin.logging)
                 implementation(libs.kotlin.coroutines.core)
                 implementation(libs.kotlin.serialization.json)
-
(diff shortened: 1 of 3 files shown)


odzhychko:
> I don't understand this PR. The title says that the model-server code is affected, but I only see changes in light-model-client. Also, some explanations in the commit message and/or the PR would be good.

I updated the commit message. It should be more clear now.

PR #966, opened 2024-08-12
fix(model-server): do not use blocking network calls

odzhychko:
Additionally, tested error response in touched code.

Fixes the issue described in https://issues.modelix.org/issue/MODELIX-998

* [ ] Code is readable

### DIFF
--- a/model-server/src/main/kotlin/org/modelix/model/server/handlers/HttpException.kt
+++ b/model-server/src/main/kotlin/org/modelix/model/server/handlers/HttpException.kt
@@ -66,22 +66,6 @@ open class HttpException(problem: Problem, cause: Throwable? = null) : RuntimeEx
 }
 
 /**
- * Indicates a bad request from the client with missing or invalid information provided.
- *
- * @param details the detailed message to expose to the caller
- * @param typeSuffix A detailed type making this bad request instance uniquely identifiable. [PROBLEM_NAMESPACE] will
- *                   automatically be prepended. This field should be written in kebab-case.
- * @param cause The causing exception for the bad request or null if none.
- */
-class BadRequestException(details: String, typeSuffix: String, cause: Throwable? = null) : HttpException(
-    HttpStatusCode.BadRequest,
-    title = "Bad request",
-    details = details,
-    type = problemType(typeSuffix),
-    cause = cause,
-)
-
-/**
  * A [HttpException] indicating that a branch was not found in a repository.
  *
  * @param branch name of the missing branch
@@ -105,7 +89,7 @@ class BranchNotFoundException(branch: String, repositoryId: String, cause: Throw
 /**
  * A [HttpException] indicating that a version inside a repository or branch was not found.
  *
- * @param versionHash has of the missing version
+ * @param versionHash hash of the missing version
  * @param cause The causing exception for the bad request or null if none.
  */
 class VersionNotFoundException(versionHash: String, cause: Throwable? = null) :
@@ -114,5 +98,81 @@ class VersionNotFoundException(versionHash: String, cause: Throwable? = null) :
         title = "Version not found",
         details = "Version '$versionHash' doesn't exist",
         type = "/problems/version-not-found",
+        cause = cause,
+    )
+
+/**
+ * An [HttpException] indicating that a provided repository name is not valid.
+ *
+ * @param invalidRepositoryId the invalid repository ID
+ * @param cause The causing exception for the bad request or null if none.
+ */
+class InvalidRepositoryIdException(invalidRepositoryId: String, cause: Throwable? = null) :
+    HttpException(
+        HttpStatusCode.BadRequest,
+        title = "Invalid repository ID.",
+        details = "Repository ID `$invalidRepositoryId` is not valid.",
+        type = "/problems/invalid-repository-id",
+        cause = cause,
+    )
+
+/**
+ * An [HttpException] indicating that an object key was uploaded without the corresponding value.
+ *
+ * @param objectKey the uploaded object key
+ */
+class ObjectKeyWithoutObjectValueException(objectKey: String) :
+    HttpException(
+        HttpStatusCode.BadRequest,
+        title = "Uploaded object key without object value.",
+        details = "Uploaded object key `$objectKey` without object value.",
+        type = "/problems/object-key-without-object-value",
+    )
+
+/**
+ * An [HttpException] indicating that an invalid object key was uploaded.
+ *
+ * @param objectKey the uploaded object key
+ */
+class InvalidObjectKeyException(objectKey: String) :
+    HttpException(
+        HttpStatusCode.BadRequest,
+        title = "Uploaded invalid object key.",
+        details = "Uploaded invalid object key `$objectKey`.",
+        type = "/problems/invalid-object-key",
+    )
+
+/**
+ * An [HttpException] indicating that an object key and value were uploaded that do not match.
+ *
+ * @param uploadedObjectKey the uploaded object key
+ * @param expectedObjectKey the expected object key based on the object value
+ * @param objectValue the uploaded object value
+ */
+class MismatchingObjectKeyAndValueException(
+    uploadedObjectKey: String,
+    expectedObjectKey: String,
+    objectValue: String,
+) :
+    HttpException(
+        HttpStatusCode.BadRequest,
+        title = "Uploaded mismatching object key and value.",
+        details = "Uploaded object key `$uploadedObjectKey` does not match expected object key `$expectedObjectKey` " +
+            "for object value `$objectValue`.",
+        type = "/problems/mismatching-object-and-value",
+    )
+
+/**
+ * A [HttpException] indicating that an object value was not found.
+ *
+ * @param objectHash hash of the missing object
+ * @param cause The causing exception for the bad request or null if none.
+ */
+class ObjectValueNotFoundException(objectHash: String, cause: Throwable? = null) :
+    HttpException(
+        HttpStatusCode.NotFound,
+        title = "Object value not found.",
+        details = "Object value with hash `$objectHash` does not exist.",
+        type = "/problems/object-value-not-found",
         cause = cause,
     )
--- a/model-server/src/main/kotlin/org/modelix/model/server/handlers/ModelReplicationServer.kt
+++ b/model-server/src/main/kotlin/org/modelix/model/server/handlers/ModelReplicationServer.kt
@@ -21,7 +21,7 @@ import io.ktor.server.application.ApplicationCall
 import io.ktor.server.application.call
 import io.ktor.server.request.acceptItems
 import io.ktor.server.request.receive
-import io.ktor.server.request.receiveStream
+import io.ktor.server.request.receiveChannel
 import io.ktor.server.response.respond
 import io.ktor.server.response.respondBytesWriter
 import io.ktor.server.response.respondText
@@ -31,6 +31,7 @@ import io.ktor.util.cio.use
 import io.ktor.util.pipeline.PipelineContext
 import io.ktor.utils.io.ByteWriteChannel
 import io.ktor.utils.io.close
+import io.ktor.utils.io.readUTF8Line
 import io.ktor.utils.io.writeStringUtf8
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.flow.Flow
@@ -239,16 +240,28 @@ class ModelReplicationServer(
 
     override suspend fun PipelineContext<Unit, ApplicationCall>.postRepositoryObjectsGetAll(repository: String) {
         checkPermission(ModelServerPermissionSchema.repository(repository).objects.read)
-        runWithRepository(repository) {
-            val keys = call.receiveStream().bufferedReader().use { reader ->
-                reader.lineSequence().toHashSet()
+        val channel = call.receiveChannel()
+        val keys = hashSetOf<String>()
+        while (true) {
+            val line = channel.readUTF8Line() ?: break
+            keys.add(line)
+        }
+
+        val objects = runWithRepository(repository) {
+            withContext(Dispatchers.IO) {
+                modelClient.store.getAll(keys)
             }
-            val objects = withContext(Dispatchers.IO) { modelClient.store.getAll(keys) }.checkValuesNotNull {
-                "Object not found: $it"
+        }
+
+        for (entry in objects) {
+            if (entry.value == null) {
+                throw IllegalStateException("Object not found: ${entry.value}")
             }
-            call.respondTextWriter(contentType = ImmutableObjectsStream.CONTENT_TYPE) {
-                ImmutableObjectsStream.encode(this, objects)
-            }
+        }
+        @Suppress("UNCHECKED_CAST")
+        objects as Map<String, String>
+        call.respondTextWriter(contentType = ImmutableObjectsStream.CONTENT_TYPE) {
+            ImmutableObjectsStream.encode(this, objects)
         }
     }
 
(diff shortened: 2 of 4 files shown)


model-server/src/main/kotlin/org/modelix/model/server/handlers/ModelReplicationServer.kt
```diff
@@ -240,8 +240,10 @@ class ModelReplicationServer(
...
         checkPermission(ModelServerPermissionSchema.repository(repository).objects.read)
         runWithRepository(repository) {
-            val keys = call.receiveStream().bufferedReader().use { reader ->
-                reader.lineSequence().toHashSet()
+            val keys = withContext(Dispatchers.IO) {
+                call.receiveStream().bufferedReader().use { reader ->
+                    reader.lineSequence().toHashSet()
+                }
             }
             val objects = withContext(Dispatchers.IO) { modelClient.store.getAll(keys) }.checkValuesNotNull {
```
languitar, line 248: Any reason not to combine the two `withContext` blocks into one?

odzhychko: Done. For now they can go into one.

(Later model client should not be called directly from here.)

model-server/src/main/kotlin/org/modelix/model/server/handlers/ModelReplicationServer.kt
```diff
@@ -240,11 +240,11 @@ class ModelReplicationServer(
     override suspend fun PipelineContext<Unit, ApplicationCall>.postRepositoryObjectsGetAll(repository: String) {
         checkPermission(ModelServerPermissionSchema.repository(repository).objects.read)
         runWithRepository(repository) {
-            val keys = call.receiveStream().bufferedReader().use { reader ->
-                reader.lineSequence().toHashSet()
-            }
-            val objects = withContext(Dispatchers.IO) { modelClient.store.getAll(keys) }.checkValuesNotNull {
-                "Object not found: $it"
+            val objects = withContext(Dispatchers.IO) {
+                val keys = call.receiveStream().bufferedReader().use { reader ->
```
languitar, line 244: Looking back at the ktor documentation: why don't we use `receiveChannel` as a non-blocking version?

odzhychko: I found no reason why it was not used.
I have the other occurrence for consistency too.

model-server/src/main/kotlin/org/modelix/model/server/handlers/ModelReplicationServer.kt
```diff
@@ -153,6 +154,7 @@ class ModelReplicationServer(
         val repositoryId = try {
             RepositoryId(repository)
         } catch (e: IllegalArgumentException) {
+            // TODO Olekz How does the "invalid-request" should look?
```
odzhychko, line 157:
```
 * @param details the detailed message to expose to the caller
 * @param typeSuffix A detailed type making this bad request instance uniquely identifiable. [PROBLEM_NAMESPACE] will
 *                   automatically be prepended. This field should be written in kebab-case.
 * @param cause The causing exception for the bad request or null if none.
 */
class BadRequestException(details: String, typeSuffix: String, cause: Throwable? = null) : HttpException(
    HttpStatusCode.BadRequest,
    title = "Bad request",
    details = details,
    type = problemType(typeSuffix),
    cause = cause,
)
```

@languitar  What should be the `typeSuffix` when creating a `BadRequestException`.
Should each occurrence have a different `typeSuffix`?

languitar: `typeSuffix` is meant to be a unique identifier for the type of problem resulting in this error condition. For instance, something like "out-of-credit". Cf. https://www.rfc-editor.org/rfc/rfc9457.html#name-type

odzhychko: Ok, I created subclasses of `HttpException` instead of using `BadRequestException`.


(Creating all these subclasses instead of having functions that create instances of `HttpException` is so Java-like 😆)

languitar: We can still decide which way we want to go. I guess consistent with other frameworks would be to at least have classes mapping to HTTP response codes.

model-server/src/main/kotlin/org/modelix/model/server/handlers/HttpException.kt
```diff
@@ -116,3 +116,79 @@ class VersionNotFoundException(versionHash: String, cause: Throwable? = null) :
...
+class UploadedObjectKeyWithoutObjectValueException(objectKey: String) :
+    HttpException(
+        HttpStatusCode.BadRequest,
+        title = "Uploaded object key without object value.",
+        details = "Uploaded object key `$objectKey` without object value.",
+        type = "/problems/uploaded-object-key-without-object-value",
+    )
+
+/**
+ * An [HttpException]  indicating that an invalid object key was uploaded.
```
languitar, line 133:
```suggestion
 * An [HttpException] indicating that an invalid object key was uploaded.
```

model-server/src/main/kotlin/org/modelix/model/server/handlers/HttpException.kt
```diff
@@ -116,3 +116,79 @@ class VersionNotFoundException(versionHash: String, cause: Throwable? = null) :
...
+ *
+ * @param objectKey the uploaded object key
+ */
+class UploadedInvalidObjectKeyException(objectKey: String) :
```
languitar, line 137:
```suggestion
class fInvalidObjectKeyException(objectKey: String) :
```

Wouldn't that be enough?

model-server/src/main/kotlin/org/modelix/model/server/handlers/HttpException.kt
```diff
@@ -116,3 +116,79 @@ class VersionNotFoundException(versionHash: String, cause: Throwable? = null) :
...
+        type = "/problems/invalid-repository-id",
+        cause = cause,
+    )
+
+/**
+ * An [HttpException] indicating that an object key was uploaded without the corresponding value.
+ *
+ * @param objectKey the uploaded object key
+ */
+class UploadedObjectKeyWithoutObjectValueException(objectKey: String) :
```
languitar, line 124:
```suggestion
class ObjectKeyWithoutObjectValueException(objectKey: String) :
```

Wouldn't that be enough?

model-server/src/main/kotlin/org/modelix/model/server/handlers/HttpException.kt
```diff
@@ -116,3 +116,79 @@ class VersionNotFoundException(versionHash: String, cause: Throwable? = null) :
...
+    )
+
+/**
+ * An [HttpException]  indicating that an object key and value were uploaded that do not match.
+ *
+ * @param uploadedObjectKey the uploaded object key
+ * @param expectedObjectKey the expected object key based on the object value
+ * @param objectValue the uploaded object value
+ */
+class UploadedMismatchingObjectKeyAndValueException(
```
languitar, line 152:
```suggestion
class MismatchingObjectKeyAndValueException(
```

model-server/src/main/kotlin/org/modelix/model/server/handlers/HttpException.kt
```diff
@@ -116,3 +116,79 @@ class VersionNotFoundException(versionHash: String, cause: Throwable? = null) :
...
+class UploadedInvalidObjectKeyException(objectKey: String) :
+    HttpException(
+        HttpStatusCode.BadRequest,
+        title = "Uploaded invalid object key.",
+        details = "Uploaded invalid object key `$objectKey`.",
+        type = "/problems/uploaded-invalid-object-key",
+    )
+
+/**
+ * An [HttpException]  indicating that an object key and value were uploaded that do not match.
```
languitar, line 146:
```suggestion
 * An [HttpException] indicating that an object key and value were uploaded that do not match.
```

model-server/src/main/kotlin/org/modelix/model/server/handlers/ModelReplicationServer.kt
```diff
@@ -239,17 +240,29 @@ class ModelReplicationServer(
...
-        runWithRepository(repository) {
-            val keys = call.receiveStream().bufferedReader().use { reader ->
-                reader.lineSequence().toHashSet()
-            }
-            val objects = withContext(Dispatchers.IO) { modelClient.store.getAll(keys) }.checkValuesNotNull {
-                "Object not found: $it"
+        val channel = call.receiveChannel()
+        val keys = hashSetOf<String>()
+        while (true) {
+            val line = channel.readUTF8Line() ?: break
```
languitar, line 246: This pattern seems to repeat below. What about factoring out something like `toLineStream`?

model-server/src/main/kotlin/org/modelix/model/server/handlers/ModelReplicationServer.kt
odzhychko: I'm fine classes.

model-server/src/main/kotlin/org/modelix/model/server/handlers/ModelReplicationServer.kt
```diff
@@ -239,17 +240,29 @@ class ModelReplicationServer(
...
-                ImmutableObjectsStream.encode(this, objects)
+        }
+
+        for (entry in objects) {
+            if (entry.value == null) {
+                throw ObjectValueNotFoundException(entry.key)
             }
         }
+        @Suppress("UNCHECKED_CAST")
+        objects as Map<String, String>
```
languitar, line 262: Why isn't this type already declared when the variable is created in line 250?

model-server/src/test/kotlin/org/modelix/model/server/handlers/ModelReplicationServerTest.kt
```diff
@@ -440,4 +450,110 @@ class ModelReplicationServerTest {
...
+            type = "/problems/uploaded-object-key-without-object-value",
+            title = "Uploaded object key without object value.",
+            status = 400,
+            detail = "Uploaded object key `7-m7S*flQz_CNc9F3ipSQ5Iz7SZJSyn2r0_1PPVvh8yA` without object value.",
+        )
+        assertEquals(expectedProblem, response.body<Problem>())
+    }
+
+    @Test
+    fun `putRepositoryObjects invalid key`() = runWithTestModelServer { _, _ ->
```
languitar, line 481:
```suggestion
    fun `putRepositoryObjects reports invalid key`() = runWithTestModelServer { _, _ ->
```

model-server/src/test/kotlin/org/modelix/model/server/handlers/ModelReplicationServerTest.kt
```diff
@@ -440,4 +450,110 @@ class ModelReplicationServerTest {
...
+            contentType(ContentType.Text.Plain)
+            setBody(invalidRequestBody)
+        }
+
+        val expectedProblem = Problem(
+            type = "/problems/uploaded-object-key-without-object-value",
+            title = "Uploaded object key without object value.",
+            status = 400,
+            detail = "Uploaded object key `7-m7S*flQz_CNc9F3ipSQ5Iz7SZJSyn2r0_1PPVvh8yA` without object value.",
+        )
```
languitar, line 476: I'm not sure if it is a good idea to test all aspects of the generated response. For instance, the detail is meant as a human-readable explanation. If we later decide that the details need a slight rewording to be better understandable, we'd have a million of tests to update. So I don't see much benefit in comparing this string but some drawbacks.

languitar: btw, a we now have kotest, you could use `shouldBeEqualIgnoringFields` to achieve such a comparison.

model-server/src/test/kotlin/org/modelix/model/server/handlers/ModelReplicationServerTest.kt
```diff
@@ -182,13 +184,21 @@ class ModelReplicationServerTest {
...
                     appendPathSegments("v2", "repositories", "invalid with spaces", "branches", "master")
                 }
             }
-            assertEquals(HttpStatusCode.BadRequest, response.status)
-            assertContains(response.bodyAsText(), "Invalid repository name")
+
+            val expectedProblem = Problem(
+                type = "/problems/invalid-repository-id",
+                title = "Invalid repository ID.",
+                status = 400,
```
languitar, line 198: Not sure if this is better?

```suggestion
                status = HttpStatusCode.BadRequest.value,
```

languitar (commented):
Some nitpicks.

Apart from that it looks like we have one or two new validations on the server side. Shouldn't these be part of a separate commit?

model-server/src/main/kotlin/org/modelix/model/server/handlers/ModelReplicationServer.kt
odzhychko: In line 250 the type is `Map<String, String?>`. Only in line 257 (`entry.value == null`) we check that the values are not null.

languitar: Ah, right. That missing bit of smart contracts in Kotlin.

model-server/src/test/kotlin/org/modelix/model/server/handlers/ModelReplicationServerTest.kt
odzhychko: I think user messages should also be tested, because they can break.
And rewording will not happen often.

model-server/src/main/kotlin/org/modelix/model/server/handlers/ModelReplicationServer.kt
languitar: Can't we go to a `Sequence` and then reuse the old implementation with `windowed`, which is overloaded for Sequence to return a Sequence.

model-server/src/test/kotlin/org/modelix/model/server/handlers/ModelReplicationServerTest.kt
languitar: We could move all of this to a single test for the exception class, asserting about the generated detail message and ignore the rest here. That way, when the next test asserts about the 400 response here, we at least don't have two spots checking the exact same detail message generation pattern.

model-server/src/main/kotlin/org/modelix/model/server/handlers/ModelReplicationServer.kt
odzhychko: Not, sure if this is possible.
Sequences and Coroutines use suspending functions but in a different way.
grafik


> The sequence builder should not use suspending operations other than yielding operations[3](https://kt.academy/article/cc-sequence#definition-3). If you need, for instance, to fetch data, it's better to use Flow, as will be explained later in the book. The way its builder works is similar to the sequence builder, but Flow has support for other coroutine features.

See https://kt.academy/article/cc-sequence

languitar: What a mess

languitar:
Would look fine to me now. Any reason this is still in draft state?

odzhychko:
> Would look fine to me now. Any reason this is still in draft state?

You are right, I forgot to mark it as ready for review.

languitar (approved):

PR #982, opened 2024-08-14
fix(bulk-model-sync-gradle): fix MPS Runner configuration for server …

odzhychko:
…sources with base revision

When a base revision was specified, the specified branch was not passed as a system property to the bulk sync.

When a base revision was specified without a source revision, the property value for the source revision was the string `"null"`.

To write test with TestKit we would need further adjustments to the build setup. For this fix, I considered them out of scope.

I tried to test execution with Gradles TestKit first. See https://docs.gradle.org/current/userguide/test_kit.html But the Plugin tries to download the package "org.modelix:bulk-model-sync-mps" as `mpsDependencies`. The names of the downloaded files are then used as `classPathElements` in the MPS Runner configuration. Therefore, the plugin cannot be tested with TestKit without publishing "org.modelix:bulk-model-sync-mps" and its dependents locally or including the "modelix.core" project from within a test execution.

* [ ] Code is readable

### DIFF
--- a/bulk-model-sync-gradle/src/test/kotlin/org/modelix/model/sync/bulk/gradle/MpsRunnerConfigurationTest.kt
+++ b/bulk-model-sync-gradle/src/test/kotlin/org/modelix/model/sync/bulk/gradle/MpsRunnerConfigurationTest.kt
@@ -59,7 +59,7 @@ class MpsRunnerConfigurationTest {
             "-Dmodelix.mps.model.sync.bulk.input.continueOnError=false",
             "-Xmx2g",
             "-Dmodelix.mps.model.sync.bulk.server.repository=aRepositoryId",
-            "-Dmodelix.mps.model.sync.bulk.server.version.hash=null",
+            "-Dmodelix.mps.model.sync.bulk.server.branch=aBranchName",
             "-Dmodelix.mps.model.sync.bulk.server.url=aUrl",
             "-Dmodelix.mps.model.sync.bulk.server.version.base.hash=aBaseRevision",
         )
--- a/bulk-model-sync-gradle/src/test/kotlin/org/modelix/model/sync/bulk/gradle/MpsRunnerConfigurationTest.kt
+++ b/bulk-model-sync-gradle/src/test/kotlin/org/modelix/model/sync/bulk/gradle/MpsRunnerConfigurationTest.kt
@@ -1,0 +1,99 @@
+package org.modelix.model.sync.bulk.gradle
+
+import io.kotest.matchers.collections.shouldContainExactlyInAnyOrder
+import org.modelix.model.sync.bulk.gradle.config.LocalTarget
+import org.modelix.model.sync.bulk.gradle.config.ServerSource
+import org.modelix.model.sync.bulk.gradle.config.SyncDirection
+import java.io.File
+import kotlin.test.Test
+
+class MpsRunnerConfigurationTest {
+
+    private val jsonDir = File("/jsonDir")
+    private val classPathElements = emptySet<File>()
+
+    @Test
+    fun `build configuration for local target with a server source without a base revision`() {
+        val serverSource = ServerSource(
+            url = "aUrl",
+            repositoryId = "aRepositoryId",
+            branchName = "aBranchName",
+        )
+        val localTarget = LocalTarget(repositoryDir = File("/repositoryDir"))
+        val syncDirection = SyncDirection("syncDirection", serverSource, localTarget)
+
+        val config = buildMpsRunConfigurationForLocalTarget(syncDirection, classPathElements, jsonDir)
+        val jvmArgs = config.jvmArgs
+
+        val expectedJvmArgs = listOf(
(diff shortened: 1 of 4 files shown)


bulk-model-sync-gradle/src/test/kotlin/org/modelix/model/sync/bulk/gradle/MpsRunnerConfigurationTest.kt
```diff
@@ -0,0 +1,95 @@
...
+        val expectedJvmArgs = listOf(
+            "-Dmodelix.mps.model.sync.bulk.input.path=/jsonDir",
+            "-Dmodelix.mps.model.sync.bulk.input.modules=",
+            "-Dmodelix.mps.model.sync.bulk.input.modules.prefixes=",
+            "-Dmodelix.mps.model.sync.bulk.repo.path=/repositoryDir",
+            "-Dmodelix.mps.model.sync.bulk.input.continueOnError=false",
+            "-Xmx2g",
+        )
+
+        assertEquals(expectedJvmArgs, jvmArgs)
```
languitar, line 33: Maybe use set comparison because the argument order probably doesn't matter?

odzhychko: Yeah, that would be cleaner.
I changed it to Kotest to use the appropriate assertions.

languitar (commented):
I can't judge the other changes as I haven't seen much about how MPS is launched so far.

bulk-model-sync-gradle/src/test/kotlin/org/modelix/model/sync/bulk/gradle/MpsRunnerConfigurationTest.kt
```diff
@@ -0,0 +1,95 @@
...
+        jvmArgs shouldContainExactlyInAnyOrder expectedJvmArgs
+    }
+
+    @Test
+    fun `build configuration for local target with a server source with a base revision and branch name`() {
+        val serverSource = ServerSource(
+            url = "aUrl",
+            repositoryId = "aRepositoryId",
+            branchName = "aBranchName",
+            baseRevision = "aBaseRevision",
```
mhuster23, line 42: We also need similar test cases for `baseRevision=null`, as it is also valid to provide a revision or pair of repoId and branchName without a baseRevision.

These cases are broken right now with this PRs changes.
The fact that this isn't covered by existing tests right now emphasizes our need to be able to test multiple gradle configurations.

mhuster23 (changes requested):
Small side note: From a reviewer's perspective it would have been nice to separate the actual fix and the refactoring into separate commits. It took me a bit to spot the fix.

bulk-model-sync-gradle/src/test/kotlin/org/modelix/model/sync/bulk/gradle/MpsRunnerConfigurationTest.kt
odzhychko: > as it is also valid to provide a revision or pair of repoId and branchName without a baseRevision.

If a `baseRevision` is not provided, the `repositoryId` and `branchName` are not passed to the `MPSBulkSynchronizer`. The data is downloaded in a separate MPS task.

So that case should not be broken.

> The fact that this isn't covered by existing tests right now emphasizes our need to be able to test multiple gradle configurations.

Definitely. But for now, this is out of scope for me.

odzhychko:
> Small side note: From a reviewer's perspective it would have been nice to separate the actual fix and the refactoring into separate commits. It took me a bit to spot the fix.

I will do that.

bulk-model-sync-gradle/src/test/kotlin/org/modelix/model/sync/bulk/gradle/MpsRunnerConfigurationTest.kt
mhuster23: > If a baseRevision is not provided, the repositoryId and branchName are not passed to the MPSBulkSynchronizer.

Ah yes, you are right it is only relevant for the `ExportFromModelServerTask`.

> But for now, this is out of scope for me.

Agreed.

odzhychko:
> Small side note: From a reviewer's perspective it would have been nice to separate the actual fix and the refactoring into separate commits. It took me a bit to spot the fix.

Done

bulk-model-sync-gradle/src/test/kotlin/org/modelix/model/sync/bulk/gradle/MpsRunnerConfigurationTest.kt
odzhychko: > If a baseRevision is not provided, the repositoryId and branchName are not passed to the MPSBulkSynchronizer.

I update the test case `build configuration for local target with a server source without a base revision` to test this, just to be sure.

mhuster23 (dismissed):

mhuster23 (approved):

PR #1003, opened 2024-08-23
Migrate to ESLint 9

odzhychko:
With the update to v9, ESLint did not support reading configuration in subdirectories.
We relied on this feature when running ESLint with pre-commit.
Therefore, the configuration was unified at the top-level of the repository.
This resulted in a cleanup of linting related build code in the `ts-model-api` and `vue-model-api` directories.
The unification makes maintenance easier anyway.

The official ESLint hook is not used anymore because of the issue described in https://github.com/eslint/eslint/issues/18465
The local hook works around that issue.
It also does not require duplicating versions of NPM packages.
But it requires installing NPM dependencies before running pre-commit hooks.

* [ ] Code is readable


odzhychko:
The build fails because of the problems described here: https://github.com/eslint/eslint/issues/18465

eslint.config.mjs
```diff
@@ -0,0 +1,38 @@
...
+import pluginJs from "@eslint/js";
+import tseslint from "typescript-eslint";
+import eslintConfigPrettier from "eslint-config-prettier";
+
+export default [
+  { files: ["**/*.{js,mjs,cjs,ts}"] },
+  { ignores: ["**/.gradle", "**/dist/", "**/build/"] },
+  {
+    languageOptions: {
+      ecmaVersion: "latest",
```
languitar, line 11: Is that safe? Or might we get spurious linting error when "latest" increments?

eslint.config.mjs
```diff
@@ -16,6 +16,13 @@ export default [
   pluginJs.configs.recommended,
   ...tseslint.configs.recommended,
   eslintConfigPrettier, // > Turns off all rules that are unnecessary or might conflict with Prettier.
+  {
+      // The Redocly plugin uses CommonJS syntax for imports and exports as in the Redocly documentation.
+      files: ["model-server-openapi/**/*"],
+      rules: {
+        "@typescript-eslint/no-require-imports": "off"
+      }
+  },
```
languitar, line 25: Maybe this change could be part of the first commit?

languitar (approved):
Just two thoughts while reviewing. Looks fine otherwise.

vue-model-api/.eslintrc.js
```diff
@@ -1,44 +0,0 @@
-module.exports = {
-  env: {
-    browser: true,
-    es2021: true,
```
odzhychko, line 4: I missed to take the `es2021: true` to the new configuration.

eslint.config.mjs
odzhychko: I just applied the value we had before.
I have to think about what effect specifying `[ecmaVersion](https://typescript-eslint.io/packages/parser/#ecmaversion)` would have.

eslint.config.mjs
odzhychko: Done

vue-model-api/.eslintrc.js
odzhychko: Done by using `...globals.es2021`.

eslint.config.mjs
odzhychko: I do not fully understand what it does.
For me, it is out of scope to look into what effects changing the `ecmaVersion` has and to reevaluate our previous configuration of it.
Therefore, I want to keep it as it was in this PR.

languitar (approved):

PR #1029, opened 2024-09-02
MODELIX-1015 model-api-gen-gradle unusable in latest releases

mhuster23:

* [ ] Code is readable

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -45,6 +45,7 @@ testcontainers = "1.20.1"
 [libraries]
 
 kotlin-stdlib-common = { group = "org.jetbrains.kotlin", name = "kotlin-stdlib-common", version.ref = "kotlin" }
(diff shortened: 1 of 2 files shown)


languitar:
Why wasn't this noticed by tests? Can we update the tests to somehow make this visible?

mhuster23:
Our tests don't cover publication at all. In the past we published these locally and then used them in the test-projects, but we moved away from that since it was too easy to forget the core build or the `publishToMavenLocal` step and accidentally test the wrong version.

I'm not aware of plugins which can automatically validate the POM.

languitar:
For verifying the overall functionality, we should have, at the highest level of the test pyramid, a real e2e test verifying that the integrated product is usable. This would have been caught that way. It would probably be enough to test this on main only and only in CI to cover these issues and to prevent automatic tagging on releases.

benedekh (approved):

PR #1039, opened 2024-09-06
fix(model-server): make specifying a key ID optional

odzhychko:
If no static key ID is configured, then the key ID specified in the JWT is used to verify the token. The key is looked up at the JWK URL with the key ID from the token.

This is a fix because it restores functionality that accidentally broke in version 8.14.0.

* [ ] Code is readable

### DIFF
--- a/authorization/src/main/kotlin/org/modelix/authorization/AuthorizationConfig.kt
+++ b/authorization/src/main/kotlin/org/modelix/authorization/AuthorizationConfig.kt
@@ -19,6 +19,7 @@ package org.modelix.authorization
 import com.auth0.jwk.JwkProvider
 import com.auth0.jwk.JwkProviderBuilder
 import com.auth0.jwt.JWT
+import com.auth0.jwt.JWTVerifier
 import com.auth0.jwt.algorithms.Algorithm
 import com.auth0.jwt.interfaces.DecodedJWT
 import io.ktor.server.application.Application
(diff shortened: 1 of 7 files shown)


odzhychko:
The build is broken because of new bug in Redocly.
See https://github.com/Redocly/redocly-cli/pull/1717

model-server/src/test/kotlin/org/modelix/model/server/AuthorizationTest.kt
```diff
@@ -0,0 +1,315 @@
...
+                      ]
+                    }
+                  }
+                ]
+              }
+            }
+        """
+
+        // Reuse on container across all tests. The configuration and state does not change in between.
+        private val keycloak: GenericContainer<*> = GenericContainer("quay.io/keycloak/keycloak:25.0.4")
```
languitar, line 178: There would be a distinct testcontainers keycloak module: https://testcontainers.com/modules/keycloak/

model-server/src/test/kotlin/org/modelix/model/server/AuthorizationTest.kt
```diff
@@ -0,0 +1,315 @@
...
+            val modelClient = createModelClient(token)
+
+            assertDoesNotThrow {
+                modelClient.init()
+            }
+        }
+    }
+
+    @Test
+    fun `authorization fails to uses key ID if it does not exist at JWK URI`() {
```
languitar, line 251:
```suggestion
    fun `authorization fails to use key ID if it does not exist at JWK URI`() {
```

model-server/src/test/kotlin/org/modelix/model/server/AuthorizationTest.kt
```diff
@@ -0,0 +1,315 @@
...
+            val modelClient = createModelClient(token)
+
+            val exception = assertThrows<ClientRequestException> { modelClient.init() }
+
+            assertEquals(HttpStatusCode.Unauthorized, exception.response.status)
+        }
+    }
+
+    @Test
+    fun `authorization succeeds if key ID in token matches configured key ID from JWK URI`() {
```
languitar, line 267:
```suggestion
    fun `authorization succeeds if key ID in token matches the configured key ID from JWK URI`() {
```

model-server/src/test/kotlin/org/modelix/model/server/AuthorizationTest.kt
```diff
@@ -0,0 +1,315 @@
...
+            val modelClient = createModelClient(token)
+
+            assertDoesNotThrow {
+                modelClient.init()
+            }
+        }
+    }
+
+    @Test
+    fun `authorization fails if key ID in token does not match configured key ID from JWK URI`() {
```
languitar, line 286:
```suggestion
    fun `authorization fails if key ID in token does not match the configured key ID from JWK URI`() {
```

authorization/src/main/kotlin/org/modelix/authorization/AuthorizationConfig.kt
```diff
@@ -198,3 +218,8 @@ private fun getBooleanFromEnv(name: String): Boolean? {
         throw IllegalArgumentException("Failed to read boolean value $name", ex)
     }
 }
+
+internal fun getVerifierForSpecificAlgorithm(algorithm: Algorithm): JWTVerifier =
+    JWT.require(algorithm)
+        .acceptLeeway(0L)
```
languitar, line 224: [From the ](https://github.com/auth0/java-jwt/blob/fb6d00ad9773c6e7624c518feb2d06ed191287fa/lib/src/main/java/com/auth0/jwt/JWTVerifier.java#L71)

0 is actually the default, which makes sense, as that reasonably assumes synchronized clocks across devices.

languitar (commented):
I'm not totally confident on the configuration changes, which are pretty hard to understand given the already existing complexity. Here's just my two cents of things I found while reading.

authorization/src/main/kotlin/org/modelix/authorization/AuthorizationConfig.kt
odzhychko: Ok, then I will remove it.

model-server/src/test/kotlin/org/modelix/model/server/AuthorizationTest.kt
odzhychko: We can currently not use it, because it does not support Java 11.

See https://github.com/dasniko/testcontainers-keycloak/issues/158

authorization/src/main/kotlin/org/modelix/authorization/AuthorizationConfig.kt
odzhychko: Done

model-server/src/test/kotlin/org/modelix/model/server/AuthorizationTest.kt
languitar: wtf, ok :D

PR #1042, opened 2024-09-06
Revert to redocly 1.22.1

odzhychko:
Revert because of newly introduced bug https://github.com/Redocly/redocly-cli/issues/1715

* [ ] Code is readable

### DIFF
--- a/model-server-openapi/build.gradle.kts
+++ b/model-server-openapi/build.gradle.kts
@@ -19,6 +19,7 @@ val bundleSpecs = tasks.register<NpxTask>("bundleSpecs") {
 
     dependsOn(tasks.getByName("npmInstall"))
 
+    inputs.file("package-lock.json")
     inputs.dir(specDir)
 
     outputs.cacheIf { true }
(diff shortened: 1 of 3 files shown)


mhuster23 (approved):

model-server-openapi/build.gradle.kts
```diff
@@ -19,6 +19,7 @@ val bundleSpecs = tasks.register<NpxTask>("bundleSpecs") {
 
     dependsOn(tasks.getByName("npmInstall"))
 
+    inputs.file("package-lock.json")
```
languitar, line 22: We need this hack because npmInstall and bundleSpecs don't share an output file in between?

odzhychko: This is needed so that the task is run if the version of Redocly changes.

The issues were that we merged a broken version of Redocly accidentally because the task outputs were taken from the cache in the build of the PR created (and merged) by Dependabot.

languitar: I'm just wondering if that single file is the correct marker to use here. The [upstream documentation uses also the `package.json` file](https://github.com/node-gradle/gradle-node-plugin/blob/main/docs/usage.md#executing-npm-commands-via-npx) and we should also include the redocly configuration here, which I seem to have missed.

odzhychko: I think the `package.json` is only relevant if it also contains configuration that influences the Gradle task or contains scripts that are invoked by the Gradle task.

Our Gradle task depends solely on the installed version of Redocly and the configuration, which is a separate issue.

So depending on `package-lock.json` without `package.json` seems fine.

PR #956, opened 2024-08-06
MODELIX-914 Remove in-memory query cache and support querying multiple large branches in parallel

slisson:
TODO description of the PR

## To be verified by reviewers

* [ ] Relevant public API members have been documented
* [ ] Documentation related to this PR is complete
  * [ ] Boundary conditions are documented
  * [ ] Exceptions are documented
  * [ ] Nullability is documented if used
* [ ] Touched existing code has been extended with documentation if missing
* [ ] Code is readable
* [ ] New features and fixed bugs are covered by tests

PR #1069, opened 2024-09-16
fix(model-datastructure): avoid stack overflows in `IBulkQuery.Value.onReceive`

odzhychko:
The handlers passed to `onReceive` are now always executed in `IBulkQuery.executeQuery`.

Previously, the handler could be executed immediately if the value was loaded from the cache.
This resulted in stack overflow if the handlers passed `onReceive` called `onReceive` recursively on values that were immediately loaded from the cache.

* [ ] Code is readable


odzhychko:
Might be resolved by #956 .
If not we have the issue [MODELIX-1013](https://issues.modelix.org/issue/MODELIX-1013) for it, that can be worked on after #956 is finished.

PR #1078, opened 2024-09-20
fix(model-datastructure): correctly calculate linear history between the same version

odzhychko:
Previously, calculating the linear history between the same version resulted in linear history that contained more than only that one version.

This in turn resulted in needles computations and deltas that were too big.

Fixes: MODELIX-1025

* [ ] Code is readable


languitar (approved):
This PR should probably also improve the documentation of `load` to explicitly say that both from and to version are always the start and end of the returned list (aka inclusive).

languitar (approved):

PR #1089, opened 2024-09-26
Improve loggig

languitar:
* Adds the branch name to the in-memory model loading messages for better tracing
* Enables call logging to support tracing individual requests on the model-server.

* [ ] Code is readable

### DIFF
--- a/model-server/src/main/kotlin/org/modelix/model/server/Main.kt
+++ b/model-server/src/main/kotlin/org/modelix/model/server/Main.kt
@@ -28,11 +28,16 @@ import io.ktor.server.engine.embeddedServer
 import io.ktor.server.http.content.staticResources
 import io.ktor.server.netty.Netty
 import io.ktor.server.netty.NettyApplicationEngine
+import io.ktor.server.plugins.callloging.CallLogging
+import io.ktor.server.plugins.callloging.processingTimeMillis
 import io.ktor.server.plugins.contentnegotiation.ContentNegotiation
 import io.ktor.server.plugins.cors.routing.CORS
 import io.ktor.server.plugins.forwardedheaders.ForwardedHeaders
+import io.ktor.server.plugins.origin
 import io.ktor.server.plugins.statuspages.StatusPages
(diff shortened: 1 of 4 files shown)


model-server/src/main/kotlin/org/modelix/model/server/Main.kt
```diff
@@ -192,6 +197,19 @@ object Main {
                     permissionSchema = ModelServerPermissionSchema.SCHEMA
                 }
                 install(ForwardedHeaders)
+                install(CallLogging) {
+                    format { call ->
+                        // Resemble the default format but include remote host and user agent for easier tracing on who issued a certain request.
```
odzhychko, line 202: This fixes https://issues.modelix.org/issue/MODELIX-865/Enable-KTOR-request-logging-for-model-server

> Request should be logged at debug level.

By default `INFO` seems to be used.
See https://ktor.io/docs/server-call-logging.html#logging_level

> Also log thread ID for request, if possible

Can we add the thread ID?

languitar: I wasn't aware of the ticket. Picked it up.

Why was it decided to deviate from the default log level in that ticket? Even if I disable all other logging, that's the most basic information I need to trace what's going on on the model-server. I wouldn't want to hide that in the debug-noise.

languitar: Regarding the thread ID: I can log the thread ID at the time the call logging message is assembled. But that's not guaranteed to be the same thread that executed (parts of) the request. I don't think that helps but instead confuses more.

odzhychko: I agree on both.
Lets just leave a note in the ticket why we deviated from the description.

odzhychko (approved):

model-server/src/main/kotlin/org/modelix/model/server/Main.kt
languitar: noted in the ticket.

PR #1172, opened 2024-11-08
build(model-client): remove jacoco Gradle plugin

odzhychko:
Coverage is reported with kotlinx-kover.
Remove the `java-library` plugin as it was only necessary for the `jacoco` plugin.

This also solves the warning of incompatibility between `java-library` and `kotlin("multiplatform")`.

* [ ] Code is readable


odzhychko:
[Here](https://github.com/modelix/modelix.core/pull/1172#issuecomment-2464478411) you can see, that coverage works without Jacoco for the model-client.

languitar (approved):
Must have missed that singular instance of coverage when adding the general solution.

PR #1228, opened 2024-12-05
feat(model-server)!: use a consistent status code for query failures

languitar:
When a modelql query fails due to a mismatch between what's expected by the query and the underlying data, always returns the same status code.

I have chosen 422 here. 5xx status codes are reserved for actual application failures instead of business logic errors. 404 would be too specific. 422 looks like it covers all kinds of errors related to being unable to "process" an otherwise valid query successfully.

* [ ] Code is readable

### DIFF
--- a/model-server-openapi/specifications/model-server-v2.yaml
+++ b/model-server-openapi/specifications/model-server-v2.yaml
@@ -322,8 +322,8 @@ paths:
       responses:
         "200":
           $ref: '#/components/responses/200json'
-        "404":
-          $ref: '#/components/responses/modelql-404'
+        "422":
+          $ref: '#/components/responses/ModelQlQueryExecutionFailed'
         default:
           $ref: '#/components/responses/GeneralError'
   /repositories/{repository}/init:
@@ -429,8 +429,8 @@ paths:
       responses:
         "200":
           $ref: '#/components/responses/200json'
-        "404":
-          $ref: '#/components/responses/modelql-404'
+        "422":
+          $ref: '#/components/responses/ModelQlQueryExecutionFailed'
         default:
           $ref: '#/components/responses/GeneralError'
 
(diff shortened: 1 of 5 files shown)


odzhychko (commented):
I search for other occurrences of `IllegalStateException`/`RuntimeException` in `modelql-core`. Many of them also seem to qualify to produce a 422 status code rather than a 500 status code. 

I think, we should also take care of them or leave everything as is.
I would prefer consistent odd behavior over inconsistent behavior.

languitar:
> I search for other occurrences of `IllegalStateException`/`RuntimeException` in `modelql-core`. Many of them also seem to qualify to produce a 422 status code rather than a 500 status code.
> 
> I think, we should also take care of them or leave everything as is. I would prefer consistent odd behavior over inconsistent behavior.

500 is not an option due to the retry behavior of the client by default. So we should probably make everything a 422 then I guess?

odzhychko:
We also could disable the retrying for the ModelQL endpoints. But then we would lose the retries for the cases when a service is really not reachable for a moment...

So being consistent with returning `422` seems like the way to go.

languitar:
@slisson any opinion on this?

slisson:
This retry mechanism seems to cause more issues than it solves. In most cases it's not a temporary issue, and if it is temporary then we don't know when it will recover. The caller should always be prepared for unsuccessful request and have it's own strategy.

I would disable the retries and forward unsuccessful request directly to the caller.

languitar:
> This retry mechanism seems to cause more issues than it solves. In most cases it's not a temporary issue, and if it is temporary then we don't know when it will recover. The caller should always be prepared for unsuccessful request and have it's own strategy.
> 
> I would disable the retries and forward unsuccessful request directly to the caller.

I'm fine with that. But we would still misuse the 5xx namespace for business errors. Any proxy wrapping around the model-server might again introduce such issues if we continue to use 500 for these errors.

PR #1234, opened 2024-12-08
fix(modelql)!: disable implicit queries to avoid accidental performan…

slisson:


languitar (changes requested):
Wouldn't it be a better idea to make the choice if implicit queries are allowed or not a constructor or builder argument of the client? That way, application can decide on a suitable configuration mechanism for them to apply.

Otherwise, the new option would need to be explained in the docs.

odzhychko (changes requested):
The code and idea look good.
But the PR has two commits with the same message.

odzhychko:
> Wouldn't it be a better idea to make the choice if implicit queries are allowed or not a constructor or builder argument of the client? That way, application can decide on a suitable configuration mechanism for them to apply.
> 
> Otherwise, the new option would need to be explained in the docs.

This would also allow clients to be configured differently in one application.

slisson:
I was thinking about removing this feature completely, but didn't want to make switching to a new version unnecessarily difficult. A poor performance is still better than exceptions. Still not sure wether we should support this feature or deprecate it.

PR #1238, opened 2024-12-10
Fix broken publishing workflow

slisson:
See https://github.com/modelix/modelix.core/actions/runs/12248521140


.nvmrc
languitar, line 1: Missing newline at EOF

languitar: We should probably add a config for renovate to update this?

languitar (approved):

PR #1250, opened 2024-12-13
fix(authorization): reload keys when the file changes

slisson:
If the Kubernetes secret containing the RSA key changed, the container usually isn't restarted, so we have to make sure we don't use an outdated key.


languitar:
Usually, one could also apply something like https://github.com/stakater/Reloader to avoid having to deal with this internally. However, that assumes that rolling releases are possible.

PR #1260, opened 2024-12-18
fix(model-server): do not put objects into store in request thread

odzhychko:
This is already done in `putRepositoryObjects`.
In `postRepositoryBranch` this should also be done.

It was not a critical issue because the model client only sent at most 1000 objects to `postRepositoryBranch`. Still, if something blocking happened in Ignite, it might have blocked the request thread.

* [ ] Code is readable

### DIFF
--- a/model-server/src/main/kotlin/org/modelix/model/server/handlers/ModelReplicationServer.kt
+++ b/model-server/src/main/kotlin/org/modelix/model/server/handlers/ModelReplicationServer.kt
@@ -216,8 +216,11 @@ class ModelReplicationServer(
         val branchRef = repositoryId(repository).getBranchReference(branch)
         val deltaFromClient = call.receive<VersionDelta>()
         deltaFromClient.checkObjectHashes()
-        @OptIn(RequiresTransaction::class) // no transactions required for immutable store
-        repositoriesManager.getStoreClient(RepositoryId(repository), true).putAll(deltaFromClient.getAllObjects())
+        val objectsFromClient = deltaFromClient.getAllObjects()
+        withContext(Dispatchers.IO) {
+            @OptIn(RequiresTransaction::class) // no transactions required for immutable store
+            repositoriesManager.getStoreClient(RepositoryId(repository), true).putAll(objectsFromClient)
+        }
         @OptIn(RequiresTransaction::class)
         val mergedHash = runWrite {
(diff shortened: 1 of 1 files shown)


slisson:
Operations on the immutable store shouldn't be blocking and therefore don't need to run on `Dispatchers.IO`.

odzhychko:
> Operations on the immutable store shouldn't be blocking and therefore don't need to run on `Dispatchers.IO`.

Are you sure?
Even though it cannot be blocked by transactions, `IgniteCache.putAll` still makes a network request which blocks the thread until it has sent all data.

slisson:
> > Operations on the immutable store shouldn't be blocking and therefore don't need to run on `Dispatchers.IO`.
> 
> Are you sure? Even though it cannot be blocked by transactions, `IgniteCache.putAll` still makes a network request which blocks the thread until it has sent all data.

But it's not like we can't let the default dispatcher do anything. It would be easier to let Ktor run on the IO dispatcher then, but we also don't want the thread pool to grow without limit if there are many requests.

odzhychko:
> But why doesn't Ktor run on the IO dispatcher by default? Wouldn't that make everything easier?

Running on `Dispatchers.IO` would make the code easier, but also less performant.

It would be similar to traditional implementation of web servers.
Traditional implementations of servers just had a thread pool for handling requests.
And each thread could handle only one request simultaneously.

Ktor does not use `Dispatchers.IO` because it uses non-blocking IO.
This ideally utilizes compute resource better because a thread can handle multiple requests simultaneously.

---

For now, what we have works. We at least profit from async IO when sending and receiving request data.

But if we want to simplify code, we could look for a web framework that supports lightweight JVM threads (Project Loom) like Spring[^1] and Quarkus[^2] or try using it with Ktor.
There is a presentation of using it with Ktor[^3].

Another extreme would be to go fully async by using the async Ignite-API (e.g. `putAllAsync`).
This would avoid `Dispatchers.IO`(and maybe better utilize threads) but complicate the code.

[^1]: https://www.baeldung.com/spring-6-virtual-threads
[^2]: https://quarkus.io/guides/virtual-threads
[^3]: https://resources.jetbrains.com/storage/products/kotlinconf-2023/Coroutines%20and%20Loom%20Behind%20The%20Scenes.pdf

slisson (approved):

PR #1296, opened 2025-01-07
fix(modelql): set `INodeResolutionScope` when running query on `INode`

odzhychko:
Fixes https://issues.modelix.org/issue/MODELIX-997

* [ ] Code is readable

### DIFF
--- a/modelql-untyped/build.gradle.kts
+++ b/modelql-untyped/build.gradle.kts
@@ -17,7 +17,9 @@ kotlin {
         }
         val commonTest by getting {
             dependencies {
+                implementation(project(":model-datastructure"))
                 implementation(kotlin("test"))
+                implementation(libs.kotlin.coroutines.test)
             }
         }
         val jvmMain by getting {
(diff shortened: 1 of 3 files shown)


odzhychko:
> Is there any reason we didn't have this in place before?

I don't know.
My guess is, that it was forgotten because `INode.query`  is only used very rarely.

languitar:
> > Is there any reason we didn't have this in place before?
> 
> I don't know. My guess is, that it was forgotten because `INode.query` is only used very rarely.

At least that way is documented: https://docs.modelix.org/modelix/main/core/howto/modelql.html#_run_query_on_an_inode

odzhychko:
> > > Is there any reason we didn't have this in place before?
> > 
> > 
> > I don't know. My guess is, that it was forgotten because `INode.query` is only used very rarely.
> 
> At least that way is documented: https://docs.modelix.org/modelix/main/core/howto/modelql.html#_run_query_on_an_inode

Ah, but we probably never triedout to resolve a node reference inside `INode.query`.

PR #1875, opened 2025-08-18
fix(mps-sync-plugin): properly check connection state when having multiple connections

abstraktor:
So far, this job fell asleep as soon as any connection is requiring authorization. This resulted in a lack of connection checking for other connections.

This fixes the following issues:

- When opening one project without logging in; And opening another project, Then "Click to login" would not disappear even after clicking it and logging in
- Sometimes, The local server would already be stopped, resulting in being redirected to an error page at the end of the login workflow


AlexeiQ:
Functionally this works now 👍 

- Open project_1 in MPS without loggin in and close it
- Open project_2 in MPS, log in and the modelix-indicator doesn't blink anymore.
- The indicator-tooltip shows status connected for project_2 and status authorization-required for project_1, as expected.

slisson:
The issue with the proposed change was that the connectionCheckingJob kept launching new coroutines even if there is already one waiting for the server.

Now each connection has it's own connectionCheckingJob.

PR #1901, opened 2025-08-28
feat(model-client): load readonly version in js client

abstraktor:


model-client/src/jsMain/kotlin/org/modelix/model/client2/ClientJS.kt
```diff
@@ -192,6 +201,19 @@ internal class ClientJSImpl(private val modelClient: ModelClientV2) : ClientJS {
...
 
+    override fun loadReadonlyVersion(repositoryId: String, versionHash: String): Promise<VersionInformationWithModelTree> {
+        return GlobalScope.promise {
+            val version = modelClient.loadVersion(RepositoryId(repositoryId), versionHash, null)
+            VersionInformationWithModelTree(
+                VersionInformationJS(
+                    (version as CLVersion).author,
+                    version.getTimestamp()?.toJSDate(),
+                    version.getContentHash()),
+                MutableModelTreeJsImpl(VersionedModelTree(version).withAutoTransactions())
```
slisson, line 212: Was there a reason not to use `version.getModelTree().asMutableSingleThreaded()`?

abstraktor: When I use `MutableModelTreeJsImpl(version.getModelTree().asMutableSingleThreaded())` then I get an `UnsupportedOperationException` upon initialization of the `MutableModelTreeJsImpl`. It is calling `addListener` there, which the provided `SingleThreadMutableModelTree` doesn't support.

slisson: OK, then let's keep it like that and refactor it later.

PR #2032, opened 2025-10-07
fix(model-datastructure): handle division by zero in getIntervalIndex…

abstraktor:
… and add corresponding tests


slisson:
Timestamps are only stored with seconds precision so I reverted the usage of `toEpochMilliseconds()` back to `epochSeconds` to stay consistent. I also handle zero length intervals now in the constructor to make it more readable.

slisson (approved):

PR #2607, opened 2026-04-23
feat(model-client): stamp version attributes on ReplicatedModel commits and expose them in history queries

SaSteffen:
- `ReplicatedModel` accepts a `versionAttributes` supplier that stamps key-value metadata on every locally created version
  - History queries (sessions, intervals, pagination) aggregate and expose attributes across versions                                                                                                                                                                                                                               
  - JS API gains full parity: supplier is called per commit, not snapshotted at connect time

### DIFF
--- a/model-datastructure/src/commonMain/kotlin/org/modelix/datastructures/history/HistoryIndexNode.kt
+++ b/model-datastructure/src/commonMain/kotlin/org/modelix/datastructures/history/HistoryIndexNode.kt
@@ -28,6 +28,7 @@ sealed class HistoryIndexNode : IObjectData {
     abstract val firstVersion: ObjectReference<CPVersion>
     abstract val lastVersion: ObjectReference<CPVersion>
     abstract val authors: Set<String>
+    abstract val attributes: Map<String, Set<String>>
     abstract val size: Long
     abstract val height: Long
     abstract val minTime: Instant
@@ -60,6 +61,7 @@ sealed class HistoryIndexNode : IObjectData {
             firstVersion = firstVersion,
             lastVersion = other.lastVersion,
             authors = authors + other.authors,
+            attributes = unionAttributes(attributes, other.attributes),
             size = size + other.size,
             height = max(height, other.height) + 1,
             minTime = minTime,
@@ -97,6 +99,7 @@ sealed class HistoryIndexNode : IObjectData {
                         maxTime = times.getOrElse(1) { times[0] },
                         child1 = referenceFactory.fromHashString(parts[6], HistoryIndexNode),
                         child2 = referenceFactory.fromHashString(parts[7], HistoryIndexNode),
+                        attributes = parts.getOrNull(8)?.let { deserializeAttributes(it) } ?: emptyMap(),
                     )
                 }
                 "L" -> {
@@ -105,6 +108,7 @@ sealed class HistoryIndexNode : IObjectData {
                             .map { referenceFactory.fromHashString(it, CPVersion) },
                         authors = parts[2].split(Separators.LEVEL2).mapNotNull { it.urlDecode() }.toSet(),
                         time = Instant.fromEpochSeconds(parts[3].toLong()),
+                        attributes = parts.getOrNull(4)?.let { deserializeAttributes(it) } ?: emptyMap(),
                     )
                 }
                 else -> error("Unknown type: " + parts[0])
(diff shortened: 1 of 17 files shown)


model-datastructure/src/commonMain/kotlin/org/modelix/datastructures/history/HistoryIndexNode.kt
```diff
@@ -28,6 +28,7 @@ sealed class HistoryIndexNode : IObjectData {
     abstract val firstVersion: ObjectReference<CPVersion>
     abstract val lastVersion: ObjectReference<CPVersion>
     abstract val authors: Set<String>
+    abstract val attributes: Map<String, Set<String>>
```
slisson: If `attributes` aggregates all unique values it can become very large. The same is theoretically also true for `authors` but was ignored there because the number of unique authors is excepted to be small. For generic `attributes` we can't make this assumption.

I will fix this scalability issue.

SaSteffen: I came across this thought during development but wanted to hear your opinion. Both truncation and specifying keys to actually accumulate are viable measures (used in combination), that wouldn't hurt my particular use case.

model-datastructure/src/commonMain/kotlin/org/modelix/model/VersionMerger.kt
```diff
@@ -116,6 +116,12 @@ constructor(private val idGenerator: IIdGenerator?) {
                 .autoMerge(commonBase.obj.ref, leftVersion.obj.ref, rightVersion.obj.ref)
                 .operations(appliedOps.map { it.getOriginalOp() })
                 .currentTime()
+                .also { builder ->
```
slisson: I don't think this behavior is desired for all use cases.

One example is that the git importer stores the git commit hash as a version attribute. That should only be assigned to the exact version that was created by the importer and not propagate to any other version.

What is your use case for version attributes?

model-server/src/test/kotlin/org/modelix/model/server/TestAttributeHelpers.kt
```diff
@@ -0,0 +1,9 @@
+package org.modelix.model.server
+
+internal fun mergeAttributes(
+    accumulated: Map<String, Set<String>>,
+    incoming: Map<String, String>,
+): Map<String, Set<String>> =
+    (accumulated.keys + incoming.keys).associateWith { key ->
+        (accumulated[key] ?: emptySet()) + setOfNotNull(incoming[key])
```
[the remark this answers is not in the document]
SaSteffen: fixed

model-datastructure/src/commonMain/kotlin/org/modelix/model/VersionMerger.kt
SaSteffen: My use case is to stamp additional information about AI (co-)authorship. I think it would be fine if we revert this change entirely.

SaSteffen: As discussed: Removed it

model-server/src/test/kotlin/org/modelix/model/server/ReplicatedModelTest.kt
```diff
@@ -121,6 +127,83 @@
...
 
+    @Test
+    fun versionAttributesAreStoredOnCreatedVersions() = runTest {
+        val client = createModelClient()
+        val repositoryId = RepositoryId(UUID.randomUUID().toString())
+        val branchReference = repositoryId.getBranchReference()
+        client.initRepository(repositoryId)
+
+        val attrs = mapOf("env" to "staging", "pipeline" to "ci-42")
+        val scope = CoroutineScope(Dispatchers.Default)
```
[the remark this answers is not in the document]
SaSteffen: @slisson Since this was done in the test before, I think we can dismiss the warning, or shall I look into it?

model-server/src/test/kotlin/org/modelix/model/server/ReplicatedModelTest.kt
```diff
@@ -121,6 +127,83 @@
...
+
+    @Test
+    fun versionAttributesCanChangeBetweenVersions() = runTest {
+        val client = createModelClient()
+        val repositoryId = RepositoryId(UUID.randomUUID().toString())
+        val branchReference = repositoryId.getBranchReference()
+        client.initRepository(repositoryId)
+
+        val currentAttrs = mutableMapOf("run" to "1")
+        val scope = CoroutineScope(Dispatchers.Default)
```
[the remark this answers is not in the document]
SaSteffen: see above

model-datastructure/src/commonMain/kotlin/org/modelix/datastructures/history/HistoryIndexNode.kt
slisson: I first thought about using a b-tree for storing the entries in a scalable way but then realized that these HistoryIndexNodes are used to show a summary to the user and storing all values isn't even useful. I chose to truncate the values and limit it to 20 unique values (the first and last 10 values).

PR #2750, opened 2026-06-18
feat(model-client): delta-load readonly versions against the last loaded one

abstraktor:
## What

Switching between readonly versions of the same repository previously reloaded the entire version object graph on every switch. This change loads each readonly version as a **delta against the previously loaded version of the same repository**, so the server only sends the delta and the surviving client's object graph is reused.

`ClientJSImpl` now keeps a per-repository memo of the last loaded readonly version and passes it as the `baseVersion`. Both readonly load sites share the memo:
- `startReplicatedModels` (the `versionHash` / readonly path)
- `loadReadonlyVersion`

The memo's lifetime is tied to the client, so the remembered version always shares the client's per-repository object graph — satisfying the `baseVersion.graph === graph` guard in `ModelClientV2.createVersion`.

## Tests

`model-client/src/jsTest/.../ReadonlyVersionDeltaLoadJsTest.kt` (TDD red→green), asserting via `MockEngine`:
- the second readonly load of a repo sends `lastKnown=<previousHash>`, while the first load and a different repo send none;
- the memo is shared across both load sites.

`./gradlew :model-client:jsTest` is green (browser + node).

## Context

Part of the "accelerate version switching" effort (itemis #1042).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


slisson (approved):
Only needs a rebase after https://github.com/modelix/modelix.core/pull/2751 is merged.

PR #2861, opened 2026-07-30
fix(model-server): index model.repository to speed up repository deletion

abstraktor:
## Problem

Deleting a repository (observed with a large TARA repo, `secure-repository-service` → model-server) times out at the client's socket timeout and fails. The delete cost scales with **total server data, not with the repo being deleted**.

## Root cause

`removeRepositoryObjectsFromDatabase` runs `DELETE FROM model WHERE repository = ?` (`IgniteStoreClient.kt:178`). The `model` table's only index is the composite primary key `(key, repository)` — **key-leading** (`SqlUtils.kt`). Postgres cannot range-scan a composite index on its trailing column for an equality on that column alone, so this `DELETE` becomes a **sequential scan of the entire `model` table** (every object of every repository) on every repository delete. On a populated server this exceeds the client socket timeout.

## Fix

Add a `repository`-leading index so the bulk delete uses an index scan:

```sql
CREATE INDEX IF NOT EXISTS model_repository_idx ON <schema>.model (repository);
```

Created during schema init in `SqlUtils.kt` via a new `ensureIndexIsPresent` helper (analogue of the existing `ensureColumnIsPresent`), so existing deployments pick it up on next startup. Doubly idempotent: a `DatabaseMetaData.getIndexInfo` pre-check plus `CREATE INDEX IF NOT EXISTS`.

## Test

`IgnitePostgresRepositoryRemovalTest` gains a case asserting `model_repository_idx` exists in `pg_indexes` after schema init (Testcontainers Postgres).

## Verification status

- ktlint (pre-commit) ✅
- detekt `:model-server:detektMain` + `:model-server:detektTest` ✅
- compile + non-container tests ✅
- ⚠️ The Testcontainers Postgres tests (incl. the new one) could **not** be run in the authoring environment (no Docker daemon). **Please confirm the full `:model-server:jvmTest` passes in CI/a Docker-capable environment before merge.**

## Not in scope (secondary follow-up)

`IgniteStoreClient.removeRepositoryObjects` (`IgniteStoreClient.kt:128-147`) still evicts cache entries via a full-cache `ScanQuery` across all repositories — O(total cached data) per delete. The DB index fixes the dominant timeout cause; partitioning/affinity-keying the cache by repository so eviction is O(repo size) is recommended future work, deliberately left out of this focused fix.

---
_Investigation and fix drafted with AI assistance (Claude)._

### DIFF
--- a/model-server/src/main/kotlin/org/modelix/model/server/SqlUtils.kt
+++ b/model-server/src/main/kotlin/org/modelix/model/server/SqlUtils.kt
@@ -90,6 +90,48 @@ internal class SqlUtils(private val connection: Connection) {
         return false
     }
 
+    /**
+     * Checks whether an index with the given name exists but is marked INVALID in the catalog.
+     *
+     * An index built with `CREATE INDEX CONCURRENTLY` that is interrupted (e.g. the connection or
+     * pod dies mid-build) is left behind in an INVALID state: it is unusable by the query planner,
+     * yet a later `CREATE INDEX IF NOT EXISTS` — and [isIndexExisting], which via `getIndexInfo`
+     * also reports invalid indexes — would treat the name as taken and never rebuild it. Detecting
+     * the invalid state lets the caller drop and recreate the index instead of skipping it forever.
+     *
+     * Returns `false` when no such index exists (`to_regclass` yields NULL, matching no row).
+     */
+    @Throws(SQLException::class)
+    fun isIndexInvalid(schemaName: String, indexName: String): Boolean {
+        connection.prepareStatement(
+            """
+                SELECT NOT i.indisvalid AS is_invalid
+                FROM pg_index i
+                WHERE i.indexrelid = to_regclass(?)
+            """.trimIndent(),
+        ).use { stmt ->
+            stmt.setString(1, "$schemaName.$indexName")
+            stmt.executeQuery().use { rs ->
+                return rs.next() && rs.getBoolean("is_invalid")
+            }
+        }
+    }
+
+    /**
+     * Ensures the index exists and is valid, (re)building it via the supplied [creationSql].
+     *
+     * The caller is expected to supply a `CREATE INDEX CONCURRENTLY` statement so schema
+     * initialization does not take a table-level `ShareLock` that would block every writer for the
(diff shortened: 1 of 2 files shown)


abstraktor:
### Real-world before/after (production dataset, ~5.6M rows / 2.33 GB `model` table)

Validated the missing-index problem and this fix on a live dataset. Probe uses a **non-existent** repository key so it matches zero rows — this isolates the pure scan cost of the `WHERE repository = ?` delete predicate:

```sql
EXPLAIN (ANALYZE, BUFFERS) SELECT count(*) FROM model WHERE repository = '<non-existent key>';
```

**Before** (only the `(key, repository)` primary key exists):
- `Parallel Seq Scan on model` — `Rows Removed by Filter` ≈ 5.6M
- `Buffers: read ≈ 207,725` (~1.6 GB), I/O-bound
- **Execution Time: 13,230 ms**

**After** `CREATE INDEX CONCURRENTLY model_repository_idx ON model (repository)`:
- `Index Only Scan using model_repository_idx`, `Heap Fetches: 0`
- `Buffers: hit=3`
- **Execution Time: 0.05 ms** (~260,000× faster)

Note the scan cost is ~13 s for a **zero-match** query — i.e. independent of the target repository's size, and already over a typical client socket timeout before any rows are deleted or the cache eviction runs. The `Index Only Scan` also confirms the planner actually **chooses** the index for this predicate (the one thing the unit test can't assert).

### Recommendation: build the index `CONCURRENTLY`

As written, the migration runs a plain (non-concurrent) `CREATE INDEX` synchronously during store-client initialization on startup. On a table this size that takes a lock which conflicts with writes (a `ShareLock` blocks INSERT/UPDATE/DELETE) for the entire build — which can stall writers during a rolling deploy, or crash-loop a pod if the build outlasts the startup/liveness probe (the index is then rolled back and rebuilt from scratch on the next boot).

The datasource connection is autocommit (not inside a transaction block), so `CREATE INDEX CONCURRENTLY IF NOT EXISTS` is permitted here and avoids the blocking lock. One caveat to handle: a failed concurrent build leaves an `INVALID` index that `IF NOT EXISTS` would then skip — so a validity check / drop-and-rebuild guard is worth adding.

_Posted with AI assistance._

